PR #35, opened 2016-10-20
Fix documentation

### LINKED ISSUE #14: Getting started on the website is invalid and incomplete
It does not reflect the use of the new "dev-tools" folder, but also there is no explanation on where to expect the console to be accessible from the browser, what username/pwd to use, etc.
I think the new README.md in dev-tools/vagrant has some up-to-date content that should now be copied over to the website.

riccardomodanese:
fix #14


riccardomodanese:
Since dev-tool module has no source to compile or resources to package, I placed it under a specific profile to prevent the genartion of an empty jar.
In that way the target directory will remain empty during build time if no "dev-tool" profile is activated.
At the same time eclipse it's still able to import this module from the root pom.

PR #60, opened 2016-10-25
added connection id managing on device entity

### LINKED ISSUE #57: Add connection identifier to the device on birth message proecessing
To keep track of which physical connection is used by a device a new field should be managed on birth message processing.
In that way we can associate devices and connections.

riccardomodanese:

fix #57 (back end service only)


Coduz:
Hi @riccardomodanese ,

I'll pick up from there and i'll integrate this also on console side.

Regards ,

_Alberto_

PR #118, opened 2016-11-26
Added BDD Tests with Cucumber.

muros-ct:
Changes to Maven pom files
--------------------------
User Service implementation pom file has additional cucumber dependencies for running Cucumber BDD tests.
Master pom contains version of Cucumber.
Surefire plugin has specified custom MockedLocator for running Cucumber tests.

Implementation changes
----------------------
No changes to implementation of User service.

Implementation of MockedLocator added to test module. Used together with Mockito to mock services
that are not under test.

Tests changes
-------------
Tests with Cucumber BDD are implemented for CRUD operations on User service.

UsersJAXBContextProvider was changed because it didn't provide KapuaTmetadata to JAXB
context.

Cucumber scenario "Create user that already exists" is disabled because it fails.
Check the reason with user create implementation. Problem with transaction. Is it test
or implementation problem?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,6 +45,7 @@
         <slf4j.version>1.7.2</slf4j.version>
         <shiro.version>1.2.3</shiro.version>
         <validation-api.version>1.1.0.Final</validation-api.version>
+        <cucumber.version>1.2.4</cucumber.version>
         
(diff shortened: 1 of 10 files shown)


kartben:
please make sure to file a test-only CQ for this dependency...

PR #132, opened 2016-11-30
Added "userType" and "externalId" to User

### LINKED ISSUE #99: Add a "userType" attribute and a "externalId" attribute in user model
The userType attribute will identify the type of the authorization:

- external: the user is authenticated by sso
- device: the user represents a device identity

The externalId attribute will identify the id used for the authentication

- If the userType is external then externalId is the sso id
- If the userType is device then externalId is the client_id

lorthirk:
Closes #99

### DIFF
--- a/service/user/api/src/main/java/org/eclipse/kapua/service/user/UserType.java
+++ b/service/user/api/src/main/java/org/eclipse/kapua/service/user/UserType.java
@@ -1,3 +1,15 @@
+/*******************************************************************************
+ * Copyright (c) 2011, 2016 Eurotech and/or its affiliates and others
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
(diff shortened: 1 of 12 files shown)


service/user/internal/src/main/java/org/eclipse/kapua/service/user/internal/UserImpl.java
```diff
@@ -51,6 +52,14 @@
     @Basic
     @Column(name = "phone_number")
     private String            phoneNumber;
+    
+    @Enumerated(EnumType.STRING)
+    @Column(name = "user_type", updatable = false)
```
Coduz: Hi @lorthirk 

please also add ", nullable = false".

Thanks,

_Alberto_

service/user/api/src/main/java/org/eclipse/kapua/service/user/UserType.java
```diff
@@ -0,0 +1,21 @@
+package org.eclipse.kapua.service.user;
```
lorthirk: Forgot here as well. Adding them right away. Need to enforce them automatically...

Issue #124, opened 2016-11-29
Add more tests to user service


stefanomorson:
Could be useful to add some more unit tests to the user service:

- changeUserName() 
  - check the update on every field (except the fields of the abstract entity classes)
- add test for find()
- add test for findByName()
- add test for query
  - check the query retrieves exactly the expected users
- add test for parameters (validator)
  - spell check username

stefanomorson:
Parameter testing (validators) will be done in a separated issue

lorthirk:
@muros-ct can you confirm that all these tests have been implemented?

muros-ct:
Yes. This was implemented long time ago.
Leonardo also worked on other issue with validators, so that this is more than closed.

Coduz:
Closing this

PR #136, opened 2016-12-01
Added user service CRUD tests and assertion checks

muros-ct:
Changes are based on eclipse/kapua issue #124.

Tests ware added to basic functionalities of User service - CRUD operations.
Checking of user creation and update are based on all user field members.
Find by name and find by id also checks all user fields, as find all users do.

Previous commit had files with wrong line endings. Those ware also changed to Unix style line endings.
This is why changes are on full content of file.


Coduz:
Hi @muros-ct ,

is this ok to be merged? 

Regards, 

_Alberto_

Issue #94, opened 2016-11-20
Enable dependency injection for kapua services


stefanomorson:

stefanomorson:
As part of the commit (7b01290) also the issue #96 is solved.

stefanomorson:
Now the dependency injection with the Kapua services is working. Possible improvements can be the following:

- Switch to dependency injection when referencing kapua services inside service impleentations
- Reduce the verbosity of the locator.xml files (hopefully by removing the need to list all the provided api)
- Add support for interceptors

PR #149, opened 2016-12-07
Enable dependency injection for kapua services

stefanomorson:
Closes: 	Enable DI for kapua services #94


stefanomorson:
Can anyone take a look at this ?

muros-ct:
You mean to try it with cucumber tests from develop?
If they still pass and what changes it needs if not?

Issue #125, opened 2016-11-29
Remove tests for LockoutPolicy configuration from UserService


stefanomorson:
Lockout policy management have to be moved to authentication (CredentialService) and consequently the configuration  parameters. 
The scenarios in UserService.feature can be removed. The code of the tests (UserServiceSteps) can be maintained for checking other configuration parameters.

PR #150, opened 2016-12-09
Removal of configuration tests #125

muros-ct:
Lockout policy and other metadata tests were removed from UserService (issue #125).
Commented out test for multiple equal user creation was added and it passes ok.
Test for special characters in user name was added.


muros-ct:
Shure, you can merge it.
It is just some tests cleanup.

PR #170, opened 2016-12-23
Initial batch of Device Registry CRUD tests.

andrejn-ct:
## CRUD tests for the Device Registry service
This first set of Device Registry tests exercises the majority of the main Device Registry service (package org.eclipse.kapua.service.device.registry.internal). Cucumber based tests fortThe other packages that make up the Device Registry are not yet implemented.
**Note**: The existing jUnit tests for the Device Registry service packages are disabled (@Ignore). Mixing the preexisting jUnit and new cucumber based tests leads to failure of the jUnit tests (caused by the use of the MockedLocator). Eventually all jUnit tests will be replaced by suitable cucumber based tests.

The Device Registry implementation pom file is modified with additional Cucumber dependencies.
Surefire plugin is configured with the custom MockedLocator. This allows the Device Registry service to be tested in isolation from other services (specially authorization and authentication).

### Implementation changes
The implementation of the Device registry service has not been changed.

### Test changes
A new set of Cucumber based tests is implemented to test CRUD operations on the Device Registry service.


andrejn-ct:
Troubles with rebasing code.

PR #172, opened 2017-01-02
JWT Authentication

lorthirk:
This PR extracts everything about JWT Authentication from the `impl-consoleV2` branch in order to get it merged sooner into `develop`. It's only cherrypicking of commits already existing on `impl-consoleV2` (plus a new commit that only brings some code that was in a commit with other stuff).


Coduz:
Hi all, 

I've rebased the branch onto the latest develop.

Regards, 

_Alberto_

PR #182, opened 2017-01-10
Added integration test module with example

muros-ct:
Added qa module to master pom and project structure.

Added sql scripts that drop, create and seed full database in one step.
This will probably be removed and per service scripts will be used.

In qa module added steps and cucumber scenario for integration testing of
User service. All used services are live services and are not mocked.

KapuaSecurtyUtils is changed by using kapua-sys as a user in priveledged mode.
Change is necessary for data preparation in integration tests using cucumber step.

Kapua security utils adapted changes from new authentication.
Added table for securty token for temporarry create/drop all sql scripts.

locator.services file removed, it is replaced by locator.xml
Class references added to locator.xml to work with new authentication.

Added style to index of Cucumber test reports and kapua logo.
Style is based on eclipse kapua web page.
Added qa integration test to Cucumber reports.


muros-ct:
@stefanomorson 
Hi this is basic integration test with User, Authorization and Authentication service.
It uses Kapua Locator, no mocks and live in-memory H2.

Can we include this in develop?
This way other integration tests can be added. It will probably need refactoring, but that will only show when more integration tests are being written.

PR #185, opened 2017-01-11
Impl unit test env setup

bojanutkovic:
Set unit test environment


lorthirk:
Hi @bojanutkovic, could you please rebase on the latest `impl-consoleV2` commit to fix the conflict?

Thanks

PR #186, opened 2017-01-13
Authorization and Authentication improvements

### LINKED ISSUE #177: Improve credential model and Credential service
Hi all,

This is a proposal for improving Credential model and Credential Service.

Now a Credential is linked only to a user. This is fine when giving access to real people that want to access the platform resources. When a device needs access, we need to create a User entity that acts like a placeholder.

This is a bit misleading because a User has information like email address and home address that a device will never use. This is because the User entity represent a physical person in real life, in our model.

What I propose is to add to the Credential entity the Subject type from which the credential can be used (i.e.: User, Device,...) and the Subject id that is the "link" with the specific entity.

In this way a Credential can be assigned for a Device entity or for a User entity. This also makes credentials tight pretty simple, because you can a credential to a specific Device, or a credential that can be used for all devices.

Regards, 

_Alberto_

Coduz:
Hi all, 

this PR fixes #177 

**This PR is not yet completed.**

Regards,

_Alberto_


Coduz:
Hi all, 

This PR is now completed and ready to be reviewed. 
It is rebased also onto develop few hours ago. 
Ask if you want more details about it! 😄 

Regards,

_Alberto_

Coduz:
Hi all, 

Nevermind  I need two PR merged before.

Regards, 

_Alberto_

Coduz:
I will close this as is it no longer needed for now. 
We may return on this topic later

PR #201, opened 2017-01-20
Basic unit tests for message module

muros-ct:
Code coverage of methods that provide data access to KapuaMessage and connected types.

This tests provide safe-net for possible refactoring and changes.

Corner cases are not covered as implementation does not check for those.

Also added marshalling for KapuaPosition.

Finished tests in xml package. KapuaMetrics test suite. It has 100% coverage.

Added tests for device data part of messages.

Added tests for Lifecycle messages.

All tests are also gruped into test suites.

### DIFF
--- a/message/internal/pom.xml
+++ b/message/internal/pom.xml
@@ -39,10 +39,33 @@
             <artifactId>validation-api</artifactId>
         </dependency>
 
+        <!-- Test dependencies -->
+        <dependency>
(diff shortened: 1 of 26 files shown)


muros-ct:
I see failed tests. I am fixing it.

muros-ct:
Finally!  Problem was with different timezones in connection with testing validity of xml timestamps and dates compared with test results.

PR #202, opened 2017-01-24
Device Event service CRUD tests

andrejn-ct:
## CRUD tests for the Device Event service
This CRUD part of the Device Events tests exercise the majority of the Device Event service code (package org.eclipse.kapua.service.device.registry.event.internal).

There is no change in the pom files.

### Implementation changes
The implementation of the Device Event service has not been changed.

### Test changes
A new set of Cucumber based tests is implemented to test CRUD operations on the Device Event service.

### Misc changes
Some additional cleanup was added to the Device Registry test suite to prevent database collisions (creating already existing tables) between the various device  registry service tests (Device Registry, Device Connection, Device Event).
A test step that was mistakenly committed for the Device Connection service was also removed.


andrejn-ct:
The year in the header should be now correct. Please let me know if some additional squashing or other is needed.

stefanomorson:
Even the feature file ... :-)

PR #226, opened 2017-02-02
Fix docker generation a bit

ctron:
This change implements the following changes:

 * Fix name of docker images
 * Add missing docker containers in start list
 * Link docker containers to H2 container
 * Switch from java:8u40 to java:8 container image
 * Fix maven warnings

----

I stumbled over those issues while trying to get Kapua up and running locally. Those changes fix the startup of Kapua. But there is still the issue of a few columns and table missing (e.g. USR_USER.EXTERNAL_ID)

### DIFF
--- a/assembly/pom.xml
+++ b/assembly/pom.xml
@@ -71,7 +71,6 @@
 		<dependency>
 			<groupId>org.eclipse.kapua</groupId>
 			<artifactId>kapua-message-internal</artifactId>
-			<version>${project.version}</version>
 		</dependency>
 		<dependency>
 			<groupId>org.eclipse.kapua</groupId>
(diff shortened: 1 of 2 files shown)


hekonsek:
Hi Jens,

Actually IMHO we should narrow version of Java8 as much as possible to be sure that Docker build is reproducible. Otherwise build results will be different depending which Java8 Docker image is available in a local Docker registry of the machine used to build an image.

I propose to close this PR and split it into smaller ones. Is that OK with you?

ctron:
Created:
  * https://github.com/eclipse/kapua/pull/235
  * https://github.com/eclipse/kapua/pull/236
  * https://github.com/eclipse/kapua/pull/237
  * https://github.com/eclipse/kapua/pull/238

hekonsek:
Thanks, Jens. I will take a look at the other PRs.

PR #227, opened 2017-02-02
Fix copyright vs TM

ctron:
The correct label is "Eclipse Kura™"


hekonsek:
Agreed. Thanks a million, Jens!

PR #237, opened 2017-02-07
Allow to automatically configure database link with container links

ctron:
This change uses the container link alias "db" for connecting to the
Kapua database.

### DIFF
--- a/assembly/pom.xml
+++ b/assembly/pom.xml
@@ -465,8 +465,11 @@
 									<descriptor>/src/main/descriptors/kapua-broker.xml</descriptor>
 								</assembly>
 								<entryPoint>
-								   <shell>/maven/bin/activemq console</shell>
+									<shell>/maven/bin/activemq console</shell>
 								</entryPoint>
+								<env>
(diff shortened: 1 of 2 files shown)


hekonsek:
Hi Jens,

Actually, we resolve this on the application level [1], so I would avoid hardcoding this information to container image.

What do you think? 

[1] https://github.com/eclipse/kapua/blob/7d6ec938886bf63a4a3c64e7ec405ff16f6662c3/commons/src/main/java/org/eclipse/kapua/commons/jpa/DefaultConfigurableJdbcConnectionUrlResolver.java#L37

ctron:
I hadn't found that. So we could re-write this to:

```xml
<env>
   <SQL_SERVICE_HOST>$DB_PORT_3306_TCP_ADDR</SQL_SERVICE_HOST>
</env>
```

And still let docker to the linking.

hekonsek:
Just to be sore that I understand your intention here :) You need this for pure-Docker deployments or OpenShift deployments?

ctron:
This is done for pure docker deployments. Using the `--link` option.

ctron:
Ok .. I just figured out that you are actually calling ".getenv". So a redirect to another environment variable is not possible. If the variable would be processed by Commons Configuration it would be possible to use `${env:DB_PORT_3306_TCP_ADDR}`, which would perform this re-lookup.

So I guess it is best to stick with the original approach.

ctron:
I added the same for the broker container

hekonsek:
OK, many thanks!

PR #238, opened 2017-02-07
Switch to the most recent version of Java 8

ctron:


hekonsek:
Now it is fine :) Explicit versioning of Docker images FTW!

PR #239, opened 2017-02-07
Impl join device connections

### LINKED ISSUE #68: Add support for join entities in the same service
Hi all,

we need to add the support of join in our Query wrapper over JPA. 

Currently queries are restricted to one entity domain a time. 
This is fine when a service has only one domain, but can limit querying flexibility that we are trying to pursue. 

An example is the DeviceRegistryService. 
It has 3 domains:
- Device
- DeviceConnection
- DeviceEvent

It is fine for queries on single domains, but what if I need to find all connected devices? That would require two domains (Device and DeviceConnection) to be joined and filtered at the same time. 

Regards,

_Alberto_

Coduz:
Hi all, 

this PR closes #68 
The branch has just been rebased onto develop and it is available to review.

Regards, 

_Alberto_


Coduz:
Rebased onto last develop

Coduz:
Rebased onto develop and added a small change to the behaviour of the query method.
Fixed also some copyright headers and minor doc changes

Coduz:
Rebased again. If it passes the build, It will be merged to develop

PR #242, opened 2017-02-07
Improvements for OpenShift deployment

ctron:
##  Re-build docker file based on Fedora 25

The dockerfile was changed primarily to be able to run on OpenShift as a batch job.

The reason to change it to Fedora was due to the fact that java:8 images are rather unstable, dependencies where missing from upstream compared to the provided image and thus the dockerbuild actually re-installed java 7 during the installation of Maven.

##  Improve minor things deploying to OpenShift

* Test first if the project already exists
* Set a readyness probe for the database
* Start liquibase as an Openshift job in the same project in order to
  access the database
* Finally expose the kapua-console service


hekonsek:
+1 for merge

Also keep in mind that we will remove Liquibase image soon, on the behalf of the embedded Liquibase client

PR #250, opened 2017-02-08
Details pages for devices added

ct-mtomic:
Signed-off-by: HSL\mtomic <marko.tomic@comtrade.com>


console-v2/ui/src/devices/controllers/DeviceDetailBundlesCtrl.ts
```diff
@@ -0,0 +1,20 @@
+/*******************************************************************************
+* Copyright (c) 2011, 2016 Eurotech and/or its affiliates
```
ctron: This should be 2017 only. As those files probably are not from 2011.

PR #260, opened 2017-02-08
Re-implement translator lookup using straight forward iterable

ctron:
This change re-implements the lookup logic by simply relying on the
Iterable interface of the ServiceLocator.


hekonsek:
Many thanks, Jens!

PR #261, opened 2017-02-08
Provide a common base class for most kapua to kura translators

ctron:
This change does provide a common translator implementation which
reduces the amount of duplicate code. All but one translators could
be covered by this approach.


ctron:
I split up the PR in two commits. One for fixing the code format and the second containing the actual changes.

PR #270, opened 2017-02-09
Created Liquibase client

hekonsek:
Created Liquibase client which will be used to:
- update SQL schema in unit tests
- update schema when services start

Client scans classpath for occurences of log files.

@dejanb @ctron @stefanomorson What do you think?

### DIFF
--- a/service/liquibase/pom.xml
+++ b/service/liquibase/pom.xml
@@ -1,0 +1,48 @@
+<?xml version="1.0"?>
+<!--
+    Copyright (c) 2011, 2016 Red Hat and/or its affiliates and others
+   
+    All rights reserved. This program and the accompanying materials
+    are made available under the terms of the Eclipse Public License v1.0
+    which accompanies this distribution, and is available at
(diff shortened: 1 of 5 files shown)


Coduz:
Hi @hekonsek ,

I looked at the code and as far I understood, it refers to the file "_liquibase.sql_" and execute it right? 

Regards, 

_Alberto_

ctron:
Looks pretty simple. :+1:

dejanb:
Looks good

hekonsek:
@Coduz Yes. This is simple classpath scanner which loads scripts into Liquibase client.

Is it OK to merge it and start using in unit tests for a beginning?

Coduz:
Yes that's fine.. I was concerned about the statically linked file "liquibase.sql" but yesterday you said that adapting the existing SQLs will be done in another phase

hekonsek:
Yes. Also keep in mind that this is a test file only. We will provide dedicated "liquibase.sql" file for each service.

hekonsek:
Merged! Many thanks for reviews.

Issue #279, opened 2017-02-17
Missing APP_NAME inTranslatorAppConfigurationKapuaKura


muros-ct:
Before committing device registry integration tests, I rebased to develop and tests are failing.
I found out that there was change to TranslatorAppConfigurationKapuaKura, where
 
`propertiesDictionary.put(DeviceConfigurationAppProperties.APP_NAME, ConfigurationMetrics.APP_ID);`
  
was removed from static initializer.

This APP_NAME is used in translateChannel method, just couple of lines further that class and it causes NPE.

Coduz:
Yes, this bug has introduced by the refactoring of @ctron .

Coduz:
Thanks for the fix @muros-ct

PR #280, opened 2017-02-17
Resloving issue with translateChannel NPE (#279)

muros-ct:
APP_NAME was removed from static initialiser and was causing NPE in translateChannel method.

@ctron Please check if fix is appropriate


Coduz:
Hi @muros-ct ,

thanks for the fix. 
This PR fixes 279.

Regards,

_Alberto_

Issue #265, opened 2017-02-09
Kapua broker should come with websocket jar included


hekonsek:
Hi

One of our users requested websocket support for Kapua broker. I think this is reasonable request. IMHO we should include activemq-http jar in our distribution by default and provide a way to enable WS endpoint via environment variable in Docker image.

@dejanb @ctron What do you think, guys?

Cheers!

Coduz:
I'll add @riccardomodanese also in the discussion

hekonsek:
@dejanb is it possible to use ENV variables to enable/disable transport in standalone ActiveMQ distribution? It is certainly possible in embedded ActiveMQ, where you just can add `if` statement :) .

sheaphillips:
@hekonsek is the Dockerfile used to create the kapua broker image available somewhere?  We are aiming to deploy Kapua in OpenShift and hoping to enable websockets in ActiveMQ as we don't want to open ports other than 80 and 443.  One thought was to extend the kapua-broker Dockerfile to add the additional ActiveMQ dependencies for websockets. We'd rather no create a kapua-broker Docker distro from scratch...would rather build on what you've done and contribute back...

ccustine:
@sheaphillips The Dockerfiles are mostly generated from the maven pom.xml in the assembly subproject. There are images for the console, broker, api, etc. For example, here is the console: https://github.com/eclipse/kapua/blob/develop/assembly/pom.xml#L481-L503

You can also find some examples of running on OpenShift here: https://github.com/eclipse/kapua/tree/develop/dev-tools/src/main/openshift

Some of us are working out minor issues with this at the moment but it is nearly 100% there.

sheaphillips:
@ccustine thanks!  that's really helpful.   I think I pretty much see how things are pieced together - may be a little tricky to figure out all the transient deps for activemq-http...

On the openshift side, we've been using the scripts provided - also very helpful!  Our hangup is mainly getting websockets enabled for the broker at this stage.

jamesfalkner:
I created a derived dockerfile from kapua and added in the websockets dependencies (it's way more than the activemq-http.jar file!). I'll post it here in about an hour when I return to my desk..

sheaphillips:
@jamesfalkner terrific! I'm sure that'll save us a lot of futzing around.

jamesfalkner:
[...]

PR #286, opened 2017-02-17
#265 - enable ws connector by default

dejanb:


jamesfalkner:
Can you also expose the websockets port (61614) at https://github.com/eclipse/kapua/blob/develop/assembly/pom.xml#L473 ?

hekonsek:
BTW I think that we should try (#314) to remove this ConnectorDescriptor thing and just copy name of the connector used into message header.

PR #287, opened 2017-02-17
A set of OpenShift tweaks

ctron:
This PR provides a few tweaks for OpenShift:
* Added request timeouts and increased the initial timeout for the broker instance
* Use cluster IPs instead of `localhost`
* Add some documentation about entropy


ctron:
I added a few more tweaks and updated the initial description of the PR.

Issue #277, opened 2017-02-16
Use non-root users for docker images


ctron:
Docker containers allow the use of non-root users, which is especially important when running inside of OpenShift. However our images (as many others) don't take care of this.

I would suggest to:
* Rebase the provided docker images to be based on `java:8-alpine` which is only ~40MB for the whole image
* Create a new user inside this image (with `adduser`)
* Use this dedicated user inside this image for running

This way we can prepare the docker containers during build time to use this user and won't run into any permission issues later on.

hekonsek:
+1

mcarrer:
Closed

PR #288, opened 2017-02-20
Shrink docker images and use non-root users

ctron:
This change fixes issue #277 by using a dedicated "kapua" user. Permissions are also set to world as suggested by the [OpenShift Guidelines](https://docs.openshift.com/enterprise/3.0/creating_images/guidelines.html#use-uid).

Images are now based on the `openjdk:8-alpine` image series (locked to `8u111`) and file system operations are collapsed into a single run command so that e.g. the image size the `kapua-console` image was reduced to 192 MB from 476 MB.

Additionally a few Maven warnings were fixed.

### DIFF
--- a/assembly/pom.xml
+++ b/assembly/pom.xml
@@ -575,6 +575,13 @@
 									<port>61614</port>
 								</ports>
 								<user>kapua</user>
+								<runCmds>
+									<run><![CDATA[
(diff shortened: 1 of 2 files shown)


ctron:
@hekonsek @stefanomorson Any objections on merging this PR?

hekonsek:
Merged. Thanks!

PR #316, opened 2017-02-22
Fix code style of MessageType enum

ctron:
All uppercase is used either for constants or enum values. Enum types should follow the schema of all types (classes, interfaces).

Kupua does this, except for the `MESSAGE_TYPE` enum, so when you look through the code `Map<MESSAGE_TYPE,?>` lets you think that `MESSAGE_TYPE` is a generic type argument, where in fact it is an enum type.


lorthirk:
Hi @ctron,

It seems there's some conflict in `JmsUtil.java`. Could you please address it?

Thanks!

ctron:
Done. Thanks for all you effort!

Issue #310, opened 2017-02-22
WebSockets fails with java.lang.NoClassDefFoundError: org/apache/activemq/transport/ws/jetty9/StompSocket


jamesfalkner:
Trying websockets in the latest kapua results in some NCDFE's and CNFE's. I found that by adding `lib/optional/activemq-stomp-5.14.0.jar` to the kapua distro it works, so I think this needs to be included in order to complete the Stomp-over-WebSockets support.

dejanb:
Hi James, can you tell us a bit more about your use case? Where are you using Stomp?

mcarrer:
Closed due to inactivity.

PR #320, opened 2017-02-22
#310 include activemq-stomp JAR

jamesfalkner:
Signed-off-by: jamesfalkner <schtool@gmail.com>


ctron:
I think the plan is to with AMQP and MQTT. If STOMP is needed we need to create a derived broker image.

If this is not sufficient we need to re-iterate over this issue and can re-open it.

PR #321, opened 2017-02-22
Integration tests with embeded H2 and Broker and emulated Kura

muros-ct:
Preparation of integration tests with multiple live services.

Completely isolated test with embedded broker, mocked Kura device,
embeded H2 database.
Cucumber feature scenarios execute basic steps of registering device
and Kapua calling mocked device with basic operations like listing
packages, bundles, configuration and executing simple ls command on Kura
device.

Added system configuration of embeded H2 database.
This has to be solved somehow with liquibase.

Introduction of injected between step variables in hash map


qa/src/test/java/org/eclipse/kapua/service/StepData.java
```diff
@@ -0,0 +1,45 @@
+/*******************************************************************************
+ * Copyright (c) 2011, 2017 Eurotech and/or its affiliates and others
```
ctron: I know this is only a minor thing ... but this file seems to be created in 2017 and thus the correct copyright header line would be "Copyright (c) 2017 Eurotech…"

qa/src/test/java/org/eclipse/kapua/service/device/steps/BrokerSteps.java
```diff
@@ -0,0 +1,280 @@
...
+        deviceCommandManagementService = locator.getService(DeviceCommandManagementService.class);
+        deviceCommandFactory = locator.getFactory(DeviceCommandFactory.class);
+
+        /*
+         * Initialize broker from MQ configuration file ad start it.
+         * After issuing start wait a moment to start completely.
+         */
+        broker = BrokerFactory.createBroker(new URI("xbean:activemq.xml"));
+        broker.start();
+        Thread.sleep(BROKER_STARTUP_WAIT_MILIS);
```
ctron: I would suggest to use something like:

```
    private static void waitForPort(int port, Duration duration) throws InterruptedException {
        final Instant end = Instant.now().plus(duration);
        
        System.out.format ("Waiting for port %s ... %n", port);
        while ( true )  {
            try ( Socket socket = new Socket((String)null, port) ) {
                System.out.format ("Waiting for port %s ... done!%n", port);
                return;
            } catch (UnknownHostException e) {
                throw new RuntimeException ( e) ;
            } catch (IOException e) {
                // re-try
            }

            if ( Instant.now().isAfter(end) ) {
                throw new IllegalStateException(String.format ("Port %s did not open after %s seconds", duration.getSeconds()) );
            }
            
            Thread.sleep(1_000);
        }
    }
```

To wait for the broker to be ready, instead of a fixed timeout.

qa/src/test/java/org/eclipse/kapua/service/StepData.java
muros-ct: I never know what exactely to put in Copiright. I have a templating, I have copyright plugin, but still it is about content.

Why is there 2011 at all? I just reused from other file. Is it there as the year when the file was created.
When is the year changed? Should it be changed in all files every year? Probably not.
It is probably only when file is changed by content.

In this case that you mentioned, it is probably as you say, just 2017 and no 2011.

qa/src/test/resources/credentials.properties
```diff
@@ -0,0 +1,22 @@
+## ---------------------------------------------------------------------------
```
ctron: Could you just re-write the file and create it under the EPL license. Kapua itself cannot create content under the APL2.

qa/src/test/java/org/eclipse/kapua/service/StepData.java
ctron: I know most people just copy it from other files. :wink: 

There is the Eclipse project handbook, which might be helpful in such situations [1] and specifically a section about copyright headers [2]

[1] https://www.eclipse.org/projects/handbook/
[2] https://www.eclipse.org/projects/handbook/#ip-copyright-headers

qa/src/test/java/org/eclipse/kapua/service/device/steps/BrokerSteps.java
muros-ct: You are right. My is too optimistic.
Will do that.

qa/src/test/resources/credentials.properties
muros-ct: Shure. As the conten is copied from Active MQ.
Do you think if I add Kapua copyright and keep content as it is, that will be ok. 
Because content is as it is.

ctron: I know ... personally I don't think that a file like this should have a copyright header ... but you can do this .. if the first comment is written in your own words, then that should be fine.

qa/src/test/java/org/eclipse/kapua/service/StepData.java
muros-ct: So, as I understand it would be enough to just state the year when the file was introduced.
But it looks like in our case there is a year when file was created and the second year is year of change.

ctron: So you need to state the year of creation and the year of the last edit. If both are the same then it is simple one year. So either `2017` or `2016, 2017`.

muros-ct:
@Coduz can you check this principle of integration testing with embedded resources (broker, db, emulated kura).
If principle is ok, I can start reproducint tests for other functionality.

muros-ct:
I did rebase and my tests are failing?
I have to investigate.
So this PR is on hold.

muros-ct:
Ok, it seams that now it is ok.
@Coduz do you think that we can merge that or should I introduce liquibase into tests first?

Coduz:
Hi @muros-ct ,

sorry for the delay in the review of this PR.
I've tried first to go through the code and it seems that there is a mix of changes from you and changes from develop. 
You pulled changes from develop in your branch right?

Regards, 

_Alberto_

muros-ct:
Hi,

I have been fighting with rebase today. I just managed to rebase to develop.
Now I am running build and tests to see if everything is ok.
There was an issue with create_all.sql that I chaned and it was also removed. I have to see if tests are still ok.

I will report back as soon as I am sure it works ☺

Uros

muros-ct:
Yes, rebase was ok.
But tests are failing with new liquibase functionality.
Trying to find the cause. I suspect that there is a problem with embedded broker that also uses database.

muros-ct:
Can anyone explain what exactly this error means?

17:25:06,066 [main] INFO  TransportConnector:263  - Connector mqtt started
17:25:10,412 [Camel (camelContext) thread #5 - JmsConsumer[Consumer.eurotech:mainRoutelisteners:EXACTLY_ONCE.VirtualTopic.%3E]] ERROR DefaultJmsMessageListenerContainer:931  - Could not refresh JMS Connection for destination 'Consumer.eurotech:mainRoutelisteners:EXACTLY_ONCE.VirtualTopic.>' - retrying using FixedBackOff{interval=5000, currentAttempts=2, maxAttempts=unlimited}. Cause: Could not create Transport. Reason: java.io.IOException: Server TransportAcceptListener is null.

Since last rebase of my pull request for integration tests this error fails all the tests?
Help is appreciated ☺

Uros

Coduz:
Hi @muros-ct 

Unfortunately I don't know from where that error comes from.
I tried to ping @riccardomodanese  on that but he also does not have a solution.

Maybe @dejanb can help?

Regards, 

_Alberto_

muros-ct:
I think the tests are ready now. If tests will fail, code should be checked.
I implemented Dejan changes to broker configuration, that helped a lot with broker issues.

Code is rebased to develop and builds and runs successfully. If tests will fail, again, it might be due to broker startup time.

@Coduz please review and merge to develop, as it might help with further development and testing of broker and device lifecycle.

lorthirk:
This PR needs to go through the CQ process since it's bigger than 1000 lines of code and it's not coming from a project committer

lorthirk:
Hi @muros-ct,

It seems there are some conflicts in your branch. I'd need you to fix them before I can proceed with the CQ.

Thanks!

muros-ct:
Broker startup has issue of messages being lost before broker starts.
Maybe this is only issue in integration tests.

lorthirk:
CQ created

ctron:
Nope:
> On IPZilla it says: Only Eclipse committers can use IPZilla.
> Is there any way to track this CQ.

PR #331, opened 2017-02-28
Add a security scan for dependencies based on OWASP Dependency Check

ctron:
This change does add a security scan of used artifacts using the OWASP Dependency Check [1].

In order to activate this check a top level project has to be built, enabling the profile `security-scan` with `-Psecurity-scan` (which is disabled by default). The reason why this is disabled by default is because the scans are rather time consuming.

The final output is an HTML file located at `kapua.git/target/dependency-check-report.html`.

[1] https://www.owasp.org/index.php/OWASP_Dependency_Check
[2] https://jeremylong.github.io/DependencyCheck/index.html

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -617,6 +617,31 @@
                 <module>dev-tools</module>
             </modules>
         </profile>
+        <profile>
+            <id>security-scan</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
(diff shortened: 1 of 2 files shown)


ctron:
I would also like to enable this profile by default on the Eclipse Hudson instance and publish the results.

ctron:
This is the summary of the current build: [dependency-check-vulnerability.pdf](https://github.com/eclipse/kapua/files/807340/dependency-check-vulnerability.pdf)

PR #333, opened 2017-02-28
Change version of commons-fileupload to 1.3.2

ctron:


Coduz:
Can you please link the CQ?
So it will be easy if we need to track the status of a dependency?

Thanks, 

_Alberto_

ctron:
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=12792

PR #336, opened 2017-02-28
Fix the way the Kura/ESF version is being transmitted

ctron:
This change does implement a common way to transmit the Kura/ESF version information. Falling back to defaults for ESF and Kura.

The change also fixes the storage layer and the UI to store and show the information.


ctron:
The PR in Kura was merged eclipse/kura#1136, any objections on merging this?

PR #342, opened 2017-03-01
Write a logging adapter for Liquibase to forward to SLF4J

ctron:
This adds a logging adapter for Liquibase forwarding the output to SLF4J so that the final application can forward it to either logging framework supported by SLF4J.


ctron:
Ok ...enabling the logging did let this job run into the 4MB travis limit

Issue #368, opened 2017-03-09
Broker and Camel routes synchronization


dejanb:
The current way of starting the camel route before the broker creates a problem as Liquibase job takes time to start, so the routes don't initialize.

We need to do this in proper order: start a broker and then start routes. As I understood, the concern with this approach is that some messages can get lost until Camel is initialized. That's true only for QoS 0 messages (which we don't guarantee deliver for). But even that could be fixed with Subscription recovery policy if needed - http://activemq.apache.org/subscription-recovery-policy.html

muros-ct:
I also have this kind of problems in integration tests where I start embedded broker.
Some times tests pass, other times fail. I suspect that messages are lost.
I introduced some timeouts in tests, but it is "ugly" and it doesn't help always.

dejanb:
@muros-ct Can you test changes in the PR #369 and see if they help. If not, please point me to the tests, so I can check it out.

PR #369, opened 2017-03-09
#368 - refactor activemq.xml and start broker before camel routes

dejanb:


dejanb:
Thanks Riccardo!

muros-ct:
I used changes in qa integration tests and it also works for me.

Issue #374, opened 2017-03-13
Console fails to start because of jetty-gzip


lorthirk:
It seems that after merging 0c8d1a912f6ce4fac6b516a6390234a5e4077084 the console is failing to start because of jetty-gzip:

```
10:31:52,777 [main] INFO  EnvironmentLoader:128  - Starting Shiro environment initialization.
10:31:52,914 [main] INFO  AuthorizingRealm:248  - No cache or cacheManager properties have been set.  Authorization cache cannot be obtained.
10:31:52,914 [main] INFO  IniSecurityManagerFactory:112  - Realms have been explicitly set on the SecurityManager instance - auto-setting of realms will not occur.
10:31:52,916 [main] INFO  EnvironmentLoader:141  - Shiro environment initialized in 135 ms.
Starting Jetty on port 8889
   [WARN] failed jetty-gzip: java.lang.NoClassDefFoundError: javax/servlet/AsyncListener
   [WARN] failed com.google.gwt.dev.shell.jetty.JettyLauncher$WebAppContextWithReload@22b53226{/,/Users/claudio.mezzasalma/Projects/sources/eclipse/kapua/console/target/console}: java.lang.NoClassDefFoundError: javax/servlet/AsyncListener
   [WARN] failed RequestLogHandler@1fcb4808: java.lang.NoClassDefFoundError: javax/servlet/AsyncListener
   [WARN] Error starting handlers
java.lang.NoClassDefFoundError: javax/servlet/AsyncListener
	at java.lang.Class.getDeclaredConstructors0(Native Method)
	at java.lang.Class.privateGetDeclaredConstructors(Class.java:2671)
	at java.lang.Class.getConstructor0(Class.java:3075)
	at java.lang.Class.newInstance(Class.java:412)
	at org.mortbay.jetty.servlet.Holder.newInstance(Holder.java:153)
	at org.mortbay.jetty.servlet.FilterHolder.doStart(FilterHolder.java:92)
	at org.mortbay.component.AbstractLifeCycle.start(AbstractLifeCycle.java:50)
	at org.mortbay.jetty.servlet.ServletHandler.initialize(ServletHandler.java:593)
	at org.mortbay.jetty.servlet.Context.startContext(Context.java:140)
	at org.mortbay.jetty.webapp.WebAppContext.startContext(WebAppContext.java:1220)
	at org.mortbay.jetty.handler.ContextHandler.doStart(ContextHandler.java:513)
[...]

PR #376, opened 2017-03-13
Drop mandatory gzip compression

ctron:
This PR drops the gzip compression servlet filter.

1) The servlet is deprected and missing in most recent Jetty versions
2) When using HTTPS (and Kapua should be used over HTTPS) then gzip
compression might trigger the BREACH exploit
3) If compression is still wanted, it should be enabled in the
front-facing HTTP engine
4) It fixes an issue with the out date GWT plugin

[1] https://en.wikipedia.org/wiki/BREACH_%28security_exploit%29

### DIFF
--- a/console/pom.xml
+++ b/console/pom.xml
@@ -208,12 +208,6 @@
             <version>0.9</version>
         </dependency>
 
-        <dependency>
-            <groupId>org.eclipse.jetty</groupId>
-            <artifactId>jetty-servlets</artifactId>
-            <version>9.1.6.v20160112</version>
-        </dependency>
-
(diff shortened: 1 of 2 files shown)


ctron:
Fixes #374

lorthirk:
I opened an [issue](https://bugs.eclipse.org/bugs/show_bug.cgi?id=513625) on Bugzilla for further discussions on this item

lorthirk:
The followup issue is tracked [here](https://bugs.eclipse.org/bugs/show_bug.cgi?id=513625). Any input is really appreciated.

PR #379, opened 2017-03-15
Fix configuration of ElasticSearch transports

ctron:
This change fixes a few bugs in the ElasticSearch transport
configuration. It allows IPv6 addresses to be used, allows the
commons-configuration lists to be used, actually implements the
use of multiple nodes.

This change also provides a unit test with test coverage over 90%.


ccustine:
I have local fixes for setting the cluster name on the ES container and corresponding changes to the broker in the template (es nodes, ports, etc), but I am sure from looking at this PR that you already have those too. Let me know if you need any of that info to finish this off.

PR #383, opened 2017-03-16
CertificateService first implementation

lorthirk:
First implementation for CertificateService. Currently it only handles JWT private and public keys, thus replacing `RSAUtils` who served the same purpose before.


service/security/shiro/src/main/resources/kapua-authentication-setting.properties
```diff
@@ -29,4 +29,7 @@ authentication.credential.apiKey.pre.length=8
 authentication.credential.apiKey.pre.separator=:
 authentication.credential.apiKey.key.length=24
 authentication.credential.apiKey.cache.enabled=true
-authentication.credential.apiKey.cache.ttl=300000
\ No newline at end of file
+authentication.credential.apiKey.cache.ttl=300000
+
+authentication.jwk.private.key=
```
Coduz: Can you please change this two new properties form `authentication.jwt.*` to `authentication.session.jwt.*` and move them around line 18?
Just to keep similar properties near :)

lorthirk: Done!

PR #392, opened 2017-03-20
Click on item, wrong  parameter, accordion expanding

ct-mtomic:
Click on item from list, wrong last event type parameter, accordion expanding

Signed-off-by: HSL\mtomic <marko.tomic@comtrade.com>


ct-mtomic:
Yes, i will fix before merging.

PR #394, opened 2017-03-21
Fix Swagger API docs

ccustine:
Signed-off-by: Chris Custine <ccustine@apache.org>

### DIFF
--- a/rest-api/src/main/webapp/WEB-INF/web.xml
+++ b/rest-api/src/main/webapp/WEB-INF/web.xml
@@ -52,7 +52,7 @@
     </servlet>
     <servlet-mapping>
         <servlet-name>Jersey Web Application</servlet-name>
-        <url-pattern>/v1/*</url-pattern>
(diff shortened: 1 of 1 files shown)


lorthirk:
Hi @ccustine,

Thanks for the PR, I was just going to look into this.

Of course the servlet root and the related Swagger configuration should point to the same location; however, maybe we should make them point to the root of the web server? I just tried to deploy the WAR file in my local Tomcat via the manager application, and having the servlet being served at /api would result in a final URL like `http://localhost:8080/api/api/v1`. 

I'll work on an alternative PR trying to cover both the case of a "contextualized" deployment (http://localhost:8080/api/...) and a root deployment, which seems to be the current Docker approach.

lorthirk:
@ccustine, could you please try #396 and see if it covers your case?

lorthirk:
Closing because #396 has been already approved and merged

PR #396, opened 2017-03-21
SwaggerUI improvements for base path detection

lorthirk:
This PR is an alternative to #394, enabling SwaggerUI to dynamically understand its context (whether is being served on root or on a specific path) and thus correctly locating the swagger.json file and addressing HTTP requests in the test client


lorthirk:
Also @hekonsek this should address your issues about the Swagger URLs

hekonsek:
I will give it a try, thanks!

hekonsek:
Works like a charm! Many thanks Claudio!

PR #403, opened 2017-03-22
Upgrade Apache POI to 3.9 for re-using CQ

ctron:
With this PR we could piggyback on CQ7691


ctron:
Yes, that test fails every now and then. I guess someone should fix that.

I restarted the Travis build!

Coduz:
I suppose that is something related to the refresh time of ES...

PR #405, opened 2017-03-23
Datastore REST API resources

Coduz:
Hi all, 

this is the PR for introducing datastore REST API resources.

They will mount in:
`{host}/v1/data/`

This PR is missing updated header copyrights, code formatting and code clean up, which i'm going to do, but the overall design should not change.

Swagger API will be added in a separate PR.

Regards, 

_Alberto_


Coduz:
Just rebased onto latest develop

Coduz:
This PR should be ready! 
I've added also the swagger UI annotations!

PR #407, opened 2017-03-23
Corrections added to the name of containers

TristanCP:
-jetty for jetty containers, no extension for tomcat containers, like it was meant to be.

### DIFF
--- a/assembly/README.md
+++ b/assembly/README.md
@@ -42,7 +42,9 @@ Navigate your browser to http://localhost:8080 and log in using the following cr
 
 It is also possible to use Tomcat as a web container. For this use the following run commands instead:
 
-    docker run -td --name kapua-console-jetty --link kapua-sql:db --link kapua-broker:broker -p 8080:8080 kapua/kapua-console
-    docker run -td --name kapua-api-jetty --link kapua-sql:db --link kapua-broker:broker -p 8081:8080 kapua/kapua-api
+    docker run -td --name kapua-console --link kapua-sql:db --link kapua-broker:broker -p 8080:8080 kapua/kapua-console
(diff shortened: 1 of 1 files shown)


ctron:
That was actually on purpose. The names of the containers should stay stable IMHO.

TristanCP:
It was on purpose to call "-jetty" the Tomcat container ? Tell me what I'm missing then

ctron:
Hehe .. no, your are right ... calling tomcat containers "-jetty" is definitely wrong 😉  Sorry for not reading your patch completely!

So could you please:
* Sign the ECA [1], if you need help, just let me know and I will try to guide you through the process
* Amend the PR to simply remove the `-jetty` suffix on all names, so that all names are consistent without `-jetty`

It would be best to do the ECA stuff first and push then, because then the "ip-validation" check will pass.

[1] https://www.eclipse.org/legal/ecafaq.php

TristanCP:
I think the problem was my email was not public in my github profile ("noreply@github.com" shown).
Now I don't know how to submit my commit again to the ip-validation, do I need to do a fake change ?

ctron:
You need to squash both commits and add the "Signed-off-by:" footer entry. But every commit needs this footer.

TristanCP:
It looks like my "Signed-off-by: " footer is ignored.

Here is my comment with footer : 
-------------------------------------------------
Container names corrected

and vital new line added

Signed-off-by: Tristan <trrteamlog@gmail.com>

ctron:
I just fixed the branch and squashed the commits.

Please have a look if that it ok with you so I can merge this.

ctron:
Ok, I had to merge it to prevent conflicts with other PRs.

PR #409, opened 2017-03-24
Set Elasticsearch node address in Docker images

lorthirk:
With the release of Datastore REST APIs we need to override Elasticsearch node address and port in APIs and Console Docker images. Also updated documentation, and modified the property override in the ActiveMQ image since we're not using Elasticsearch cluster yet.

I guess the OpenShift's `kapua-template.yml` should be adapted as well.


ctron:
@lorthirk What is the best way to test of this works?

lorthirk:
@ctron What I did is attach a device running Kapua to the broker, let it send some messages, and then use the Elasticsearch REST APIs to retrieve some data. Unfortunately I have no better solutions...

ctron:
Do you have some `curl` (or something else) statement which I can use to query data?

ctron:
Looks good to me. Although I couldn't really test it up to now.

lorthirk:
To get indexes:
`curl -X GET "http://127.0.0.1:9200/_cat/indices"` 
To get messages:
`curl -X GET "http://127.0.0.1:9200/{INDEX_NAME}/message/_search?size=1000"` 

`{INDEX_NAME}` name must be one returned from the previous call

ctron:
But that queries ES directly, which gets filled by the broker image. So that wouldn't test anything related to these changes?!

lorthirk:
Sorry, you're absolutely right.

What I told you in [my previous comment](https://github.com/eclipse/kapua/pull/409#issuecomment-289463511) was not accurate at all indeed; now that I recall, I tested it using Swagger UI and asking for Data Metrics.

Sorry about that!

ctron:
@hekonsek Any objections merging this?

hekonsek:
@ctron No, everything seems to be OK.

PR #443, opened 2017-04-07
Possible fix to failing datastore tests

muros-ct:
When inserting data tests were waiting for time that was set to the same
value as ES index interval. As indexing takes some time, time between
insert and query should be index interval plus some fixed time.


muros-ct:
I did a small change. When there is 100 inserts, wait is disabled. This caused tests to unnecessary  run much longer.
Now setting is for ES to index on 5 seconds and wait after insert is additional 2 sec. This is time for index to complete before query is made.

It works for me, hope that it works also for mr. Hudson.

PR #445, opened 2017-04-07
Fixed group service tests because of added resource limitations

muros-ct:
Group service tests were updated with resource limitations.

delete_all.sql was also updated not to delete seeded data for group and role service.

Signed-off-by: Uros Mesaric Kunst uros.mesaric-kunst@comtrade.com


muros-ct:
It was not committed properly and tests were failing. Will provide new PR.

PR #448, opened 2017-04-07
Fix of tests of group service because introduction of resource limits

muros-ct:
Group service test are introduced to resource limits by adding step
to gherkin scenarios.


muros-ct:
Didn't work. Will try again!

PR #451, opened 2017-04-08
Service resource limits

lorthirk:
Hi all,

This PR introduces scoped configurations for the following service:

- AccountService: infinite child accounts, child accounts limit
- GroupService: infinite groups, group limit
- RoleService: infinite roles, role limit
- DeviceRegistryService: infinite devices, device limit (both not yet enforced)
- UserService: infinite roles, role limit


commons/src/main/java/org/eclipse/kapua/commons/configuration/AbstractKapuaConfigurableResourceLimitedService.java
```diff
@@ -0,0 +1,100 @@
...
+        super(pid, domain, entityManagerFactory);
+        this.serviceClass = serviceClass;
+        this.factoryClass = factoryClass;
+    }
+
+    @Override
+    protected String validateNewConfigValuesCoherence(KapuaTocd ocd, Map<String, Object> updatedProps, KapuaId scopeId, KapuaId parentId) throws KapuaException {
+        String parentValidation = "";
+        int availableChildEntitiesWithNewConfig = allowedChildEntities(scopeId, updatedProps);
+        if (availableChildEntitiesWithNewConfig < 0) {
```
Coduz: Please change this to throw the exception with a specific error code.

commons/src/main/java/org/eclipse/kapua/commons/configuration/AbstractKapuaConfigurableService.java
```diff
@@ -129,7 +129,7 @@ private void validateConfigurations(String pid, KapuaTocd ocd, Map<String, Objec
                 }
             }
 
-            String logicValidationResult = validateNewConfigValuesCoherence(ocd, updatedProps, scopeId);
+            String logicValidationResult = validateNewConfigValuesCoherence(ocd, updatedProps, scopeId, parentId);
```
Coduz: It seems that here the validateNewConfigValuesCoherence at line 151 if called, and it always returns `""` , so no validation is performed on changing configuration values.

I would move this after the current line 138 so we check values after full configuration validation.

dev-tools/src/main/database/all_delete.sql
```diff
@@ -18,15 +18,15 @@ DELETE FROM atht_access_token;
 
 DELETE FROM athz_role WHERE NOT (scope_id = 1 AND id = 1);
 
-DELETE FROM athz_role_permission WHERE NOT (scope_id = 1 AND id = 1);
+DELETE FROM athz_role_permission WHERE NOT (scope_id = 1 AND role_id = 1 AND id IN (1,2,3,4,5,6,7,8,9,10,11,12,13,14,15));
```
Coduz: Don't know if you @lorthirk  or @muros-ct  can answer about this.

Why there is an `all_delete.sql` script? Should they have migrated to liquibase stuff? And in this case isn't much easier to delete all?

commons/src/main/java/org/eclipse/kapua/commons/configuration/AbstractKapuaConfigurableService.java
lorthirk: Well, some class inheriting `AbstractKapuaConfigurableService` could provide its own validation logic in an overriding method, so it's not guaranteed that such call will always return an empty string. I thought that checking for some service-specific logic first was a good idea, to avoid reading configurations from the database when we already know that the custom logic will fail.

dev-tools/src/main/database/all_delete.sql
muros-ct: I introduced this script. I use it during tests so that database is cleaned of test created data.
Only basic seeded data is left in database. In case of some gherkin scenarios i don't want to create database from scratch using liquibase. It is so sloooow. But it is not meant for enithing more than testing.

But maybe this script should be moved into test module. What do you think?

Coduz: Hi @muros-ct ,

thank you for the clarification. Yes I think src/test/... is a more appropriate folder. 

We can create anoter pull request to do that so we don't need to change this one.

muros-ct: @Coduz I agree. When merged, I will move it to common test module.

Coduz:
Sorry @lorthirk , 

I've pointed you to `AccountServiceImpl` that I thought was ok but then I discovered it was not.
I've fixed it and now is all ok for me 👍  We can merge it

Regards, 

_Alberto_

PR #453, opened 2017-04-10
Possible fix to device broker QA test

muros-ct:
Integration tests with embedded broker are failing, but
just on Hudson CI. For that reson, additional logging was
added and timeout after birth message was set to 5 sec.


ctron:
Let's give it a try

PR #459, opened 2017-04-11
Removed JWT private and public keys

### LINKED ISSUE #458: Remove checked in keys for JWT
The key pair in the repository at https://github.com/eclipse/kapua/tree/develop/service/security/shiro/src/main/resources should be removed.

lorthirk:
Fallback to runtime generation for private and public JWT keys

Fixes #458


docs/user-manual/en/jwt_security.md
```diff
@@ -2,13 +2,13 @@
 
 To leverage JWT security features, a pair of RSA keys (a public one and a private one) must be loaded in Kapua.
 
-## Use default Kapua keys
+## Use random generated keys
 
-A pair of RSA keys is already shipped within Kapua, and can be used **ONLY FOR DEVELOPMENT PURPOSES SINCE THEY ARE PUBLICLY AVAILABLE IN THE KAPUA REPOSITORY**. These keys will be automatically loaded when the system properties `authentication.session.jwt.private.key` and `authentication.session.jwt.public.key` are not set, or if the file entered in said properties cannot be read.
+By default Kapua will look for keys in the path specified by `authentication.session.jwt.private.key` and `authentication.session.jwt.public.key` system properties at startup (see below). If such properties are not set, or if Kapua can't find those files, it will automatically generate new random private and public keys.
```
ctron: I would say that if those files are configured, but not existing then this is a major error! Silently falling back to generated keys should not be done in this case.

lorthirk: Right. I fixed this.

ctron:
Let's merge this once Travis gives its 👌

PR #460, opened 2017-04-11
Allow permission wildcard domain

Coduz:
Hi all, 

this PR is to make assigning of the permissions.
Now it is possible to use * as domain level. 
So you can assign to the default admin role of an account the permission `*:*:{scopeId}:*` instead of having to create a single permission for each domain available.

Regards, 

_Alberto_


Coduz:
ok for me 👍

Issue #457, opened 2017-04-11
Drop SHA-1 support


ctron:
As SHA-1 is dead and insecure [1] and Kapua already has support for bcrypt, I would suggest do simply drop support for SHA1 right away.

[1] https://www.theverge.com/2017/2/23/14712118/google-sha1-collision-broken-web-encryption-shattered

ctron:
Closed by PR #461

PR #461, opened 2017-04-11
#457: Drop SHA1 support, add SHA512 and default to it

ctron:


service/security/shiro/src/main/resources/kapua-crypto-setting.properties
```diff
@@ -8,10 +8,10 @@
...
 # Contributors:
 #     Eurotech - initial API and implementation
-#
+#     Red Hat Inc
 ###############################################################################
 
 crypto.bCrypt.logRounds=12
 
-crypto.sha.algorithm=SHA1
+crypto.sha.algorithm=SHA256
```
lorthirk: Should the value be `SHA512` for the default one?

ctron: I can do this before merging this PR if everything else is ok.

PR #465, opened 2017-04-13
Fix account service entity manager usage

Coduz:
Hi all, 

this PR corrects a small issue on Account service implementation on entityManagerSession usage.
Now it uses the utility classes available.

Regards, 

_Alberto_


Coduz:
Rebased and fixed conflicts

PR #485, opened 2017-04-19
Drop loading skin resources

ctron:
This changes removes the code to load "skin.js" and "skin.css" as
those resources don't exist and there is nothing which could map to it.

On the other hand loading those requests flood the logs with the
following error messages:

```
17:07:01,179 [qtp1152808799-43] ERROR KapuaErrorHandlerServlet:157  - Processed HTTP error! Code: 404 - Request: /console/skin/skin.js - Error: Not Found
17:07:01,179 [qtp1152808799-52] ERROR KapuaErrorHandlerServlet:157  - Processed HTTP error! Code: 404 - Request: /console/skin/skin.css - Error: Not Found
```

### DIFF
--- a/console/src/main/webapp/js/kapuaconsole/console.js
+++ b/console/src/main/webapp/js/kapuaconsole/console.js
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2016 Eurotech and/or its affiliates and others
+ * Copyright (c) 2011, 2017 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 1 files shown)


Coduz:
I saw that you dropped the Skin Servlet 10 days ago in https://github.com/eclipse/kapua/commit/2cb233a066df4b6bc0dad28ae7c6c58e4b30dfe0

Why? Before it was possible to provide a custom CSS file to customize a bit the console without having to touch code.

ctron:
As you can see, this was never referenced. The servlet was never bound. So it didn't work back then either.

lorthirk:
I'm merging this, but I also think we should think about a skin customization feature. I opened #514 to keep track of the implementation.

PR #487, opened 2017-04-20
The Data View stuck when no messages fix

bojanutkovic:
The Data View in the console get stuck when there are no messages stored

### DIFF
--- a/console/src/main/java/org/eclipse/kapua/app/console/client/data/TopicsTable.java
+++ b/console/src/main/java/org/eclipse/kapua/app/console/client/data/TopicsTable.java
@@ -114,7 +114,7 @@ public class TopicsTable extends LayoutContainer {
             public void onFailure(Throwable t) {
                 FailureHandler.handle(t);
                 topicInfoGrid.unmask();
-                Window.alert(MSGS.noDataMessage());
+                Window.alert((String)MSGS.noDataMessage());
             }
         };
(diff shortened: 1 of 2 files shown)


console/src/main/java/org/eclipse/kapua/app/console/client/data/TopicsTable.java
```diff
@@ -114,7 +114,7 @@ public void onSuccess(List<GwtTopic> topics) {
             public void onFailure(Throwable t) {
                 FailureHandler.handle(t);
                 topicInfoGrid.unmask();
-                Window.alert(MSGS.noDataMessage());
+                Window.alert((String)MSGS.noDataMessage());
```
ctron: Can you explain why this is necessary. Because it will be removed again when the Kapua cleanup profile will be applied.

ctron: Ok, it looks like GWT 2.4 is still bound to Java 6. Maybe that is the problem.

ctron: So this works locally for me when not having the explicit cast.

ctron:
Merged in by cherry pick: 99ac29adc192db400f00191d65cf62d25a573062

PR #496, opened 2017-04-21
Redirect to home uri instead of reloading on logout

### LINKED ISSUE #484: Unable to log out when using SSO
Currently, when using the web console with SSO, the query parameter `access_token` triggers the automatic login using SSO. However, then logging out, the query parameter is still present and will automatically re-login the user with the same token. Making it impossible to log out.

lorthirk:
Fixes #484


ctron:
This probably works. But requires for all and every setup to configure this URI. Otherwise no user will be able to log out.  Maybe there is some solution like `request.getContextPath()`.

lorthirk:
Currently the URI has to be configured anyway, otherwise the SSO wouldn't know where to redirect after it completes its authentication process.

I'll see if we can ditch the configuration key altogether using something like `request.getContextPath()`.

PR #521, opened 2017-04-27
Update the java base image to 8u121

ctron:


ctron:
Looks like 131 is not yet available

PR #530, opened 2017-05-02
Switch to a recent jacoco version by default

ctron:
This change switches to Jacoco 0.7.9, which is the most recent version.
Only for the Eclipse Hudson instance we fall back to 0.7.4.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -76,7 +76,7 @@
         <nexus-staging-maven-plugin.version>1.6.5</nexus-staging-maven-plugin.version>
         <sql-maven-plugin.version>1.5</sql-maven-plugin.version>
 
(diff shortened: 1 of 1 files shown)


muros-ct:
Please check if Hudson jacoco reporting is still working after new jacoco version is introduced. I remember I had some issues with that, because output of jacoco .exec files was not right for Hudson plugin.

ctron:
Yes, Hudson needs an older version. But this changes leaves the old version for Hudson and just switches to the most recent version for the rest. It still is always possible to downgrade this version from the command line when running Maven in addition.

ctron:
The Hudson builds seems to be fine after the merge.

PR #543, opened 2017-05-09
Make Kapua re-buildable from scratch

ctron:
Currently the REST API uses two swagger artifacts which are only
available from an external download site. However it should be
possible to re-build Kapua from scratch without the need to resort
to BLOB files which cannot be re-built.

This also speeds up the build process as it removes an additional
maven repository entry.


lorthirk:
Hi @ctron,

Since we had some issues in the past with SwaggerUI JS files, please let me give this a try before merging. Thanks!

ctron:
Sure that is why I created this as a PR. We also have to wait for the CQ.

ctron:
Ok, as the CQ has been approved I will merge this today.

PR #544, opened 2017-05-09
Disable javadoc code formatting inside <pre>

ctron:
This will fix issues where with

    <pre>{@code}</pre>

Which get re-formatted by Eclipse and the `@code` actually gets escaped, breaking the Javadoc comment.


Coduz:
I can see the change and that is fine. 
What are other addition?

ctron:
The came in by writing out the file with Eclipse Neon. So they are the defaults as they are now.

ctron:
Is this ok to be merged?

PR #552, opened 2017-05-10
Chng datastore rest client

riccardomodanese:
Refactoring datastore to allow the module to use different transport connectors.


qa/pom.xml
```diff
@@ -248,6 +260,29 @@
...
+        <dependency>
+	    	<groupId>org.apache.logging.log4j</groupId>
+	    	<artifactId>log4j-api</artifactId>
+	    	<version>${log4j-to-slf4j.version}</version>
+	    	<scope>test</scope>
+		</dependency>
+		<dependency>
+	    	<groupId>org.apache.logging.log4j</groupId>
+	    	<artifactId>log4j-core</artifactId>
+	    	<version>${log4j-to-slf4j.version}</version>
```
ctron: Please stick to slf4j and logback if possible.

qa/src/test/resources/kapua-datastore-transport-client-setting.properties
```diff
@@ -0,0 +1,32 @@
...
+#      - a static getInstance method (that returns the instance)
+datastore.elasticsearch.client.provider=org.eclipse.kapua.service.datastore.client.embedded.EsEmbeddedClientProvider
+#datastore.elasticsearch.client.provider=org.eclipse.kapua.service.datastore.client.transport.EsTransportClientProvider
+
+#timeouts
+datastore.query.timeout=60000
+datastore.scroll.timeout=60000
+
+datastore.elasticsearch.cluster=kapua-datastore
+datastore.elasticsearch.node=192.168.33.10
```
ctron: If possible please don't use any hard-coded IP addresses

service/datastore/client-api/pom.xml
```diff
@@ -0,0 +1,41 @@
+<!--
+    /*******************************************************************************
+     * Copyright (c) 2011, 2017 Eurotech and/or its affiliates
```
ctron: This file is not from 2011! Some for the other files.

service/datastore/client-embedded/src/main/java/org/eclipse/kapua/service/datastore/client/embedded/EsEmbeddedEngine.java
```diff
@@ -0,0 +1,80 @@
...
+ * Elasticsearch embedded node engine
+ * 
+ * @since 1.0
+ *
+ */
+public class EsEmbeddedEngine {
+
+    private static final Logger logger = LoggerFactory.getLogger(EsEmbeddedEngine.class);
+
+    private static final String DEFAULT_DATA_DIRECTORY = "target/elasticsearch/data"+ UUIDs.randomBase64UUID();
```
ctron: Could there be a slash missing at the end?

service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/MessageStoreServiceImpl.java
```diff
@@ -54,31 +56,29 @@
     private final AuthorizationService authorizationService = locator.getService(AuthorizationService.class);
     private final PermissionFactory permissionFactory = locator.getFactory(PermissionFactory.class);
 
-    private final MessageStoreFacade esMessageStoreFacade;
+    private final MessageStoreFacade messageStoreFacade;
 
     /**
-     * Default constructor
+     * Defualt constructor
```
ctron: Typo

assembly/src/main/descriptors/kapua-broker.xml
```diff
@@ -167,7 +167,11 @@
                 <include>org.apache.shiro:shiro-core</include>
 
                 <include>org.elasticsearch:elasticsearch</include>
+                <include>org.elasticsearch.client:transport</include>
                 <include>org.elasticsearch:securesm</include>
+                <include>org.slf4j:slf4j-api</include>
+                <include>org.slf4j:log4j-over-slf4j</include>
+                <include>org.apache.logging.log4j:log4j-to-slf4j</include>
```
muros-ct: I think that with 5.3 and on it is kind of requirement log4j or bridge from it.
https://www.elastic.co/guide/en/elasticsearch/client/java-api/5.3/_using_another_logger.html

service/datastore/client-embedded/src/main/java/org/eclipse/kapua/service/datastore/client/embedded/EsEmbeddedEngine.java
riccardomodanese: I agree, may be a good idea to place a slash at the end!

service/datastore/client-api/pom.xml
riccardomodanese: Sorry!

qa/src/test/resources/kapua-datastore-transport-client-setting.properties
riccardomodanese: The IP is used with the ES transport client connector and it is supposed to be used with the From the default configuration, the qa and the other tests use the ES embedded node, so this IP is not checked at all during the standard build test phase.
Do you suggest to use another default value (in any case this configuration can be overwritten at startup right?)

broker-core/pom.xml
```diff
@@ -142,6 +153,24 @@
...
 			<scope>provided</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.elasticsearch.client</groupId>
+			<artifactId>transport</artifactId>
+			<version>${elasticsearch-client-transport.version}</version>
+			<scope>provided</scope>
+		</dependency>
+		<dependency>
+	    	<groupId>org.apache.logging.log4j</groupId>
```
riccardomodanese: About formatting issues in general I have no idea about these. I have the style correctly loaded in my Eclipse IDE so I supposed the code was formatted properly.

assembly/src/main/descriptors/kapua-broker.xml
riccardomodanese: About logging, from my undestanding there is no alternatives.
Or you give a true log4j implementation to ES (ES throws an exception end exits at startup, if no log4j implementation is found), or you can use the bridge (using slf4j).
I'm quite sure I told about this new dependency in a previous scrum and sent you an email about that.
In any case if someone has a better solution that I didn't figure out I'm glad to change the dependencies.
https://www.elastic.co/guide/en/elasticsearch/client/java-api/current/_using_another_logger.html

service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/MessageStoreServiceImpl.java
riccardomodanese: :(

qa/pom.xml
riccardomodanese: Sorry, I don't understand. Could you explain the correct log dependencies tree should be used?

assembly/src/main/descriptors/kapua-broker.xml
ctron: So if log4j is required, then we could use log4j-over-slf4j, which we already have, right?

broker-core/pom.xml
ctron: So maybe just do it manually then. You can always enable "Show whitespaces" in Eclipse to see what's happening.

qa/pom.xml
ctron: Instead of log4j, log4j-over-slf4j can be used. Which diverts all logging to slf4j and thus allows to use any logging framework later on. See: https://www.slf4j.org/legacy.html

qa/src/test/resources/kapua-datastore-transport-client-setting.properties
ctron: Well I suggest to use some mechanism which can be run locally, on travis, husdon and jenkins.

If that currently works then there is no change required.

If it isn't required, it could be removed, right?

riccardomodanese:
About "I couldn't really figure out how the ES refresh is done now, maybe you could point me towards it"
The new cache implementation does a query, if no entry is found for a key (https://github.com/eclipse/kapua/blob/chng-datastoreRestClient/service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/ClientInfoRegistryFacade.java#L104 and similar code for the other registryes).
In that way the index creation is avoided if the index is present in the datastore but not in the cache.
The current datastore delete is not fully implemented so the index cache may be only incremental.

qa/src/test/resources/kapua-datastore-transport-client-setting.properties
riccardomodanese: It is not required by the standard build since we are using the embedded node but for a test against a real environment it should be specified (or replaced by the node list property).
So I prefer to keep it into the configuration file or at least leave it as comment so it will be easy to configure it without looking at the documentation and/or code to find the proper paramenter name.
Ok, your comment regarding the qa module so if someone wants to execute qa against an external ES he should has a proper code knowledge and he knows which properties should be set.
Do you agree if I leave the paramenter in the file as comment?

ctron:
I was more interested in the index refresh, not the index creation. Currently the unit test for an index refresh before making queries, to ensure that unit tests don't fail in Hudson/Jenkins/Travis.

How is this solved with this PR?

qa/pom.xml
riccardomodanese: Sorry, I reviewed the dependencies and now it's clear you appointment. May I did something wrong during the rebases/merge I did beacuse the log4j core is present in few pom.
Since the log4j core should be replaced by
<dependency>
    <groupId>org.apache.logging.log4j</groupId>
    <artifactId>log4j-to-slf4j</artifactId>
</dependency>
there is no need to have it in the dependencies.
Right?

riccardomodanese:
What we are doing in the test, so forcing the index refresh after store operations, has no sense in a real application scenario or at least it is very dangerous from a performance point of view.
For test purpose, as correctly suggested by you, we added that sync to prevent the test to wait for the index synchronization automatically done by ES.
But in a real world we cannot force a index refresh for every message inserted. If new use cases need to be sure that the inserted data is close to be immediatly available for query we can think to add a new store method. But such method must be used carefully by developers.
In any case I added the index refresh parameter in the configuration (and also the shards and replicas, see the link below), so it can be tuned according to the user needs. From our ES knowledge the default value (5s) could be considered the lower bound.


https://github.com/eclipse/kapua/blob/chng-datastoreRestClient/service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/schema/Schema.java#L231

qa/pom.xml
ctron: log4j should be replaced with org.slf4j:log4j-over-slf4j

qa/src/test/resources/kapua-datastore-transport-client-setting.properties
ctron: The please add some proper description why this IP is configured in the way it is and what a user needs to do in order to make use of this exact IP.

ctron:
I think we all already agreed on the fact that we don't do this in production, but that we have to do this for testing.

So where is that "sync" you mentioned? Could you point me towards this.

riccardomodanese:
The sync is called externally by the test.
The datastore entry point is:
https://github.com/eclipse/kapua/blob/chng-datastoreRestClient/service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/mediator/DatastoreMediator.java#L244

ctron:
That is what I meant. Thanks!

riccardomodanese:
Hi all, I'm asking your support to check the deployment in the Docker/OpenShift/Vagrant demo-box since I noticed missing dependencies, related to the new ES version and client modules I added, in the assembly step.
The new dependencies needed by all modules should be
kapua-datastore-client-api
kapua-datastore-client-transport
org.elasticsearch:elasticsearch
org.elasticsearch.client:transport

ES customization, needs by the new version:
jvm
vm.max_map_count=262144

ES configuration
see https://github.com/eclipse/kapua/blob/chng-datastoreRestClient/dev-tools/src/main/vagrant/develop/develop-Vagrantfile#L77

system (file descriptor up to 64k)
* soft nofile 65536
* hard nofile 65536

Thank you!

riccardomodanese:
Hi Jens, I was investigating about log dependencies and, removing log4j core from the dependency works fine (in the datstore-internal tests) using the external ES, but switching to the embedded node I got a runtime exception from ES transport (java.lang.NoClassDefFoundError: org/apache/logging/log4j/core/config/Configurator)
May I added this for that scope, so may it's not a rebase/merge issue.
If this is confirmed by more deep test, could we keep log4j core in the test dependencies or do you figure out a better solution?
Could we have a call tomorrow to beter focus the issue/solution?

ctron:
So you simply need to add "log4j-over-slf4j" as a `<scope>test</scope>` dependency. No need to log4j.

But sure, let's have a call. Just ping me.

riccardomodanese:
still have the same issue, let's have a call. I ping you by email

ctron:
I think I understand where this is all coming from, this is actually log4j 2, not log4j (1.x).

In this case we need:

```xml
<dependency>
    <groupId>org.apache.logging.log4j</groupId>
    <artifactId>log4j-to-slf4j</artifactId>
    <version>2.7</version>
</dependency>
```

riccardomodanese:
I think we have no alternative
looking at https://github.com/elastic/elasticsearch/issues/19415 ( mvmn commented on 17 Jan )
:(

riccardomodanese:
I already used
org.apache.logging.log4j:log4j-to-slf4j
It works with external es but not with the embedded node
With the embedded node I got this exception:
Exception in thread "elasticsearch[6_kP2vA][clusterService#updateTask][T#13]" java.lang.NoClassDefFoundError: org/apache/logging/log4j/core/config/Configurator

ctron:
I think I do have a solution for the log4j2 issue. I have it running locally, but I need to fix this up.

But in general the situation really is a mess. I think that they really seem to stop supporting the embedded variant at some point, so we should look for alternatives.

riccardomodanese:
yes, I agree we should find an alternative way to execute tests since the embedded node is no more "officially" supported

ctron:
So I did write a small replacement for the log4j components required by ES, you can enable this with:

```xml
<dependency>
    <groupId>de.dentrassi.elasticsearch</groupId>
    <artifactId>log4j2-mock</artifactId>
    <version>0.0.1</version>
    <scope>test</scope>
</dependency>
```
You will still need `org.apache.logging.log4j:log4j-to-slf4j`, as this provides the SLF4J bridge support. With this the unit tests for fine for me. It should only be required in the `kapua-qa` project.

riccardomodanese:
Thank you Jens, I fixed the dependencies and now tests are ok locally, with embedded node and also with the transport connector and external ES.
Can you take a look to my last commit?

service/datastore/client-api/pom.xml
```diff
@@ -0,0 +1,41 @@
+<!--
+    /*******************************************************************************
+     * Copyright (c) 2017, 2017 Eurotech and/or its affiliates
```
ctron: That would just be "2017"

assembly/README.md
```diff
@@ -28,7 +28,7 @@ Pushing to a specific docker registry under a specific account:
 ### Run
 
     docker run -td --name kapua-sql -p 8181:8181 -p 3306:3306 kapua/kapua-sql
-    docker run -td --name kapua-elasticsearch -p 9200:9200 -p 9300:9300 elasticsearch:2.4 -Dcluster.name=kapua-datastore -Ddiscovery.zen.minimum_master_nodes=1
+    docker run -td --name kapua-elasticsearch -p 9200:9200 -p 9300:9300 elasticsearch:5.3.0 -Ecluster.name=kapua-datastore -Ediscovery.zen.minimum_master_nodes=1 -Enetwork.host=0.0.0.0
```
lorthirk: I tried to launch the `kapua-elasticsearch` with this new command, but it seems the command is wrong since I get a 
```
docker: Error response from daemon: oci runtime error: container_linux.go:247: starting container process caused "exec: \"-Ecluster.name=kapua-datastore\": executable file not found in $PATH".
```

What worked for me was
```
docker run -td --name kapua-elasticsearch -p 9200:9200 -p 9300:9300  -e "cluster.name=kapua-datastore" -e "transport.tcp.connect_timeout=30s" -e "transport.ping_schedule=-1" -e "xpack.security.enabled=false" docker.elastic.co/elasticsearch/elasticsearch:5.3.0 
```

I can provide a fix if that's ok to you

dejanb: @lorthirk that works for me as well, so go ahead with the fix. Out of curiosity, which version of docker are you using?

lorthirk: Version 17.03.1-ce-mac5 (16048)
Channel: stable
b18e2a50cc

(on MacOS)

assembly/pom.xml
```diff
@@ -302,6 +303,51 @@
...
+			<groupId>org.elasticsearch.plugin</groupId>
+			<artifactId>percolator-client</artifactId>
+			<version>${elasticsearch-client-transport.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.logging.log4j</groupId>
+			<artifactId>log4j-api</artifactId>
+			<version>2.7</version>
+		</dependency>
+		<dependency>
```
ctron: I thought that we agreed on not using log4j?

assembly/pom.xml
```diff
@@ -302,6 +303,51 @@
...
+		    <groupId>io.netty</groupId>
+		    <artifactId>netty-all</artifactId>
+		    <version>4.1.5.Final</version>
+		</dependency>
+		<dependency>
+		    <groupId>io.netty</groupId>
+		    <artifactId>netty</artifactId>
+		    <version>3.10.6.Final</version>
+		</dependency>
```
ctron: Do we really need netty 3 & 4. And do we really the "-all" package?

assembly/pom.xml
riccardomodanese: yes, after a full test I'll remove it (do you agree @dejanb ?)

assembly/pom.xml
riccardomodanese: I'm looking at it with @lorthirk

dejanb: Yeah, I'm not happy with all this either and consider it a work in progress. The thing is that Elasticsearch prebuilt client comes with "prebuilt" with a lot of plugins that introduce all these dependencies. The things were failing to start with "class not found" errors until I added all these deps. I wanted to just make it work as a first step and then see how we can make it better and include only things we want to use.

dejanb: BTW I get the same when I try to access the data from the console
`1) Error injecting constructor, org.eclipse.kapua.service.datastore.client.ClientUnavailableException: Error: Cannot load the provided client class name [org.eclipse.kapua.service.datastore.client.transport.TransportDatastoreClient]. Check the configuration.
  at org.eclipse.kapua.service.datastore.internal.ClientInfoRegistryServiceImpl.<init>(ClientInfoRegistryServiceImpl.java:84)
  while locating org.eclipse.kapua.service.datastore.internal.ClientInfoRegistryServiceImpl
  at org.eclipse.kapua.locator.guice.KapuaModule.configure(KapuaModule.java:99)
  while locating org.eclipse.kapua.service.datastore.ClientInfoRegistryService

1 error
	at com.google.inject.internal.InjectorImpl$2.get(InjectorImpl.java:1028)
	at com.google.inject.internal.InjectorImpl.getInstance(InjectorImpl.java:1054)
	at org.eclipse.kapua.locator.guice.GuiceLocatorImpl.getService(GuiceLocatorImpl.java:56)
	at org.eclipse.kapua.service.datastore.internal.mediator.DatastoreMediator.<clinit>(DatastoreMediator.java:68)
	... 85 common frames omitted
Caused by: org.eclipse.kapua.service.datastore.client.ClientUnavailableException: Error: Cannot load the provided client class name [org.eclipse.kapua.service.datastore.client.transport.TransportDatastoreClient]. Check the configuration.
	at org.eclipse.kapua.service.datastore.internal.client.DatastoreClientFactory.getInstance(DatastoreClientFactory.java:72)
	at org.eclipse.kapua.service.datastore.internal.ClientInfoRegistryFacade.<init>(ClientInfoRegistryFacade.java:71)
	at org.eclipse.kapua.service.datastore.internal.ClientInfoRegistryServiceImpl.<init>(ClientInfoRegistryServiceImpl.java:95)
	at org.eclipse.kapua.service.datastore.internal.ClientInfoRegistryServiceImpl$$FastClassByGuice$$e97ca0d7.newInstance(<generated>)
	at com.google.inject.internal.DefaultConstructionProxyFactory$FastClassProxy.newInstance(DefaultConstructionProxyFactory.java:89)
	at com.google.inject.internal.ConstructorInjector.provision(ConstructorInjector.java:111)
	at com.google.inject.internal.ConstructorInjector.construct(ConstructorInjector.java:90)
	at com.google.inject.internal.ConstructorBindingImpl$Factory.get(ConstructorBindingImpl.java:268)
	at com.google.inject.internal.FactoryProxy.get(FactoryProxy.java:56)
	at com.google.inject.internal.ProviderToInternalFactoryAdapter$1.call(ProviderToInternalFactoryAdapter.java:46)
	at com.google.inject.internal.InjectorImpl.callInContext(InjectorImpl.java:1092)
	at com.google.inject.internal.ProviderToInternalFactoryAdapter.get(ProviderToInternalFactoryAdapter.java:40)
	at com.google.inject.internal.SingletonScope$1.get(SingletonScope.java:194)
	at com.google.inject.internal.InternalFactoryToProviderAdapter.get(InternalFactoryToProviderAdapter.java:41)
	at com.google.inject.internal.InjectorImpl$2$1.call(InjectorImpl.java:1019)
	at com.google.inject.internal.InjectorImpl.callInContext(InjectorImpl.java:1092)
	at com.google.inject.internal.InjectorImpl$2.get(InjectorImpl.java:1015)
	... 88 common frames omitted
Caused by: java.lang.reflect.InvocationTargetException: null
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.eclipse.kapua.service.datastore.internal.client.DatastoreClientFactory.getInstance(DatastoreClientFactory.java:68)
	... 104 common frames omitted
Caused by: org.eclipse.kapua.service.datastore.client.ClientUnavailableException: Error: Cannot load the provided client class name [org.eclipse.kapua.service.datastore.client.transport.EsTransportClientProvider]. Check the configuration.
	at org.eclipse.kapua.service.datastore.client.transport.TransportDatastoreClient.<init>(TransportDatastoreClient.java:153)
	at org.eclipse.kapua.service.datastore.client.transport.TransportDatastoreClient.getInstance(TransportDatastoreClient.java:117)
	... 109 common frames omitted
Caused by: java.lang.reflect.InvocationTargetException: null
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.eclipse.kapua.service.datastore.client.transport.TransportDatastoreClient.<init>(TransportDatastoreClient.java:148)
	... 110 common frames omitted
Caused by: org.eclipse.kapua.service.datastore.client.ClientUnavailableException: Error: Failed to configure ElasticSearch transport
	at org.eclipse.kapua.service.datastore.client.transport.EsTransportClientProvider.<init>(EsTransportClientProvider.java:198)
	at org.eclipse.kapua.service.datastore.client.transport.EsTransportClientProvider.<init>(EsTransportClientProvider.java:174)
	at org.eclipse.kapua.service.datastore.client.transport.EsTransportClientProvider.<init>(EsTransportClientProvider.java:164)
	at org.eclipse.kapua.service.datastore.client.transport.EsTransportClientProvider.init(EsTransportClientProvider.java:91)
	... 115 common frames omitted
Caused by: java.lang.NoClassDefFoundError: org/apache/logging/log4j/Logger
	at org.elasticsearch.common.logging.Loggers.getLogger(Loggers.java:101)
	at org.elasticsearch.common.xcontent.support.AbstractXContentParser.<init>(AbstractXContentParser.java:57)
	at org.elasticsearch.common.xcontent.json.JsonXContentParser.<init>(JsonXContentParser.java:44)
	at org.elasticsearch.common.xcontent.json.JsonXContent.createParser(JsonXContent.java:103)
	at org.elasticsearch.common.settings.Setting.parseableStringToList(Setting.java:832)
	at org.elasticsearch.common.settings.Setting.lambda$listSetting$27(Setting.java:786)
	at org.elasticsearch.common.settings.Setting.listSetting(Setting.java:791)
	at org.elasticsearch.common.settings.Setting.listSetting(Setting.java:786)
	at org.elasticsearch.common.network.NetworkService.<clinit>(NetworkService.java:50)
	at org.elasticsearch.client.transport.TransportClient.newPluginService(TransportClient.java:98)
	at org.elasticsearch.client.transport.TransportClient.buildTemplate(TransportClient.java:126)
	at org.elasticsearch.client.transport.TransportClient.<init>(TransportClient.java:268)
	at org.elasticsearch.transport.client.PreBuiltTransportClient.<init>(PreBuiltTransportClient.java:125)
	at org.elasticsearch.transport.client.PreBuiltTransportClient.<init>(PreBuiltTransportClient.java:111)
	at org.elasticsearch.transport.client.PreBuiltTransportClient.<init>(PreBuiltTransportClient.java:101)
	at org.eclipse.kapua.service.datastore.client.transport.EsTransportClientProvider.<init>(EsTransportClientProvider.java:191)
	... 118 common frames omitted
Caused by: java.lang.ClassNotFoundException: org.apache.logging.log4j.Logger
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at org.eclipse.jetty.webapp.WebAppClassLoader.loadClass(WebAppClassLoader.java:560)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 134 common frames omitted
`

riccardomodanese: yes, I had the same problem but the last dependencies I pushed should solve the problem. So using the slf4j-api, log4j-to-slf4j, log4j-over-slf4j and the logback should solve the problem

riccardomodanese: [diff.txt](https://github.com/eclipse/kapua/files/1007871/diff.txt)
With these changes it work for me... could you confirm me?

riccardomodanese: just to recap,
having logback we shouldn't need the log4j anymore, but since the ES embedded node uses log4j objects not present in the logback @ctron solved it writing the log4j2-mock.
So the theoretical scenario is seeing these dependencies:
slf4j-api
log4j-to-slf4j (log4j 2)
log4j-over-slf4j
logback-classic
logback-core

and (only for the test scenario may be) the
log4j2-mock
Am I right?

ctron: +1

service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/converter/ModelContextImpl.java
```diff
@@ -127,7 +127,11 @@ private DatastoreMessage unmarshalDatastoreMessage(Map<String, Object> messageMa
         
         KapuaId scopeId = new KapuaEid(new BigInteger((String) messageMap.get(MessageSchema.MESSAGE_SCOPE_ID)));
         message.setScopeId(scopeId);
-        KapuaId deviceId = new KapuaEid(new BigInteger((String) messageMap.get(MessageSchema.MESSAGE_DEVICE_ID)));
+        KapuaId deviceId = null;
+        String deviceIdStr = (String) messageMap.get(MessageSchema.MESSAGE_DEVICE_ID);
+        if (deviceIdStr != null) {
+            deviceId = new KapuaEid(new BigInteger(deviceIdStr));
```
ctron: I think it would be better to use BigInterger#valueOf as small numbers will be cached and this the GC would need to do less work. Improving overall performance.

riccardomodanese: Unfortunately the deviceId is a String. The valueOf allows only long to be used (at least for java 8)

ctron: Sorry, I overlooked that it was a string. Too bad. because in most cases this will be in the cache I guess. 

We should consider making KapuaEid#valueOf(String) method. But this is outside of the scope of this PR.

ctron:
I think the copyright headers still need a cleanup.

riccardomodanese:
For sure, I didn't fix yet the headers. I leave this task as last.

service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/converter/ModelContextImpl.java
riccardomodanese: I agree, making a static method to build a KapuaId from the string will be a good idea, since we are doing this "conversion" in few parts of the code.
But, as you said, since it's out of the PR scope we can consider to implement it outside this PR.

ctron:
The currently documented ES container starts up with:

```
[2017-05-19T09:32:05,309][INFO ][o.e.l.LicenseService     ] [O_XEhW4] license [be2f3ff4-f973-4962-9cce-5df54f42fe87] mode [trial] - valid
```

I guess we need something different than a trail version, right?

ctron:
The broker fires up with the following error:

```
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/maven/lib/optional/logback-classic-1.1.8.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/maven/lib/optional/slf4j-log4j12-1.7.13.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [ch.qos.logback.classic.util.ContextSelectorStaticBinder]
```

ctron:
log4j2 is referenced with two different versions. I guess those should be aligned:

```
        <log4j-api.version>2.7</log4j-api.version>
        <log4j-to-slf4j.version>2.8.2</log4j-to-slf4j.version>
```

riccardomodanese:
you are right, btw any improvement about remove log4j-api needs?

ctron:
I tried to re-cap that in my last e-mail. log4j-api is indeed required, log4j-over-slf4j only replaces log4j-core.

ctron:
For the console WAR file we will need apply the following patch:

```patch
diff --git a/console/pom.xml b/console/pom.xml
index a5989bd..02e1f60 100644
--- a/console/pom.xml
+++ b/console/pom.xml
@@ -270,6 +270,16 @@
 		    <artifactId>netty</artifactId>
 		    <version>${elasticsearch-netty-3.version}</version>
 		</dependency>
+		
+		<!-- log4j2 for ES -->
+		<dependency>
+			<groupId>org.apache.logging.log4j</groupId>
+			<artifactId>log4j-api</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.logging.log4j</groupId>
+			<artifactId>log4j-to-slf4j</artifactId>
+		</dependency>
     </dependencies>
 
     <build>

```

ctron:
I guess the same is true for the REST API war file.

lorthirk:
about https://github.com/eclipse/kapua/pull/552#issuecomment-302656183 and the License: this is actually coming from the x-pack plugin, which is included in the official elasticsearch docker image. since we don't need that plugin (and its features are indeed disabled by default in the `docker run` command) we just have to remove such plugin, either by using a different base image or removing the x-pack folder inside it.

riccardomodanese:
I already did for the rest-api (https://github.com/eclipse/kapua/pull/552#issuecomment-302662398). I removed these dependencies from the console since were already added by the assembly through the console descriptor. I'm not happy about this asymmetrical behaviour between the 2 descriptors (rest-api and console) but, regardless that, I agree to add these dependencies in the console pom.

ctron:
Do we get xpack through a CQ? Which license is it?

dejanb:
Do we need these

```[WARNING] The following patterns were never triggered in this artifact inclusion filter:
o  'com.twitter:jsr166e'
o  'org.joda:joda-convert'
o  'com.spatial4j:spatial4j'
o  'com.ning:compress-lzf'
```

as they currently are not included in the broker image. I can fix it, just want to confirm are they needed.

riccardomodanese:
I don't think we need them

dejanb:
ok, removed

riccardomodanese:
Hi all, since I rebased this branch, please update your local branch before work on it

dev-tools/src/main/openshift/kapua-template.yml
```diff
@@ -21,7 +21,7 @@ parameters:
   value:            jolokia
 - name:           ELASTIC_SEARCH_MEMORY
   descrption:       The minimum and maximun Java heap space for Elasticsearch
-  value:            512m
+  value:            "-Xms512m -Xmx512m"
```
ctron: This parameter really should be the memory, not the Xms/Xmx settings, please transform this to the proper JVM settings below: `-Xms${…} -Xmx${…}`

dev-tools/src/main/openshift/kapua-template.yml
```diff
@@ -379,7 +379,7 @@ objects:
       spec:
         containers:
         - env:
-          - name: ES_HEAP_SIZE
+          - name: ES_JAVA_OPTS
```
ctron: This generates a duplicate entry

Issue #507, opened 2017-04-24
Prevent NPE with SortFieldImpl


ctron:
As most users of the `SortFieldImpl` probably can't handle `null` field names, this should be prevented in the first place.

The should be no default constructor allowing to set the field name to `null`.

ctron:
Ideally it could be prevented to use an `Impl` class at all by:

```java
interface SortField {
   public static SortField ascending( Field field ) {
     …
   }
   public static SortField ascending( String  field ) {
     …
   }
}
```

riccardomodanese:
May I don't understand your snippet, are you suggesting to add a constructor in an interface? From my knowledge java 1.8 (or 1.7?) added static methods and default methods to the interfaces but not the constructors. Am I wrong?
If I'm understanding your concern, we can handle the use case through replacing the SortField interface with an abstract class SortField with a default constructor that validates the field.
So every possible future implementation must call the SortField(String field) constructor.

public abstract class SortField {

     private SortField() {
     }

     protected SortField(String field) throws SomeException {
           if (org.apache.commons.lang3.StringUtils.isEmpty(field)) {
                  throw new SomeException();
           }
           this.field = field;
     }
}

ctron:
No, I am not talking about a constructor, but a static method, much like the comparator methods in Java 8 (which should also be possible before Java 7 IIRC): https://docs.oracle.com/javase/8/docs/api/java/util/Comparator.html#comparing-java.util.function.Function-

The way you describe it now make a class instead of a field out of this, but from what I've understood it should be an interface.

Not handling this in a constructor has the advantage that you can actually have different paths of handling different situations **before** the constructor is called. e.g. like returning `null` from that call.

So you can also leave this with your custom implementation (if required) or use the static methods on the interface.

ctron:
I made an example in PR #554

riccardomodanese:
Sorry, may I explained myself badly, in my opinion there is no reason to have an interface for that Object.
Moreover, talking also with @Coduz, there is no need to have an abstract class for that. We can just provide a default concrete SortableField implementation.
In any case I'll take a look to your pr

ctron:
I amended the PR to go with a simple class instead of an interface then.

riccardomodanese:
Hi @ctron the fix is already included in the codebase so I close the issue.

Regards

PR #554, opened 2017-05-11
Clean up SortField/SortFieldImpl API

ctron:
Also see #507


service/datastore/api/src/main/java/org/eclipse/kapua/service/datastore/model/query/SortField.java
```diff
@@ -8,43 +8,49 @@
...
+        return this.field;
+    }
+
+    public SortDirection getSortDirection() {
+        return this.sortDirection;
+    }
+
+    public static SortField of(SortDirection direction, final String fieldName) {
+        if (fieldName == null || fieldName.isEmpty()) {
+            return null;
```
riccardomodanese: Instead of returning a null object, I prefer to raise an exception if the fieldName is null, so the caller or get a valid and usable object or should handle the "malformed" object in the catch block.
What do you think?

ctron: Well this method could work with null values, by the way provided. So there is no real need to return a NPE. However the method receiving the "SortField" instance, may or may not be capable of handling null, so I would say let this method then handle the null situation.

Assume you have some UI or API which allows you to omit the sort field, then you would have a null check in this method and the method receiving the SortField instance. Letting the receiver handle the null situation alone would limit the null-check to the methods which really have a problem receiving null.

riccardomodanese: I agree, the UI can omit the sort field. Sorry I missed that this object is exposed outside from the datastore.
I'll check the datastore code to be sure that, if the sort field is null, no sort condition is mapped to ES client layer. If the check is not implemented, I'll open an issue.

PR #559, opened 2017-05-11
Refactor current SSO implementation for re-usability and extensibility

ctron:
As already discussed on the mailing list, I would like to make the current SSO implementation capable of being re-use and extended by different SSO providers, without changing any logic at the moment.

This PR is a base of discussion. Although most changes are done it is not 100% complete and not tested.


lorthirk:
Hi @ctron,

Me and @Coduz reviewed your PR and have a couple of questions about it:

- Why did you create a separate Locator hierarchy, and not leveraging the `KapuaLocator` we already use? That could be done by just marking services as `KapuaService` (that is just a marker interface to let the locator find them). This way we would still use the same locator all across the project
- Why `GenericSingleSignOnService and `KeycloakSingleSignOnService` are separate services? We're currently using Keycloak's OpenID Connect interface, so you could seamlessly integrate with Keycloak using the `GenericSingleSignOnService`. Or do you have additional features for `KeycloakSingleSignOnService` in the future?

Great job anyway!

ctron:
So the question about Generic and Keycloak is simple. The generic implementation should have a more basic approach (not requiring keycloak) where the keycloak variant makes some assumptions which only hold true when using keycloak. Ideally for keycloak you only need to configure the base URL, client id and realm and that is it. For the generic version you will need to add a few things more as there is no knowledge how the URL have to be built from the base URL, realm and client id.

Regarding the locator stuff, this is the biggest point of discussion, also for me. The current problem is that the locator stuff has no lifecycle, which we do require for this one. Http connections should get pooled to increase performance, which may result in resource leaks when such things are not properly closed. The second motivation behind that is that there should be some sort of "drop in" functionality. Dropping in another JAR with an additional SSO provider should make this available out of the box, which is possible in the current way, using the Java service locator framework. Still you have the possibility to override the service and the locator. And we could build one SingleSignOnLocator implementation based on the KapuaLocator as soon as we get the lifecycle functionality.

lorthirk:
Generic vs Keycloak provider: I see your point, but as far as the OpenID Connect support is concerned, setting realm, clientId and base URL only means to have a dynamically composed URL instead of writing it in full; with your PR we duplicate code just to achieve the dynamic composition. Maybe we can just have Keycloak inherit Generic only overriding the URL generation?

Locator: @stefanomorson is currently working on lifecycle management on KapuaLocator. Perhaps we could wait for this submission to rely on it from the beginning?

ctron:
True: At the moment we only have this for Keycloak vs Generic. And, as I tried to explain, this refactoring is not complete, but a first step. If Keycloak inherits from Generic or are two separate classes, or inheriting from the same base class, we will see.

We can always refactor the locator. But I would like to start with what we currently need. If the other locator is ready, then we can migrate to that. But let's not keep waiting.

ctron:
Do we have any update on the new lifecycle features? Or can we move on here?

stefanomorson:
Generic/Keycloak provider. I took some time to understand the pr in depth. I don't see the need to have a dedicated Keycloak provider since it looks very similar to the generic one, moreover generality and interoperability give a lot of value here. 
However I see the point, maybe we have touched here the need to provide support for pluggability for 
this kind of use cases ? In the new lifecycle features under development I tried to introduce the concept (see branch fix-disposeResources).

Lifecycle.  The new lifecycle features are not yet ready to be merged in develop although in a good
progress. I would say that the we can use what you have done right now but then we'll need to refactor 
a bit in the perspective of the new lifecycle management features.

ctron:
Agreed, so I will continue to work on this PR to merge this later on.

stefanomorson:
@ctron : in my opinion it is more effective for the project if we keep it neutral in respect of external identity providers and rely on standards. Can you simply add the api, provider-generic and provider stuff ?

ctron:
Well that is why there is a service locator, which allows for different backing implementations. In order to develop this properly we need at least two, to see if this concept works.

As you might have seen, after continuing work, the generic and keycloak share more code now. Each one specializing on what is required. So you can use any implementation based on the generic and a special one using keycloak. Adding a second provider doesn't hurt?! But adds value to the end user.

ctron:
And we also do have a works-with CQ for keycloak. So no trouble there.

stefanomorson:
Yes I saw the changes in terms of code reuse also got the locator trick.
But what will be the project layout when we'll need a class for ActiveDirectory, one per LDAP, one for OID and so on ? Should we add all of these as small projects in the Kapua code base ?
Ideally we don't want to add product specific stuff in the Kapua code base.

ctron:
I don't understand the reason not to add it. Adding it doesn't cause any downside to the rest of the project. If we need LDAP or ActiveDirectory or whatever, then we add it, yes.

So Keycloak is not a product, but a project. And we already did add project specific stuff to Kapua: ActiveMQ, Jetty, Docker, Vagrant, OpenShift, … But on this case we actually offer an alternative, the generic module.

So what would be good reasons not to merge this in?

ctron:
The patch is ready to be merged from my side. As we agreed yesterday I would like to merge this now. So that we can continue with the 0.2.0 release.

dejanb:
I also don't see any problems with merging this for the 0.2.0 release. If needed we can make more general interfaces later on when the need is there

hekonsek:
Agreed with Jens and Dejan. This is normal for open source projects to have dependencies to other open source projects. Following this logic we should remove ElasticSearch support from Kapua as well :) . You still can use Kapua without KeyCloak, so I don't see an reason not to merge this.

ctron:
So just to be sure we all understand the dependency on Keycloak here.

The "generic" and the "keycloak" provider both use the same Open ID connect implementation in the base class. The keycloak provider just is easier to configure because it knows how keycloak has to be configured and may, in the future, take more auto-configuration from Keycloak.

stefanomorson:
@ctron, @hekonsek, @dejanb 

I think that the idea of dependency from other projects is not a point of discussion here since we're 
implementing a whole platform. That's pretty consolidated concept.
Dependencies however can be of different types. For example Vagrant, Docker or OpenShift use Kapua, it is not Kapua using them. 
Then there are cases when there's no established standards and standard libraries that help you decouple the code from the specific project, that's for example the case with NoSQLs, one of which is Elasticsearch. 
For the messaging, currently implemented through ActiveMQ, we started a discussion in the last weeks to see what the future evolutions will be. 
I thought that since OpenID is a standard natively supported by Kapua, Kapua could use any external authentication service, included Keycloak, that conforms to the standard. And actually Keycloak was used for tests, and it worked. And that's great.
So I'm a bit surprised now to know that we need Keycloak specific code. That's it.

ctron:
Thanks @stefanomorson for merging this.

I will try to briefly explain that a little bit and, if you like, I can explain this in a little bit more detail on the call next Thursday.

Kapua still can work with any OpenID connect solution after this merge. However OpenID connect leaves a few areas undefined, like for example how the URLs are defined. Keycloak on the other side makes a definition here and fills in those gaps. Now the "keycloak" specific provider makes use of this, and makes it simpler (more user friendly) to configure Kapua, when using keycloak.

It is still possible to use the "generic" provider with Keycloak as well. You just need to do more configuration work and figure out the URIs and other stuff yourself. Now that we have done this first little step, we can advance and simplify the use of Keycloak even more, if possible. And when we come to some other solution which needs a bit of a tweak, we can add that as well. But we won't need any "if" branches because we have different providers now.

I hope this explains it a bit better.

stefanomorson:
@ctron 

Yes it does actually, thanks. 
Thank you also for the open discussion, it's good for the project and it means it is lively.
I'm looking forward for the release to come, a new (big) step ahead for the project and the team.

PR #561, opened 2017-05-11
Source external configurations files

### LINKED ISSUE #556: Let service configurations be uploaded from an external resource file
Each Kapua project has its own configuration file. Currently configurations are loaded following this order:

* System properties (if any)
* Environment properties (if any)
* (default) property file in the project classpath

In real world scenarios, since there are many properties that may need to be customized, we may want to specify an external custom property file (it may be located in a specific folder) to be uploaded. The new configuration load order should be:

* System properties (if any)
* Environment properties (if any)
* External file (if any)
* (default) property file in the project classpath

Since in a Kapua instance there are many of those files, one for each project, it would be easier to drop them into a single dir and specify the dir to be searched as a system property.

Coduz:
Hi all ,

this PR closes #556 

You can now specify system properties called

- `kapua.config.file` : The file path specified will be loaded as a config file. Path MUST have schema (so `file://` or `https://`)
- `kapua.config.dir` : The dir name specified will be scanned for all files that not start with `.` (hidden files) and ends with `.properties` (regular Java properties files).

Both options can be specified. 

If `kapua.config.file` is not found an error will be thrown. If it is empty no property override will be loaded

If `kapua.config.dir` is not found or is not a directory an error wil lbe thrown. If empty dir no error will occur.

Regards, 

_Alberto_

### DIFF
--- a/commons/src/main/java/org/eclipse/kapua/commons/setting/AbstractKapuaSetting.java
+++ b/commons/src/main/java/org/eclipse/kapua/commons/setting/AbstractKapuaSetting.java
@@ -38,7 +38,7 @@ public abstract class AbstractKapuaSetting<K extends SettingKey> extends Abstrac
 
     private static final Logger LOG = LoggerFactory.getLogger(AbstractKapuaSetting.class);
 
-    private static final String EXTERNAL_CONFIG_FILE_PARAM = "kapua.config.file";
+    private static final String EXTERNAL_CONFIG_URL_PARAM = "kapua.config.url";
     private static final String EXTERNAL_CONFIG_DIR_PARAM = "kapua.config.dir";
 
     /**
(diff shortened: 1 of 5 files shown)


Coduz:
As suggested by @stzilli I will improve the http/https file resource to handle self signed certificates that might not be accepted over Https

ctron:
Just a suggestion if `kapua.config.file` is actually enforcing the use of a URL, would we name it `kapua.config.url` instead. Otherwise I would assume many people run into a problem around that.

Coduz:
About the failure on QA project, on my local is working and I don't thing that my changes has caused problems on QA. Any ideas?

Coduz:
Rebased onto develop

Coduz:
Rebased onto latest develop. 
Failures are disappeared... Is it safe to merge, then?

PR #572, opened 2017-05-16
Fix/check permission on creation

### LINKED ISSUE #569: Add check to the targetScopeId when adding a permission
Hi all,

We have a bug that is also a security problem. 
When we add a new permission we do not check the `targetScopeId` set. 
Is fine when is the same of the `scopeId`  of the permission, but when they are different we need to check that the current user has the permission on the `targetScopeId` of the new permission. 
If not, any user with `role:manage` or `access-info:manage` can add permission in other accounts.

Regards, 

_Alberto_

Coduz:
Hi all, 

this PR fixes #569 .

Regards, 

_Alberto_

### DIFF
--- a/service/security/shiro/src/test/resources/features/AccessInfoService.feature
+++ b/service/security/shiro/src/test/resources/features/AccessInfoService.feature
@@ -236,6 +236,10 @@ Scenario: Create regular access permissions
     It must be possible to create sets of regular permissions.
 
     Given A scope with ID 10
+    And A user such as
+        | name     | displayName     | email              | phoneNumber     |
+        | kapua-u1 | Kapua User 1    | kapua_u1@kapua.com | +386 31 323 555 |
+    When I create the access info entity
     And The permissions "read, write, execute"
     And I create the permissions
     Given A scope with ID 20
@@ -250,6 +254,10 @@ Scenario: Find last created permision
     It must be possible to find a specific permission entry based on its ID.
 
     Given A scope with ID 10
+    And A user such as
+        | name     | displayName     | email              | phoneNumber     |
+        | kapua-u1 | Kapua User 1    | kapua_u1@kapua.com | +386 31 323 555 |
+    When I create the access info entity
     And The permissions "read, write, execute"
     And I create the permissions
     When I search for the last created permission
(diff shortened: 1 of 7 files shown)


Coduz:
Hi @muros-ct 

I've changed the authorization service feature to accomplish new checks.
Can you please have a look at them? I don't know if they are ok :) 

https://github.com/eclipse/kapua/pull/572/files#diff-24016ea42f423b1ef6e731c969af83a2

Thanks, 

_Alberto_

service/security/shiro/src/test/resources/features/AccessInfoService.feature
```diff
@@ -328,14 +359,33 @@ Scenario: Count access role entities by scope
...
+    And The role "test_role_2"
     And I create the access role
+    And The role "test_role_3"
     And I create the access role
+    
     Given A scope with ID 20
+    And A user such as
+        | name     | displayName     | email              | phoneNumber     |
+        | kapua-u2 | Kapua User 2    | kapua_u2@kapua.com | +386 31 323 555 |
+    When I create the access info entity
```
Coduz: In here I explicitly create only one access role because `When I create the access info entity` creates already an access role.

muros-ct: It seams ok to me, from standpoint of existing tests.
But all scenarios were changed so that they pass implementation change. Am I right?
Which scenario checks for failure.

Coduz: I've just changed scenarios that was failing because of new integrity checks (i.e. check that the `roleId` and `accessInfoId` exists when creating an `accessRole`. 
So the scenarios are the same as before as (I think) I changed only preconditions of scenarios, not the final checks. 
Are you suggesting to add also negative test scenarios?

muros-ct: Me or Andrej can add it. It will require some changes on tests side, but I think that if that was a bug it should be covered for regression reasons.
It could be single test that checks that your change is working.

But this should not stop your merge!

Coduz: Great! Do you prefer to merge this and add further tests after or make a pull request on this branch?

muros-ct: Yes certainly, merge your changes. Tests will be part of new PR.

PR #573, opened 2017-05-16
Modifed tests for KapuaContainer functionality

muros-ct:
While retrieving list of services duplicates are not added to list.
Changes were made in KapuaLocatorImpl.

KapuaContainer and its startup and shutdown methods added to unit tests.
For same purpose KapuaTest super class was modified.

KapuaContainer and its functionality added to QA module integration tests.
ActiveMQ broker configuration for integration tests was also modified to be
in sync witn broker implementation with KapuaContainer.


locator/guice/src/main/java/org/eclipse/kapua/locator/guice/KapuaLocatorImpl.java
```diff
@@ -117,7 +117,10 @@
         for (Binding<?> binding : bindings.values()) {
             final Class<?> clazz = binding.getKey().getTypeLiteral().getRawType();
             if (KapuaService.class.isAssignableFrom(clazz)) {
-                servicesList.add(injector.getInstance(clazz.asSubclass(KapuaService.class)));
+                KapuaService serviceInstance = injector.getInstance(clazz.asSubclass(KapuaService.class));
```
ctron: Wouldn't as `Set` be the right data type here?

muros-ct: I recommended Set as an option, but it would change method signature.
Also I am not sure if order of items is important and how are lists iterated over.

muros-ct:
I am experiencing strange test execution behaviour.
Tests run successfully on local build with maven.
Tests are successful on my fork travis build.
Tests fail on PR.

Test failures deal with functionality of locator that was changed in fix-disposeResources branch.

locator/guice/src/main/java/org/eclipse/kapua/locator/guice/KapuaLocatorImpl.java
stefanomorson: Set looks good but yeah I also prefer not to change for the same reasons.
Will check with @Coduz and @lorthirk a do it later.

PR #576, opened 2017-05-17
Change phases for docker container build/push

stzilli:
When running "mvn clean install -Pdocker" fabric8 tries to push the docker containers by default to DockerHub. I would expect this step to be executed only during the deploy phase. Also I think the creation of the container itself should be done during the install phase.


ctron:
@stzilli Please add yourself to the copyright header

Issue #531, opened 2017-05-02
Unit tests messed up?


ctron:
To me it looks a little bit as if the unit tests are messed up, or I simply don't understand what is going on.

For example in the project `kapua-account-internal` the `locator.xml` specifies an API `<api>org.eclipse.kapua.service.authentication.UsernamePasswordTokenFactory</api>` which does not exist. If the unit tests are run locally with Eclipse, the unit test runner fails with a CNFE, looking up this class.

However if run with Maven those tests succeed. So it looks like as if the classpath is messed up and the unit tests of `kapua-account-internal` actually use some different `locator.xml`.

Removing this entry for the `locator.xml` then fails the unit tests with the next exception:

```
java.lang.ClassCastException: org.eclipse.kapua.locator.guice.GuiceLocatorImpl cannot be cast to org.eclipse.kapua.test.MockedLocator
	at org.eclipse.kapua.service.account.internal.AccountServiceTestSteps.beforeScenario(AccountServiceTestSteps.java:128)
```

So maybe there is a simple explanation or a fix needed for this.

muros-ct:
Yes it is true, it is not obvious.
Tests were first made in IDE and then made to work with maven. But in between we forgot to "mention" this small configuration issue.
As this tests are testing single service in isolation, other services are mocked. But for mocks to work we need custom Locator. This locator is configured from command line with

`-Dlocator.class.impl=org.eclipse.kapua.test.MockedLocator`

This is specified in pom.xml file for surefire tests.

What would be the best option to document this?  
Could we automate this somehow? How to create run / debug configuration in eclipse?  
Suggestions?

```xml
<build>
  <plugins>
    <plugin>
      <groupId>org.apache.maven.plugins</groupId>
      <artifactId>maven-surefire-plugin</artifactId>
      <configuration>
        <systemPropertyVariables>
          <locator.class.impl>org.eclipse.kapua.test.MockedLocator</locator.class.impl>
        </systemPropertyVariables>
      </configuration>
    </plugin>
  </plugins>
</build>
```

ctron:
I think this should be somehow coded into the unit test instead of relying on the existence of a system property. As we just saw, this can cause problems.

Maybe `@BeforeClass`, some custom unit runner, explicitly using this locator, …

Coduz:
Hi @muros-ct,

#557 has fixed this. 
Is it ok to close this issue? 

Regards 

_Alberto_

muros-ct:
[...]

PR #577, opened 2017-05-17
Fixing tests that are dependent on MockedLocator

muros-ct:
Custom cucumber runner for junit was created. This runner
accepts system properties that provided by annotating junit run class.

Pom files ware cleaned of this dependency to system property and all
the tests that were using MockedLocator ware corrected.

This fixes issue #531

### DIFF
--- a/service/account/internal/pom.xml
+++ b/service/account/internal/pom.xml
@@ -15,20 +15,6 @@
 
     <artifactId>kapua-account-internal</artifactId>
 
-    <build>
-        <plugins>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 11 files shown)


ctron:
I really like the approach! You could use [repeating annotations](https://docs.oracle.com/javase/tutorial/java/annotations/repeating.html) in order to get rid of the array. That way you could simply add additional annotations as needed.

muros-ct:
Sure, I will do that. Arrays can get cluttered and hard to read.

muros-ct:
Annotation was replaced with repeatable annotation.
@ctron please take a look

muros-ct:
Hi @ctron, I discussed with Stefano. He said that is ok if you merge.

PR #583, opened 2017-05-23
Data deletion integrity at service level

Coduz:
Hi all, 

this PR introduces improvements on the data integrity. 
Currently if you delete an entity on a service, entities related to the deleted entity does not get deleted (ie: delete a device but related device events still there). This lead to inconsistent data into the database that will remain there unless a user action. 

Since some relationship between entities are very strict we can enforce them. 
Some examples are:

- Device Events -> Device
- RolePermission -> Role
- AccessPermission -> AccessInfo
- AccessRole -> AccessInfo

There are also relationship between entities that are cross-service (ie: Device -> Account) but they will be covered when we will have the support of Business Events.

This PR introduces data integrity within services. 
This is done by creating foreign keys on tables. 
This has required a bit of refactoring of the Liquibase scripts because the order of execution matters when using foreign keys. 

While working on this item I discovered also that there is another changes that Liquibase scripts need. I'll open another issue on that.

Regards, 

_Alberto_


Coduz:
This PR probably needs some adjustments but the overall structure is there :)

PR #589, opened 2017-05-24
Fix/configurable service parameters

### LINKED ISSUE #570: Bug on account configurable service permission check
Hi all, 

there is a bug on the `AbstractKapuaConfigurableResourceLimitedService` which is using `Map<String, Object> childConfigValues = getConfigValues(childEntity.getId());` on line 92. 
This works with AccountService but with no with the other services with needs to use `Map<String, Object> childConfigValues = getConfigValues(childEntity.getScopeId());`

_Alberto_

Coduz:
Hi all, 

this PR fixes #570  with some other small fixes and improvements.

Regards,

_Alberto_

### DIFF
--- a/commons/src/main/java/org/eclipse/kapua/commons/configuration/AbstractKapuaConfigurableResourceLimitedService.java
+++ b/commons/src/main/java/org/eclipse/kapua/commons/configuration/AbstractKapuaConfigurableResourceLimitedService.java
@@ -15,14 +15,17 @@ import java.util.Map;
 
 import org.eclipse.kapua.KapuaException;
 import org.eclipse.kapua.commons.jpa.EntityManagerFactory;
+import org.eclipse.kapua.commons.model.query.predicate.AttributePredicate;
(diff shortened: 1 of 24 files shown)


commons/src/main/java/org/eclipse/kapua/commons/configuration/AbstractKapuaConfigurableResourceLimitedService.java
```diff
@@ -75,7 +82,7 @@ protected int allowedChildEntities(KapuaId scopeId) throws KapuaException {
      * @return the number of child accounts spots still available
      * @throws KapuaException
      */
-    protected int allowedChildEntities(KapuaId scopeId, Map<String, Object> configuration) throws KapuaException {
+    protected int allowedChildEntities(KapuaId scopeId, KapuaId targetScopeId, Map<String, Object> configuration) throws KapuaException {
```
lorthirk: Can you add javadoc for `KapuaId targetScopeId`?

PR #593, opened 2017-05-25
Allow both strict and non-strict channel filtering 

lorthirk:
Hi all,

With this PR both strict and non-strict channel filtering will be allowed. Console will always use non-strict channel filtering (i.e. will show metrics of child channels when selecting a channels), while REST APIs will support both modes via a query parameter.

Cheers!


rest-api/src/main/java/org/eclipse/kapua/app/api/v1/resources/DataMessages.java
```diff
@@ -109,7 +113,12 @@
             }
 
             if (!Strings.isNullOrEmpty(channel)) {
-                ChannelMatchPredicate channelPredicate = storablePredicateFactory.newChannelMatchPredicate(channel);
+                StorablePredicate channelPredicate = null;
+                if (strictChannel) {
+                    channelPredicate = storablePredicateFactory.newTermPredicate(new StorableFieldImpl("channel"), channel);
```
Coduz: Can you please use `ChannelInfoField.CHANNEL.field()`?

PR #594, opened 2017-05-26
fix negative number id generator (only positive number will be genera…

riccardomodanese:
The previous version of the id generator generated also negative ids. This may be an issue for the incoming feature and also for the datastore operations.
So it is preferable to modify the implementation to generate only positive numbers.


Coduz:
Hi all ,

 this failure is related to #600 

If tests in QA occurs with some Liquibase exception, it is likely that they are not related to this PR

Regards, 

_Alberto_

PR #597, opened 2017-05-26
Console POM cleanup

lorthirk:
Hi all,

Using IntelliJ, that automatically configures the projects according to the POM file, I noticed that the console one needed to be cleaned up a bit.

Please review these changes and test them with any running configuration you need!

Cheers

### DIFF
--- a/console/pom.xml
+++ b/console/pom.xml
@@ -26,8 +26,8 @@
     <packaging>war</packaging>
 
     <properties>
-        <webappDirectory>${project.build.directory}/Kapua</webappDirectory>
-        <site.basedir>${project.parent.basedir}</site.basedir>
+        <webappDirectory>${project.build.directory}</webappDirectory>
+        <site.basedir>${project.build.directory}/${project.build.finalName}</site.basedir>
 
         <!-- Convenience property to set the GWT version -->
         <gwt.version>2.4.0</gwt.version>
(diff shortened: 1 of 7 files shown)


console/pom.xml
```diff
@@ -271,7 +271,7 @@
     </dependencies>
 
     <build>
-        <finalName>console</finalName>
+        <finalName>Kapua</finalName>
```
ctron: I think this will create a war file named `Kapua.war`, right?

lorthirk: Yes, it will. Do you think it could be a problem?

lorthirk:
As discussed on our last community meeting on June 1st, probably the best way is to rename the GWT module from "Kapua" to something more meaningful. "console" wouldn't be a good fit since GWT would then overwrite the native JavaScript "console" object. "Console" (with capital C) would probably work as a workaround, but I don't like it since it would be confusing. 

My vote is for "admin", since it is the Kapua Administration Console, and other web applications often use the word for such usages.

ctron:
+1 for admin

lorthirk:
So, I made some tests and both Vagrant and Docker seems to be fine, like the target folder. For me it's ready to be merged.

Coduz:
Demo vagrant machine is broken (broker does not start) due to changes not related to this branch. 
Console works fine
We can merge this 😄

PR #605, opened 2017-05-29
Datastore cucumber tests

muros-ct:
First test of datastore implemented with cucumber.

Runner and basic steps are defined. Based on this new
scenarios will be ported from datastore unit tests.


muros-ct:
@riccardomodanese As you were the author of original junit tests for datastore, can you check this transition to cucumber?
This is just the first tests. I will introduce all the tests that are in junit and then add more, as we discussed.

Issue #586, opened 2017-05-24
Liquibase script refactoring


Coduz:
Hi all, 

while working on #583 I discovered something odd on Liquibase.
By default it uses absolute path name as a key to know if a file has been executed already or not. 
This means that If the same script is executed from a different directory Liquibase wouldn't be able to recognize that is has already been executed and try to execute it again.

There is a property to override this called `logicalFilePath`, which value is used as a key instead of the absolute path. This property must be added to all file with changeset to execute. 
I was not able to add it to the plan SQL file and I suspect that we have to convert them to XML|YAML|JSON format to be able to specify that property. 

BTW the current layout works fine for tests and demo deployment. It could be a problem in case of more complex usage.

Regards,

_Alberto_

EDIT: 
Documentation page of this property: http://www.liquibase.org/documentation/databasechangelog.html

Coduz:
Task list:

- [ ] set default collation to recommended one (uft8_unicode_ci?)
- [ ] check existing indexes

mcarrer:
Closed, now tracked by DB migration enhancement PR #788.

PR #606, opened 2017-05-29
Fixed QA test random failure

### LINKED ISSUE #600: Random QA tests failures
Hi all,

after merging #583 some random test failures started to appear more frequently. We had already some random failures on qa tests but they were occasional.

After digging a bit I discovered that is a problem related to how Liquibase script are loaded and the fact that the Kapua Liquibase Client  is run from many different places concurrently. 

The KapuaLiquibaseClient was implemented to copy resources from the jar to a tmp directory and then execute it immediately. With the improvement contained on #583  all scripts gets copied and then executed. 

For the fact that KapuaLiquibaseClient is run from different places it can happen that the copied resources gets overwritten by another KapuaLiquibaseClient is executing them, with even more chance that now that apply of Liquibase scripts takes a bit longer.

I'll fix this making execution of Liquibase script `synchronized` as a simple patch.

Regards,

_Alberto_

Coduz:
Hi all, 

this PR fixes #600 

Regards, 

_Alberto_

### DIFF
--- a/service/liquibase/src/main/java/org/eclipse/kapua/service/liquibase/KapuaLiquibaseClient.java
+++ b/service/liquibase/src/main/java/org/eclipse/kapua/service/liquibase/KapuaLiquibaseClient.java
@@ -65,55 +65,58 @@ public class KapuaLiquibaseClient {
         try {
             if (parseBoolean(System.getProperty("LIQUIBASE_ENABLED", "true")) || parseBoolean(System.getenv("LIQUIBASE_ENABLED"))) {
                 try (Connection connection = DriverManager.getConnection(jdbcUrl, username, password)) {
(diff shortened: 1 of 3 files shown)


ctron:
What speaks against using Java 8 temp directory creation?

ctron:
Please have a look at PR #610 for a lock free solution to this.

Coduz:
@ctron  please read #586. You cannot have the directory have a variable name. Other test fails otherwise.

ctron:
Ok, I see. That should be put as a comment in the source IMHO.

Coduz:
This is something internal of Liquibase. Is something that we need to change to use the `logicalFilePath` property.

PR #610, opened 2017-05-30
Use Java 8 temp dir creation

ctron:


service/liquibase/src/main/java/org/eclipse/kapua/service/liquibase/KapuaLiquibaseClient.java
```diff
@@ -109,11 +103,15 @@ public void update() {
 
                         LOG.debug("Excuted liquibase script: {}", masterChangelog.getAbsolutePath());
                     }
+
+                } finally {
+                    FileUtils.deleteDirectory(changelogTempDirectory.toFile());
```
Coduz: I think that the delete must be performed before because you don't know if the directory has been deleted by the previous execution (i.e.: the application in was stopped before this finally while debugging). 
Moreover it would be nice to know which scripts have been found and executed after the component start.

Coduz: Nevermind, see my other comment on #606

PR #629, opened 2017-06-01
Fixed multiples headers show with same name

Coduz:
Hi all ,

this PR fixes the Metrics table on the data view  showing multiple times the same metric name.

Regards ,

_Alberto_


Coduz:
screen shot 2017-06-01 at 3 06 13 pm

example

PR #634, opened 2017-06-02
Minor Performance issue

AnaScarlet:
Changed my ECA email address to (hopefully) pass IP check.


ctron:
Thanks for the contribution. However it looks like you used 'AnaScarlet@users.noreply.hithub.com' as your e-mail.

Also the footer `Signed-off-by` seems to be missing. At least the `Signed-off-by` footer requires you to use the same e-mail address as your Eclipse account has, with the signed ECA.

Please also add yourself to the copyright header, like here for example: https://github.com/arthurdeschamps/kapua/blob/97d6ba2698f7b74fe1f99fa746dd22572eea5e34/commons/src/main/java/org/eclipse/kapua/commons/about/AboutScanner.java#L11

PR #637, opened 2017-06-02
ACL broker tests initial commit

muros-ct:
This is first example of simple ACL test for broker.

Cucumber scenario is written that connects client as system user and
publishes message to topic. Assertion is made that it recieves this
message back from broker.

Additional tests will check for ACL that are based on regular user
connection to broker and performing operations that are or not permited
for his profile.


muros-ct:
This is not required test for 0.2.0 release.

PR #640, opened 2017-06-05
Fix for unit test failures in Windows environment

andrejn-ct:
## Fix for unit test failures in Windows
This is a potential fix for the consistent failures of unit tests in several packages.
The issue seems to be the reluctance of Windows to delete files that still have open output streams associated with them.
This fix should guarantee that the output streams are reliably closed before proceeding.

There are no changes in pom files.

### Implementation changes
The _KapuaLiquibaseClient_ method _loadResourcesStatic_ is modified to close the output stream before proceeding.


ctron:
I will cherry-pick this one on the release branch.

Issue #584, opened 2017-05-24
You have an error in your SQL syntax of comment,when use mariadb


miricy:
--liquibase formatted sql

--changeset domain_actions:1

add a spacing char is ok

-- liquibase formatted sql

-- changeset domain_actions:1

 Error: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '--liquibase formatted sql

--changeset domain_actions:1

CREATE TABLE athz_domai' at line 13
Query is : -- *******************************************************************************
-- Copyright (c) 2011, 2016 Eurotech and/or its affiliates and others
--
-- All rights reserved. This program and the accompanying materials
-- are made available under the terms of the Eclipse Public License v1.0
-- which accompanies this distribution, and is available at
-- http://www.eclipse.org/legal/epl-v10.html
--
-- Contributors:
--     Eurotech - initial API and implementation
-- *******************************************************************************

--liquibase formatted sql

--changeset domain_actions:1

CREATE TABLE athz_domain_actions (
  domain_id                 BIGINT(21)    UNSIGNED NOT NULL,
  action                                        VARCHAR(255)  NOT NULL,
  
  PRIMARY KEY (domain_id, action)
 
) DEFAULT CHARSET=utf8;

Coduz:
Hi @miricy ,

Is this something that needs to be changed in all Liquibase formatted SQL files? 

Thanks,

_Alberto_

Coduz:
Hi @miricy,

Any news? 

Regards, 

_Alberto_

miricy:
Hi @Coduz ,
   there is some fix like this:
# 1. The file commons/src/main/resources/kapua-environment-setting.properties

commons.db.name=kapuadb
commons.db.username=kapua
commons.db.password=kapua
commons.db.jdbcConnectionUrlResolver=mysql

commons.db.jdbc.driver=com.mysql.jdbc.Driver
commons.db.connection.scheme=jdbc:mysql

# 2.The file commons/src/main/java/org/eclipse/kapua/commons/jpa/JdbcConnectionUrlResolvers.java
modify code:


package org.eclipse.kapua.commons.jpa;

import static org.slf4j.LoggerFactory.getLogger;

import org.eclipse.kapua.commons.setting.system.SystemSetting;
import org.eclipse.kapua.commons.setting.system.SystemSettingKey;
import org.slf4j.Logger;

public final class JdbcConnectionUrlResolvers {

[...]

PR #642, opened 2017-06-06
Fix/sql running on my sql

Coduz:
Hi all, 

This PR intoduces changes to the SQL scripts to be compatible with MariaDB (and also MySQL) as underling databases.

This should solve some problems from issue #584 

Regards, 

_Alberto_


service/user/internal/src/main/resources/liquibase/0.2.0/user.sql
```diff
@@ -34,7 +34,7 @@ CREATE TABLE usr_user (
 
   PRIMARY KEY (id),
   CONSTRAINT usr_uc_name UNIQUE (name)
-) DEFAULT CHARSET=utf8;
+) ENGINE = InnoDB DEFAULT CHARSET = utf8;
```
ctron: Will this work on H2?

Coduz: Yes this works :) Tested on demo machine and build on Travis. 
H2 and MySQL have a different way/keywords to define `COLLATION` so I was not able to specify it. I would like to use `COLLATION = utf8_unicode_ci`.

Did you have any problem with this?

Issue #659, opened 2017-06-13
Unable to create account with AccountCreator/AccountService


ctron:
Creating a new account with the `AccountCreator` / `AccountService` combination seems to fail due to a missing, but unsettable field in the case the code is running with `doPrivileged`, making it impossible to let the system create entities:

```
NULL not allowed for column "CREATED_BY"
```

<details>
Error Code: 23502
Call: INSERT INTO act_account (attributes, created_on, modified_on, name, optlock, parent_account_path, properties, created_by, id, modified_by, org_address_line_1, org_address_line_2, org_address_line_3, org_city, org_country, org_email, org_name, org_person_name, org_phone_number, org_state_province_county, org_zip_postcode, scope_id) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
	bind => [22 parameters bound]
Query: InsertObjectQuery(org.eclipse.kapua.service.account.internal.AccountImpl@68a0ef03)
	at org.eclipse.persistence.exceptions.DatabaseException.sqlException(DatabaseException.java:331)
	at org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor.executeDirectNoSelect(DatabaseAccessor.java:902)
	at org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor.executeNoSelect(DatabaseAccessor.java:964)
	at org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor.basicExecuteCall(DatabaseAccessor.java:633)
	at org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor.executeCall(DatabaseAccessor.java:560)
	at org.eclipse.persistence.internal.sessions.AbstractSession.basicExecuteCall(AbstractSession.java:2056)
	at org.eclipse.persistence.sessions.server.ClientSession.executeCall(ClientSession.java:306)
	at org.eclipse.persistence.internal.queries.DatasourceCallQueryMechanism.executeCall(DatasourceCallQueryMechanism.java:242)
	at org.eclipse.persistence.internal.queries.DatasourceCallQueryMechanism.executeCall(DatasourceCallQueryMechanism.java:228)
	at org.eclipse.persistence.internal.queries.DatasourceCallQueryMechanism.insertObject(DatasourceCallQueryMechanism.java:377)
[...]

PR #665, opened 2017-06-13
Fix/null created By

Coduz:
Hi @ctron 

This PR should solve #659 .

Can you please test it on your scenario?


Coduz:
313787f260dc136b99acbb28007f89db9f4d0815 is cherry-picked from another branch so it should disappear when that one is merged

PR #701, opened 2017-06-27
Add a gateway client SDK for Kapua

ctron:
This PR brings in a client SDK which allows to connect to Kapua in a modular way, using different transport mechanisms and communication profiles.


hekonsek:
+1 for merging. I especially like the Kura end-to-end example :) .

ctron:
If there are no objections I would like to merge this early next week.

ctron:
So as just discussed on the call, we will have a more detailed discussion about that next week in London.

PR #704, opened 2017-06-28
Added tests for password.java

LeoNerdoG:
Added tests that check functionality of password.java class.


ctron:
Is there any difference between `testPassword` and `testGetPassword`?

Maybe also a test with `null` values should be done. What do you think?

LeoNerdoG:
Hi @ctron ,
yes, I see. I copied the wrong file into passwordTest.java, because I have done this a while ago. I will change this, add tests for null and push the changes. 
regards,
Leo

ctron:
I would suggest to split off null and non-null tests to different methods

PR #708, opened 2017-06-29
Lockout and Expiration tests on user and credentials

muros-ct:
Added Expiration date and state for gherkin scenarios on User
and Credentials.
New feature file that tests Lockout and Expiration.

Two simple tests that test for status attribute on Credentials.

Updated existing credentials steps with status value.

Tests that check expiration date on Credentials and on User.

### DIFF
--- a/qa/src/test/java/org/eclipse/kapua/service/user/steps/TestCredentials.java
+++ b/qa/src/test/java/org/eclipse/kapua/service/user/steps/TestCredentials.java
@@ -11,6 +11,15 @@
  *******************************************************************************/
 package org.eclipse.kapua.service.user.steps;
 
+import org.eclipse.kapua.service.authentication.credential.CredentialStatus;
+
+import java.text.DateFormat;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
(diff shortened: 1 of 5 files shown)


qa/src/test/java/org/eclipse/kapua/service/user/steps/TestCredentials.java
```diff
@@ -35,4 +48,54 @@ public String getPassword() {
...
+
+    public CredentialStatus getStatus() {
+        if (isEnabled()) {
+            return CredentialStatus.ENABLED;
+        } else {
+            return CredentialStatus.DISABLED;
+        }
+    }
+
+    public Date getExpirationDate() {
```
ctron: Is there any reason why you are using "local" times and then converting to timezones. I would guess using plain Instants would be much simpler:

```java
Instant now = Instant.now();
now.minus(Duration.ofDays(1));
```

muros-ct: Much ceaner I agree.
I have changed it, it will be in next PR with more tests.
TNX, good code review

PR #709, opened 2017-06-29
Split up travis build in build and test phase

ctron:
This change will split up the build into different build stages on Travis.

The first step does a plain build without tests and without javadoc. Then javadoc and tests will be executed in parallel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,12 @@ jdk:
 
 sudo: false
 
+# The cache is stored per-branch
+
+cache:
+  directories:
+  - $HOME/.m2
+
 install:
   - echo ==== Setting up toolchain.xml ====
   - ls /usr/lib/jvm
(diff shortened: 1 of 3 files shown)


ctron:
You can have a look at https://travis-ci.org/ctron/kapua/builds/248353242 how this looks

lorthirk:
Is there a way to have the jobs show a description in the Travis page?

ctron:
Nope. The stages have names and are grouped by names. All jobs having the same name are executed in parallel.

Coduz:
@ctron  is this a fix for the build on Travis failing for build time limit exceeded?

ctron:
Well it does split up the build into different jobs. From what I know the timeout is limited to jobs not builds. So splitting up the build into different jobs should allow us to increase the overall build time.

Yet, if we run into the issue that one job exceeds 50 minutes we need to split it up again. But we could do this if we properly tag unit tests/integration tests and run those into different groups. We would then need to aggregate the jacoco execution data using the travis caching system.

Coduz:
@ctron  can you please rebase this, so then we can merge into develop?

ctron:
It is rebased. I would suggest to let it run through travis once more before merging it.

ctron:
The travis test looks good to me. Despite the constant failure of the kapua-qa project, which is nothing unusual at the moment.

So I would suggest to merge this.

PR #712, opened 2017-06-30
Credential Lockout tests

muros-ct:
New set of tests that cover credential lockout and failed login attempts.

Date calculation on tests was fixed on local date.

### DIFF
--- a/qa/src/test/java/org/eclipse/kapua/service/user/steps/TestCredentials.java
+++ b/qa/src/test/java/org/eclipse/kapua/service/user/steps/TestCredentials.java
@@ -16,8 +16,8 @@ import org.eclipse.kapua.service.authentication.credential.CredentialStatus;
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
-import java.time.LocalDateTime;
-import java.time.ZoneId;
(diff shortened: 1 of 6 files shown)


ctron:
That looks way better in my opinion. Big thanks!

muros-ct:
@lorthirk These are the basic Lockout and Expiration tests.
I can add more, just tell me which part should be covered with more tests.
Multiple combinations of statuses and dates could be tested.
Lockout scenarios in combination with statuses and expirations...

Most cucumber steps are ready, so it is easy to add scenarios.

muros-ct:
BTW, travis reported failure.
This is test regarding failed login in combination with lockoutPolicy.resetAfter which is potentialy issue.
@lorthirk will check

lorthirk:
Ok, I'm on it. 

Thanks!

lorthirk:
@Coduz do you have any idea about other tests we can perform?

muros-ct:
These are just basic tests that check required functionalities for Lockout and Expiration.
I would suggest that we merge my branch to yours and then merge to develop.
Then I would ask a colegue Leonardo to extend set of tests with more lockout/expiration combinations and test border scenarios.

What do you think?

PR #717, opened 2017-07-03
Cleanup PermissionFactory API

ctron:
This change cleans up the API to use default interface methods and to make use of var-arg arrays for the Action parameter. Allowing to create multiple permission objects in a single call, re-using the code which is already there.

### DIFF
--- a/service/security/registration/simple/src/main/java/org/eclipse/kapua/security/registration/simple/SimpleRegistrationProcessor.java
+++ b/service/security/registration/simple/src/main/java/org/eclipse/kapua/security/registration/simple/SimpleRegistrationProcessor.java
@@ -12,6 +12,10 @@
 package org.eclipse.kapua.security.registration.simple;
 
 import static java.util.Optional.empty;
+import static org.eclipse.kapua.service.authorization.permission.Actions.delete;
+import static org.eclipse.kapua.service.authorization.permission.Actions.execute;
+import static org.eclipse.kapua.service.authorization.permission.Actions.read;
+import static org.eclipse.kapua.service.authorization.permission.Actions.write;
 
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


service/security/registration/simple/src/main/java/org/eclipse/kapua/security/registration/simple/SimpleRegistrationProcessor.java
```diff
@@ -273,32 +277,16 @@ private User createUser(final String name, final String email, final String disp
...
-        permissions.add(permissionFactory.newPermission(new DeviceManagementDomain(), Actions.read, user.getScopeId()));
-        permissions.add(permissionFactory.newPermission(new DeviceManagementDomain(), Actions.write, user.getScopeId()));
-        permissions.add(permissionFactory.newPermission(new DeviceManagementDomain(), Actions.execute, user.getScopeId()));
-
-        permissions.add(permissionFactory.newPermission(new GroupDomain(), Actions.read, user.getScopeId()));
-
-        permissions.add(permissionFactory.newPermission(new RoleDomain(), Actions.read, user.getScopeId()));
-
-        permissions.add(permissionFactory.newPermission(new UserDomain(), Actions.read, user.getScopeId()));
+        permissions.addAll(permissionFactory.newPermissions(new AccountDomain(), user.getScopeId(), read));
```
Coduz: We want to restrict resources of the auto-registered account? Just wondering about if would be nice to give full access to all resources of the platform.

ctron: Well we can always discuss about adding more permissions to the newly created account, but I think this is off-topic for this PR.

PR #726, opened 2017-07-05
A new batch of datastore tests reimplemented with cucumber.

andrejn-ct:
## Simple integration tests for the DataStore service
This is the batch of DataStore tests, converted from jUnit to Cucumber implementation. At the moment all the existing jUnit tests are still present and enabled. Eventually all jUnit tests will be reimplemented by suitable cucumber based tests and the original jUnit tests will be removed.

There are no changes to pom files.

### Implementation changes
The implementation of the DataStore service has not been changed.

### Test changes
A new set of Cucumber based tests is implemented for the DataStore service.

### DIFF
--- a/qa/src/test/java/org/eclipse/kapua/service/datastore/steps/DataStoreServiceSteps.java
+++ b/qa/src/test/java/org/eclipse/kapua/service/datastore/steps/DataStoreServiceSteps.java
@@ -317,8 +317,6 @@ public class DataStoreServiceSteps extends AbstractKapuaSteps {
         KapuaDataMessage tmpMessage = (KapuaDataMessage) stepData.get(msgKey);
         StorableId storeId = insertMessageInStore(tmpMessage);
         stepData.put(idKey, storeId);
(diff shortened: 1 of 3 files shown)


Coduz:
HI @andrejn-ct ,

this PR is ready to be merged? It is near to 1000 lines of changes.

Regards, 

_Alberto)

qa/src/test/java/org/eclipse/kapua/service/datastore/steps/DataStoreServiceSteps.java
```diff
@@ -212,6 +321,17 @@ public void insertRandomMessageIntoDatastore(String msgKey, String idKey) throws
         waitABit(5000);
```
riccardomodanese: Hi Andrej, there are few waitABit() calls in the tests. Could you remove also all these wait call replacing them with the datastore synch call? (by calling refreshAllIndexes in the MessageStoreFacade)

PR #728, opened 2017-07-05
Exclude console when releasing

Coduz:
Hi all, 

This PR disable `console` and `assembly-assembly` module when using `maven-release-plugin` avoiding any action on console module when releasing.

Regards,

_Alberto_

### DIFF
--- a/assembly/pom.xml
+++ b/assembly/pom.xml
@@ -34,9 +34,21 @@
         <module>jetty-base</module>
         <module>sql</module>
         <module>broker</module>
-        <module>console</module>
         <module>api</module>
     </modules>
+    
+    <profiles>
+        <!-- Profile used to exclude console module on release process -->
+        <profile>
+            <id>console</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+            </activation>
+            <modules>
+                <module>console</module>
+            </modules>
+        </profile>
+    </profiles>
 
     <build>
         <pluginManagement>
--- a/pom.xml
+++ b/pom.xml
@@ -132,7 +132,7 @@
                     <version>${maven-release-plugin.version}</version>
                     <configuration>
                         <autoVersionSubmodules>true</autoVersionSubmodules>
-                        <releaseProfiles>javadoc,deploy,release,docker-push</releaseProfiles>
+                        <releaseProfiles>javadoc,deploy,release,docker-push,-console</releaseProfiles>
                         <goals>deploy</goals>
                     </configuration>
                 </plugin>
@@ -1003,7 +1003,6 @@
         <module>sso</module>
 
         <!-- Applications -->
-        <module>console</module>
         <module>rest-api</module>
 
         <!-- Simulator framework -->
@@ -1016,6 +1015,17 @@
     </modules>
 
     <profiles>
+        <!-- Profile used to exclude console module on release process -->
+        <profile>
+            <id>console</id>
+             <activation>
+                <activeByDefault>true</activeByDefault>
+            </activation>
+            <modules>
+                <module>console</module>
+            </modules>
+        </profile>
+    
         <!-- Profile used only during release process. -->
         <profile>
             <id>release</id>
--- a/pom.xml
+++ b/pom.xml
@@ -132,7 +132,7 @@
                     <version>${maven-release-plugin.version}</version>
                     <configuration>
                         <autoVersionSubmodules>true</autoVersionSubmodules>
-                        <releaseProfiles>javadoc,deploy,release,docker-push,-console</releaseProfiles>
+                        <releaseProfiles>javadoc,deploy,release,-console</releaseProfiles>
                         <goals>deploy</goals>
                     </configuration>
                 </plugin>
--- a/travis.sh
+++ b/travis.sh
@@ -29,7 +29,7 @@ HEARTBEAT_PROCESS_PID=$!
 ### Build itself
 
 "$M2_HOME"/bin/mvn -v
-"$M2_HOME"/bin/mvn -B -Dorg.eclipse.kapua.qa.waitMultiplier=2.0 -Dgwt.compiler.localWorkers=2 -Pjavadoc >> $OUTPUT 2>&1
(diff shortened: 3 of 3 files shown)


pom.xml
```diff
@@ -132,7 +132,7 @@
                     <version>${maven-release-plugin.version}</version>
                     <configuration>
                         <autoVersionSubmodules>true</autoVersionSubmodules>
-                        <releaseProfiles>javadoc,deploy,release,docker-push</releaseProfiles>
+                        <releaseProfiles>javadoc,deploy,release,docker-push,-console</releaseProfiles>
```
ctron: There are two ways to disable a profile "!" and "-", did you check that this is supported by the plugin?

assembly/pom.xml
```diff
@@ -34,9 +34,21 @@
...
         <module>broker</module>
-        <module>console</module>
         <module>api</module>
     </modules>
+    
+    <profiles>
+        <!-- Profile used to exclude console module on release process -->
+        <profile>
+            <id>console</id>
+             <activation>
```
ctron: There seems to be some indentation issue here. Should be a quick fix.

pom.xml
Coduz: Which plugin? `maven-release-plugin`?

Coduz: I've checked `maven-release-plugin` documentations and:

- `release:prepare` does not has `releaseProfiles` parameters ([link](http://maven.apache.org/maven-release/maven-release-plugin/prepare-mojo.html)) 
- `release:perform` has this configuration parameter ([link](http://maven.apache.org/maven-release/maven-release-plugin/perform-mojo.html)

So I think is supported but running `mvn release:prepare` will go through `kapua-console` module, while `mvn release:perform` it should not. 

Since `prepare` creates only git tag with dropped  `SNAPSHOT` version, I believe that this configuration is fine.

Regards, 

_Alberto_

ctron: I can't really find anything in the documentation. So either we check the source code of the plugin or give it a test run to verify. But we should do this before merging.

Coduz: I've checked source code of the `maven-release-plugin` for the `release:perform` goal in [here](http://svn.apache.org/viewvc/maven/release/tags/maven-release-2.5.3/maven-release-plugin/src/main/java/org/apache/maven/plugins/release/PerformReleaseMojo.java?revision=1708589&view=markup) and it seems that the only reference to `releaseProfiles` is in a parameter that is referenced into `getAdditionalProfiles()`.

In the [AbstractReleaseMojo](http://svn.apache.org/viewvc/maven/release/tags/maven-release-2.5.3/maven-release-plugin/src/main/java/org/apache/maven/plugins/release/AbstractReleaseMojo.java) the method `getAdditionalProfiles()` is invoked in the `createReleaseDescriptor()`.

So I can say that this should work, as the string of `additionalProfiles` is now manipulated or stripped.

@ctron  do you agree on this?

Coduz:
Hi @ctron ,

I have a question on this failure. For some reasons it seems to ignore `kapua-console` module.

I was able to reproduce this locally by adding `-Pjavadoc` option.

The situation is that `mvn clean install` (or also `mvn` as it is run on Travis) produces this reactor plan:

```
[...]
[INFO] kapua-assembly
[INFO] kapua-assembly-java-base
[INFO] kapua-assembly-jetty-base
[INFO] kapua-assembly-sql
[INFO] kapua-assembly-broker
[INFO] kapua-assembly-api
[INFO] kapua-console
[INFO] kapua-assembly-console
[INFO] kapua-dev-tools
[INFO] kapua-cucumber-reports
[...]
``` 

And `mvn clean install -Pjavadoc` produces this reactor plan:

```
[...]
[INFO] kapua-assembly
[INFO] kapua-assembly-java-base
[INFO] kapua-assembly-jetty-base
[INFO] kapua-assembly-sql
[INFO] kapua-assembly-broker
[INFO] kapua-assembly-api
[INFO] kapua-assembly-console
[INFO] kapua-dev-tools
[INFO] kapua-cucumber-reports
[...]
```

I have no idea why this is happening. 
With other profiles everything works fine. 

Do you have some hints on this? 

Regards, 

_Alberto_

Coduz:
It seems something related to this profile. 
```
        <profile>
            <id>javadoc</id>
            <build>
                <plugins>
                    <plugin>
                        <groupId>org.apache.maven.plugins</groupId>
                        <artifactId>maven-javadoc-plugin</artifactId>
                        <executions>
                            <execution>
                                <id>attach-javadocs</id>
                                <goals>
                                    <goal>jar</goal>
                                </goals>
                            </execution>
                        </executions>
                    </plugin>
                </plugins>
            </build>
        </profile>
```

If I change name and specify old `javadoc` profile it complains about not finding the profile. 
Even running with `javadoc1` profile, this behaviour is happening.

ctron:
Hm .. . Indeed, when you enable one profile manually, Maven does automatically disable all other profiles which are "activeByDefault=true".

We could try to go with a a system property instead, like:

```xml
<activation>
  <property>
    <name>!skipConsole</name>
  </property>
</activation>
```

Also see: https://maven.apache.org/guides/introduction/introduction-to-profiles.html

Coduz:
I thought about using `-Pjavadoc` disabling all `enabledByDefault=true` profiles, but when I use `-Pdev` this problem does not occur. 

One more strange thing is that when using `-Pjavadoc`, only `kapua-console` is ignored but not `kapua-assembly-console`, when both modules are under the same profile and are enabled by default. This is actually causing the error, because it excludes `kapua-console` but not `kapua-assembly-console`. 

I tried to force the `console` profile in `travis.sh` and it should work. I'm waiting for the build to finish.

Coduz:
Build has failed for build time exceeded. 
Maybe we can merge also #709 that could solve a bit this problem.
On my local:
`mvn -DskipTests -B -Dorg.eclipse.kapua.qa.waitMultiplier=2.0 -Dgwt.compiler.localWorkers=2 -Pjavadoc` this is not working

and this 

` mvn -DskipTests -B -Dorg.eclipse.kapua.qa.waitMultiplier=2.0 -Dgwt.compiler.localWorkers=2 -Pjavadoc,console` works.

So I can assume that this PR is safe to merge.

PR #730, opened 2017-07-06
Refactored MockedLocator with Guice proposal

muros-ct:
Based on UserService unit tests, changes are introduced to MockedLocator.
MockedLocator can for now be used as it was and in addition also by using
Google Guice DI.

UserService tests ware changed to work with Guice. If this is correct it will
be refactored on all service unit tests.

As part of commit, feature file for User Service was cleaned of excessive tabs
and spaces. No content changes were made.


muros-ct:
@stefanomorson can you check that this is right approach.
I did it now just for user service. If ok, all other services can be updated with new approach.

muros-ct:
Closing this PR. MockedLocator was refactored.
Overall refactoring by replacing KapuaLocator with Injection annotations was stoped, as codebase
has issues with mixing manual creation of objects with new() and injection. Injection is injection only and can not be mixed with this kind of object creation.

PR #742, opened 2017-07-13
Update kapua-template.yml

hinunbi:
Actually the possibility to run as root has been removed since Elasticsearch 5.0.
Elasticsearch should be run as 'elasticsearch' user.
see https://stackoverflow.com/questions/42807114/how-to-run-elasticsearch-5-2-1-as-root-user-in-linux-machine


ctron:
If you are running inside openshift, then the application isn't running as root in any case. It will be a random user ID and thus it also isn't possible to pick 'elasticsearch'.

PR #753, opened 2017-07-19
Console review fixes part1

### LINKED ISSUE #617: Button to Refresh lists in Data view
Hi all,

I think a Data View could use a Refresh button to refresh the lists available in the respective tab:

- Topics list and Metrics list in the By Topic tab
- Devices list and Metrics list in the By Device tab
- Devices list, Assets list and Channels list in the By Asset tab

Basically all the lists, and the results table, should be cleaned when pressing the button, and data should be immediately reloaded. This would be a UX improvement, since now a user needs to load another view and then come back to the Data view in order to refresh it.

bojanutkovic:

This resolves #617 among other issues


lorthirk:
This covers #617 among other issues

lorthirk:
After merging this PR in `develop` we discovered a few UI issues, so it has been reverted in order to fix those issues properly.

lorthirk:
PR #820 has been opened to supersede this one.

PR #758, opened 2017-07-25
Refactored MockedLocator with Guice on UserService

muros-ct:
Based on UserService unit tests, changes are introduced to MockedLocator.
MockedLocator can now be used as it was and in addition also by using
Google Guice DI.

UserService tests ware changed to work with Guice. This will be refactored
on all service unit tests.

As part of commit, feature file for User Service was cleaned of excessive tabs
and spaces. No content changes were made.

UserService refactored to use injected other services.


muros-ct:
@stefanomorson can you check it is ok, then I can proceed with other services. I think this part is pretty easy. Thou I can not use @Inject in classes that implement cucumber steps, as cucumber is also using guice for injection and also becuse I have to configure injection somewhere and  steps deffinition class is a startup class form cucumber point of view.

PR #760, opened 2017-07-26
Fix jpa configuration for named entity

Coduz:
Changed the JPA configuration for the AbstractKapuaNamedEntity.
Added new configuration to TravisCI YML file to force TravisCI itself to build with Ubuntu Precise images. With Trusty images we got 137 error on the build phase

### DIFF
--- a/commons/src/main/java/org/eclipse/kapua/commons/model/AbstractKapuaNamedEntity.java
+++ b/commons/src/main/java/org/eclipse/kapua/commons/model/AbstractKapuaNamedEntity.java
@@ -32,7 +32,7 @@ import org.eclipse.kapua.model.id.KapuaId;
 public abstract class AbstractKapuaNamedEntity extends AbstractKapuaUpdatableEntity implements KapuaNamedEntity {
 
     @Basic
-    @Column(name = "name")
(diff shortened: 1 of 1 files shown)


ctron:
As trusty is the new default we should consider fixing our issues in order not to run into any deprecation issues.

ctron:
Checking into the situation, I think we are running into the OOM situation again, just at a different location. Looking at the timetable of Travis, I strongly suggest against sticking to precise, as this will be effectively be the default by the end of August.

So we do need to fix this!

Coduz:
yes sure, but now I don't have time to fix the Travis CI build. 
As the precise is kept for the develop branch but all of the branch/PR builds now uses trusty all the PR on kapua will fail. 

By doing this I'm allowing to have PR validation, so we are sure that our PR does not break something. I will open a new issue to keep track of this problem and to switch to Trusty machines.

ctron:
The fix seemed to be rather simple: #761

So for this looks good: https://travis-ci.org/ctron/kapua/builds/257706210

Coduz:
Good for you then :)

ctron:
So I suggest to stick with trusty and merge #761

PR #762, opened 2017-07-27
Fix on failing QA tests - wait for embedded servers

muros-ct:
QA tests were failing because datastore - ES or broker was not started
as part of Before hooks in cucumber scenarios.
For that purpose delays were introduced while starting embedded servers.
This solved problem and is configurable from system properties which can
also be set in cucumber tests runners.

Datastore test were changed by deleting indices as part of scenario steps
and not in before and after hooks. This was done because startup issues,
datastore is not yet started when hook is executed.

I know that this is not perfect solution to a problem and it also increases
the execution time of QA tests. Suggestions?


muros-ct:
Changes have been made to integration tests to speed things up.
Embeded broker and datastore are started just onece per feature file and not for each scenario.

PR #765, opened 2017-08-01
remove duplicated fields already present in Kura message object

riccardomodanese:
removed duplicated code in Kura data message tha is already present in the Kura message parent.


Coduz:
Have you checked that REST API and Swagger still running fine?

riccardomodanese:
I asked @lorthirk to do that

PR #770, opened 2017-08-08
Second batch of cucumber based datastore tests.

andrejn-ct:
## Simple integration tests for the DataStore service
This is a new batch of DataStore tests, converted from jUnit to Cucumber implementation. At the moment all the existing jUnit tests are still present and enabled. 
With this batch, all existing jUnit test scenarios should be covered. A closer look should be given to these cucumber tests and if OK the old jUnit tests should be removed, since the test scenarios are now duplicated.

There are no changes to pom files.

### Implementation changes
The implementation of the DataStore service has not been changed.

### Test changes
A new set of Cucumber based tests is implemented for the DataStore service.

### DIFF
--- a/qa/src/test/java/org/eclipse/kapua/service/datastore/steps/DataStoreServiceSteps.java
+++ b/qa/src/test/java/org/eclipse/kapua/service/datastore/steps/DataStoreServiceSteps.java
@@ -62,6 +62,7 @@ import org.eclipse.kapua.service.datastore.internal.DatastoreCacheManager;
 import org.eclipse.kapua.service.datastore.internal.MetricInfoRegistryServiceProxy;
 import org.eclipse.kapua.service.datastore.internal.mediator.ChannelInfoField;
 import org.eclipse.kapua.service.datastore.internal.mediator.ClientInfoField;
+import org.eclipse.kapua.service.datastore.internal.mediator.DatastoreChannel;
 import org.eclipse.kapua.service.datastore.internal.mediator.DatastoreMediator;
 import org.eclipse.kapua.service.datastore.internal.mediator.MessageStoreConfiguration;
 import org.eclipse.kapua.service.datastore.internal.mediator.MetricInfoField;
@@ -69,6 +70,7 @@ import org.eclipse.kapua.service.datastore.internal.model.DataIndexBy;
 import org.eclipse.kapua.service.datastore.internal.model.MetricsIndexBy;
 import org.eclipse.kapua.service.datastore.internal.model.StorableIdImpl;
 import org.eclipse.kapua.service.datastore.internal.model.query.AndPredicateImpl;
+import org.eclipse.kapua.service.datastore.internal.model.query.ChannelMatchPredicateImpl;
 import org.eclipse.kapua.service.datastore.internal.model.query.RangePredicateImpl;
 import org.eclipse.kapua.service.datastore.internal.model.query.TermPredicateImpl;
 import org.eclipse.kapua.service.datastore.internal.schema.MessageSchema;
@@ -84,6 +86,7 @@ import org.eclipse.kapua.service.datastore.model.StorableId;
 import org.eclipse.kapua.service.datastore.model.StorableListResult;
 import org.eclipse.kapua.service.datastore.model.query.AndPredicate;
 import org.eclipse.kapua.service.datastore.model.query.ChannelInfoQuery;
+import org.eclipse.kapua.service.datastore.model.query.ChannelMatchPredicate;
 import org.eclipse.kapua.service.datastore.model.query.ClientInfoQuery;
 import org.eclipse.kapua.service.datastore.model.query.MessageQuery;
 import org.eclipse.kapua.service.datastore.model.query.MetricInfoQuery;
@@ -104,6 +107,7 @@ import org.slf4j.LoggerFactory;
 import cucumber.api.Scenario;
 import cucumber.api.java.After;
 import cucumber.api.java.Before;
+import cucumber.api.java.en.And;
 import cucumber.api.java.en.Given;
 import cucumber.api.java.en.Then;
 import cucumber.api.java.en.When;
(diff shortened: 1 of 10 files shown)


qa/src/test/resources/features/datastore/Datastore.feature
```diff
@@ -331,6 +331,270 @@ Feature: Datastore tests
...
+    And The metric "tst-metric-3" was last published in the list "AccountMetrics" on "03/07/2017T09:00:20"
+    And The metric "tst-metric-4" was last published in the list "AccountMetrics" on "03/07/2017T09:00:20"
+    And All indices are deleted
+
+  Scenario: Account based ClientInfo data check
+    Check the correctness of the client info data stored by retrieving the client information by account.
+
+    Given I login as user with name "kapua-sys" and password "kapua-password"
+    And Account for "kapua-sys"
+    And The device "test-device-1"
```
riccardomodanese: may we miss two conditions around this line (see https://github.com/eclipse/kapua/blob/develop/service/datastore/internal/src/test/java/org/eclipse/kapua/service/datastore/internal/MessageStoreServiceTest.java#L1070)

+ And Client "test-client-1" last published a metric in the list "AccountMetrics" on "03/07/2017T09:00:00" 
+ And Client "test-client-2" first published a metric in the list "AccountMetrics" on "03/07/2017T09:00:00"

qa/src/test/resources/features/datastore/Datastore.feature
```diff
@@ -331,6 +331,270 @@ Feature: Datastore tests
...
+      |tst-metric-3  |string    |123        |
+      |tst-metric-4  |bool      |true       |
+    Then I store the messages from list "TestMessages" and remember the IDs as "StoredMessageIDs"
+    And I refresh all database indices
+    When I query for the current account metrics and store them as "AccountMetrics"
+    Then There are exactly 4 metrics in the list "AccountMetrics"
+    And The metric info items "AccountMetrics" match the prepared messages in "TestMessages"
+    And All indices are deleted
+
+  Scenario: MetricsInfo client ID and topic data based on the client id
```
riccardomodanese: the metrics are different (see https://github.com/eclipse/kapua/blob/develop/service/datastore/internal/src/test/java/org/eclipse/kapua/service/datastore/internal/MessageStoreServiceTest.java#L1151)
TestMessage2 should have the same metric names of the TestMessages[1] but the values of the TestMessages[0]
The idea under this test is to mix the metrics type using the same metric name. So the client-1 should publish a message with the same metric names of one already published by the client-0, but with different value type.

qa/src/test/java/org/eclipse/kapua/service/datastore/steps/DataStoreServiceSteps.java
```diff
@@ -479,6 +551,42 @@ public void queryForAccountChannelInfoByDateRange(String start, String end, Stri
...
+        assertEquals("Wrong number of topics found!", tmpList.getSize(), topics.size());
+
+        for (ChannelInfo tmpInfo : tmpList.getItems()) {
+            infoTopics.add(tmpInfo.getName());
+        }
+        for (TestTopic tmpTop : topics) {
+            listTopics.add(tmpTop.getTopic());
+        }
+
+        for (String tmpTopic : listTopics) {
```
riccardomodanese: I think we can remove the lines 581/583 since we can use the topics list into this loop

qa/src/test/java/org/eclipse/kapua/service/datastore/steps/DataStoreServiceSteps.java
```diff
@@ -580,13 +703,88 @@ public void queryForTopicMetrics(String topic, String lstKey) throws KapuaExcept
...
+            }
+        }
+        fail(String.format("No metric matches the client id [%s]", clientId));
+    }
+
+    @Then("^Client \"(.+)\" last published a metric in the list \"(.+)\" on \"(.+)\"$")
+    public void checkLastPublishDateOfClientMetric(String clientId, String lstKey, String date) throws Exception {
+
+        MetricInfoListResult metList = (MetricInfoListResult) stepData.get(lstKey);
+        Date tmpCaptured = parseFeatureDateTimeString(date);
```
riccardomodanese: I was looking at that code and may I missed this check in a previous pr review.
The date parsing should be alligned to the Kapua datastore format so can you change this method implementation to use:
https://github.com/eclipse/kapua/blob/develop/commons/src/main/java/org/eclipse/kapua/commons/util/KapuaDateUtils.java#L57
With this change the incoming dates should be changed (according to the pattern)
May be better if we address this fix in a new pr, what do you think?

qa/src/test/resources/features/datastore/Datastore.feature
```diff
@@ -331,6 +331,270 @@ Feature: Datastore tests
...
+    And I set the following metrics to the message 3 from the list "TestMessages"
+      |metric        |type      |value      |
+      |tst-metric-3  |string    |123        |
+      |tst-metric-4  |bool      |true       |
+    Then I store the messages from list "TestMessages" and remember the IDs as "StoredMessageIDs"
+    And I refresh all database indices
+    When I query for the current account metrics in the date range "03/07/2017T09:00:00" to "03/07/2017T09:00:20" and store them as "AccountMetrics"
+    Then There are exactly 4 metrics in the list "AccountMetrics"
+    And The metric info items "AccountMetrics" match the prepared messages in "TestMessages"
+    And Client "test-client-1" first published a metric in the list "AccountMetrics" on "03/07/2017T09:00:00"
```
riccardomodanese: may we miss two conditions around this line (see https://github.com/eclipse/kapua/blob/develop/service/datastore/internal/src/test/java/org/eclipse/kapua/service/datastore/internal/MessageStoreServiceTest.java#L1070)

    And Client "test-client-1" last published a metric in the list "AccountMetrics" on "03/07/2017T09:00:00"
    And Client "test-client-2" first published a metric in the list "AccountMetrics" on "03/07/2017T09:00:00"

andrejn-ct:
Hi! Yes, I'm working on your suggestions.

PR #772, opened 2017-08-10
Porting of fixes from 0.2.0 into develop

Coduz:
Hi all, 

while I was working on the 0.2.0 release I made various fixes.
Fixes involve module layout changes and configuration of plugins.

I committed them directly on branch `release-0.2.x` just for ease of development.

With this PR I port those changes into `develop` branch.

I'll add another comment to this PR explaining requirements that caused the commit.

### DIFF
--- a/assembly/pom.xml
+++ b/assembly/pom.xml
@@ -41,9 +41,6 @@
         <!-- Profile used to exclude console module on release process -->
         <profile>
             <id>console</id>
-            <activation>
-                <activeByDefault>true</activeByDefault>
-            </activation>
             <modules>
                 <module>console</module>
             </modules>
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,10 @@ jobs:
     - stage: build
       script:
         - $M2_HOME/bin/mvn -v
+        - $M2_HOME/bin/mvn -B -f external/pom.xml install
+    - stage: build
+      script:
+        - $M2_HOME/bin/mvn -v
         - $M2_HOME/bin/mvn -B -Dgwt.compiler.localWorkers=2 -DskipTests install
     - stage: test
       script:
--- a/assembly/api/pom.xml
+++ b/assembly/api/pom.xml
@@ -326,7 +326,7 @@
 
     <profiles>
         <profile>
-            <id>deploy</id>
+			<id>deployVagrant</id>
             <activation>
                 <activeByDefault>false</activeByDefault>
             </activation>
--- a/assembly/api/pom.xml
+++ b/assembly/api/pom.xml
@@ -326,7 +326,7 @@
 
     <profiles>
         <profile>
-			<id>deployVagrant</id>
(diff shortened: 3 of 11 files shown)


Coduz:
961c656 - Updated console profile for release

Since there was two profiles named `console`, one on the root `pom.xml` and one on the `console/pom.xml` and the one on the `console/pom.xml` was the only profile available,  the second one was always active, even in `release:prepare` goal. 
This was causing a build error since the Hudson build does not have access to the GXT external dependencies. 


fdb3a0c - Moved external module to be a separate maven project

This was required for the fact that `release:prepare` goal was updating dependencies version. So also this external dependencies had the version changed to 0.2.0 instead of keeping their own version. This was causing a missing dependency error for REST API module.


6d22950 - Fixed build-tools module configuration to skip deployment

Since build-tools does not have `kapua` module as its parent, it was not inheriting deployment configuration. Instead of copying the `kapua` module configuration I just skipped deployment of the `build-tools` module as we don't need to distribute also that one.

67c1fee - Bumped maven-gpg-plugin version to 1.6

This has solved a problem when signing artifacts with the `maven-gpg-plugin`. This problem was showing in 1.5. Upgrading the version to 1.6 solved it.


1bcbc11 - Fixed profile duplicate profile name that was interfering with the release deploy profile

We had two `deploy` profiles. One on  the root `pom.xml` file (with plugins required for the actual deployment` and one in the `dev-tools` module that was used by vagrant scripts to deploy the vagrant machine. The `deploy` profile in the `dev-tools` module was accidentally activated and it was trying to deploy file into vagrant while releasing, therefore failing. 

30ab1f6 - Fixed checkstyle errors

This is not on the 0.2.0, but is just fixing build after cherry pick

assembly/pom.xml
```diff
@@ -41,9 +41,6 @@
         <!-- Profile used to exclude console module on release process -->
         <profile>
             <id>console</id>
-            <activation>
```
ctron: Do you know why it didn't work disabling the profile with `!console`?

ctron:
My suggestion would be to split this up into different PRs. Or cherry pick the straight forward commits like 67c1fee ... I think that would simplify the discussion.

assembly/pom.xml
Coduz: Because only `release:perform` takes in consideration additional profiles.
`release:prepare` uses the profiles given when you run the commands.

stefanomorson:
I guess the points under discussion are the disabling of the console profile and how to manage the external module (swagger). 

If so, what if we merge the pull request and add in any case the issues for fixing/improving the points above ?
Going that way we'll have something working and in the same time we can proceed in parallel with the fixes/improvements. 

For the swagger-ui thing for example, I would consider also the possibility to review what we did and use the official version from maven instead the of patched version (we'll likely need a new CQ). That would remove the need to embed the code into the project and make the build simpler.

PR #773, opened 2017-08-11
Chng connection user coupling

### LINKED ISSUE #663: Enable support for Device User Coupling
Hi all,

we need to introduce the user bound connection feature.

This is a feature that can limit/enforce the bound between a Device-connection and the User used to connect to the platform.

To introduce this feature requires the following steps:

1. Add DeviceConnectionService new configuration `deviceConnectionUserCouplingDefaultMode` of type `String` with options `LOOSE` or `STRICT` 
2. Add new field in DeviceConnection `reservedUserId` that will be the field in which we will store the reserved user id used to connect to the platform (when in STRICT mode)
3. Add new field in DeviceConnection `userId` that will be the field in which we will store the last user id used to successfully connect to the platform
4. Add new boolean field on DeviceConnection `allowUserChange` to allow the connection to change the user (if in STRICT mode and no `reservedUserId` is set, a different user than one stored in the `userId` field, is allowed to connect)
5. Add (or rename) new field on DeviceConnection `userCouplingMode` with option values `INHERIT` (takes the default value configured by the account) `LOOSE` or `STRICT`. See below per meanings.
6. Add checks on broker connection login to enforce the bound according to the configurations.

Possible configurations are (if the `userCouplingMode` value is INHERITED then the parameter is retrieved by the service configuration):

| reservedUserId | allowUserChange | STRICT                                                                | LOOSE                     |
|----------------|-----------------|-----------------------------------------------------------------------|---------------------------|
| != null        | true/false      | device must use the user set and it must be the only one that uses it | device can use any user   |
| == null        | false           | device must use the user `userId`                                       | device can use any user   |
| == null        | true            | device can change the user regardless of the `userId` set               | device can use any user   |

**NOTE - Regardless of what summarized in the table above, these conditions must be enforced always:**
1. the allowUserChange must be reset every login by the security plugin (a connection is allowed to change the user once).
2. the security plugin checks for every login attempts that the user is not set as reservedUsedId for any device (except for itself) regardless of the userCouplingMode value set for that connection
3. reservedUsedId values must be not duplicated. The system must enforce that.

riccardomodanese:
fix #663


Coduz:
Full green 👍

console/src/main/resources/org/eclipse/kapua/app/console/client/messages/ValidationMessages.properties
```diff
@@ -282,13 +282,12 @@ attachmentSize=Attachment Size
...
 #
-devInfo=Cloud Connection Information
+devInfo=Device Information
+connInfo=Connection Information
 devAttributesInfo=Attributes Information
-devSecurity=Device Credentials Security
 devHw=Hardware Information
 devSw=Software Information
 devJava=Java Information
-netInfo=Network Information
```
lorthirk: This is wrong. The `netInfo` message is used in the Device Details tab.

console/src/main/resources/org/eclipse/kapua/app/console/client/messages/ValidationMessages.properties
```diff
@@ -301,11 +301,8 @@ devCustomAttribute3=Custom Attribute 3
 devCustomAttribute4=Custom Attribute 4
 devCustomAttribute5=Custom Attribute 5
 
-devSecurityCredentialsTight=Credentials Tight Strategy
-devSecurityAllowCredentialsChange=Allow Credentials Change
 
 devClientId=Client ID
-devConnectionStatus=Connection Status
```
lorthirk: This string is needed

PR #774, opened 2017-08-11
Refactored Datastore service and its tests from locator to injection

muros-ct:
Datastore service and its unit tests were refactored to injection.
Services that datastore service and its supporting classes are using are
injected.

This refactoring needs careful review, regardless that all the tests are
passing.

Unit tests are still using Locator for services used in unit tests.


muros-ct:
This PR is closed as refactoring from Locator to Injection is not possible at the moment.
Simple replacement of KapuaLocator with Inject annotated services is not possible, as there are
cases wher new objects are manually created and this breaks injection procedure.

More complex refactoring of codebase will be needed.

Issue #751, opened 2017-07-17
Document how to connect Kura to Kapua


ctron:
There should be some guide we can point people towards when they want to configure Kura to connect to Kapua.

ctron:
Also see: https://dentrassi.de/2017/02/14/testing-kapua-with-simulated-kura-gateways/#comment-19043

LeoNerdoG:
@ctron I can make such documentation -  I can describe how to start Kapua (although already described) and connect it to Raspberry Pi. Where do I upload this document (or whom do I send it)? 
BR,
Leonardo

LeoNerdoG:
Hi @ctron,
I have created PR PR#780 which contains steps for Kura -Kapua connection. 
This is the first version, it will be updated.

Regards,
Leonardo

riccardomodanese:
closed via #780

PR #780, opened 2017-08-22
Documents with Kura Kapua connection tutorial

LeoNerdoG:
I have created document which describes Kura - Kapua connection procedure.
This is just first bigger draft as more and more people are asking how to
connect those two. I will enhance it as the time goes. 
If somebody finds mistakes, please correct them. 

This PR is "answer" to Issue #751 . 

### DIFF
--- a/docs/kuraKapuaDocs.md
+++ b/docs/kuraKapuaDocs.md
@@ -1,0 +1,191 @@
+# Kura - Kapua connection
+
+This is a short introduction into how locally connect Kura and Kapua. Although some of the info bellow can already be found on Kapua github project and Kura guides, this document contains everything needed for the connection. If you have any questions post them on [Eclipse Community Forums](https://www.eclipse.org/forums/index.php/f/340/) or join our [mailing list](https://dev.eclipse.org/mailman/listinfo/kapua-dev). In case you find any issues, [report them](https://github.com/eclipse/kapua/issues). Also if you want to contribute, [this is the place to start!](https://github.com/eclipse/kapua).
+
+## Usefull links
+
+ - [Kapua quick start guide](https://github.com/eclipse/kapua/blob/develop/README.md)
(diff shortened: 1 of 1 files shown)


LeoNerdoG:
Hi @Coduz ,
can you (or somebody else) please review and merge this document as it contains info regarding Kura-Kapua connection? I saw the issues regarding this matter and a lot of forum questions so I hope this helps. 

Regards,
Leonardo

ctron:
Thanks for the PR. I only see one issue with this. We already have two locations which describe how to start Kapua using docker. Adding a third location would make it more error prone once we begin changing this.

I would suggest that we consolidate that. One section in the documentation should describe the process, others should only reference it.

What do you think?

LeoNerdoG:
Hi @ctron ,
I am aware of that fact, but I tried to put myself into ordinary user which just tries to link Kura and Kapua. I know, that we are doubling or tripling the same document but I just wanted to make a guide, which contains everything to set up the desired connection. This way I think it is easier for a beginner to link the two. I hope you understand my dilemma. 
But yes, I agree that documentation should be revised and renewed as soon as possible, because competitive projects have much better documentation, guides and tutorials. 

Regards,
Leo

riccardomodanese:
Hi all, I'mo going to merge this pr. I'm aware of the changes requested by @ctron to organize the guide in a better way (avoiding duplication and so on).
So I'm going to open a new issue to track the reorganization of this guide ( #802 ) .

Regards

afterski:
I've tested and it seems to work fine using the default user in kapua.
The information I needed (missing account name) was:
topic.context.account-name: kapua-sys
username: kapua-broker
password: kapua-password

However serializing objects are not allowed in kapua (dec 2017).
http://activemq.apache.org/objectmessage.html

so to get data in the DB
-Dorg.apache.activemq.SERIALIZABLE_PACKAGES=*
needs to be added to
ACTIVEMQ_OPTS in kapua-broker

This is obviously not a production solution but it gets You started with development.

Cheers, Björn

PR #781, opened 2017-08-24
Refactor Device view to use EntityView template

elbert3:
This change refactors the Device view to extend the EntityView template like the other menu items.


elbert3:
Hi @Coduz, thanks for the input.  I've updated the Dialogs to use the xsrfToken from the parent class as you suggested, and removed the ContentPanel that was not being used.  Let me know if you see anything else.

PR #784, opened 2017-08-25
Console modularity

lorthirk:
# Console Modularity

Hi all,

This PR refactors the GWT Console to leverage GWT modules, and allows for extensibility of the console itself. 

So, while the GWT Console was a monolithic GWT application before, with this PR I splitted functionalities (according to our current service boundaries) across several modules. Also, this approach enables for new modules to contribute new pieces of UI as well, if needed.

## Module structure changes

The console project has been split in the following modules:

- Module API: a module containing all the classes and interfaces needed by other modules to contribute additional views to the console
- Feature Modules: all the previously available functionalities, split in separate modules
- Core Module: All the Console logic files not strictly tied to a single functionality (e.g. the website structure, the login page, the container for the left and upper menus, etc.)
- Web Module: All the Console Web files (e.g. web server descriptors, servlet definitions, CSS, etc.)

## Extending the UI

This PR introduces two approaches to customize the UI in the Console: provide new Main Views and provide new Tabs in existing views.

### Provide new Main Views

To provide a new Main View the following steps are needed:

1. Subclass `AbstractMainViewDescriptor`, implementing the following methods:
    - `String getViewId()`: used to retrieve an id to manage the view programmatically, e.g. managing the click event from `WestNavigationMenu`
    - `IconSet getIcon()`: used to retrieve the icon for the menu item
    - `Integer getOrder()`: used to retrieve an index to sort the menu items
    - `String getName()`: used to retrieve the text of the menu item
    - `View getViewInstance(GwtSession currentSession)`: used to retrieve a new instance of the actual view to show on main container
    - `Boolean isEnabled(GwtSession currentSession)`: used to indicate whether the menu item should be showed on the menu or not (e.g. checking the current user for appropriate permissions)
1.  Subclass `AbstractView` to provide the actual view. You can also choose to subclass `AbstractEntityView` which is a specialized subclass of `AbstractView` to show `KapuaEntity` collections.
1.  Edit the file `<kapua root>/console/web/src/main/webapp/WEB-INF/view-descriptors.txt` and enter the full qualified class name of the `AbstractMainViewDescriptor` created at step 1

### Provide new Tab Views to existing Main Views

Providing new Tab Views to existing Main Views is similar to providing new Main Views. The only differences are in the classes being subclassed and in the format of the `tab-descriptors.txt` file.

1. Subclass `AbstractTabDescriptor`, implementing the following methods:
    - `String getViewId()`: used to retrieve an id to manage the view programmatically
    - `Integer getOrder()`: used to retrieve an index to sort the menu items
    - `<M extends GwtEntityModel, V extends , AbstractEntityView<M>, T extends KapuaTabItem<M>> T getViewInstance(V view, viewGwtSession currentSession)`: used to retrieve a new instance of the actual tab view to show on the lower part of the main container
    - `Boolean isEnabled(GwtSession currentSession)`: used to indicate whether the tab item should be enabled or not (e.g. checking the current user for appropriate permissions)

    In contrast with `AbstractMainViewDescriptor`, the name and the icon of the tab are specified in the related `TabItem` for legacy reasons.

1.  Subclass `KapuaTabItem<M extends GwtEntityModel>` to provide the actual tab item
1.  Edit the file `<kapua root>/console/web/src/main/webapp/WEB-INF/tab-descriptors.txt` and enter a new row with the syntax `tab.descriptor.full.qualified.class.Name->view.full.qualified.class.Name`
------
Being this a pretty big set of changes, I encourage anyone to give their feedback. Thanks!


ctron:
Is there a reason why a new system for looking up implementations has been created. Would it be possible to use the Java ServiceLoader instead?

lorthirk:
Unfortunately, the Java ServiceLoader is not available in the GWT Emulation Library 😕

lorthirk:
So, I thought a bit more about this. What I said in my previous comment is basically not correct, since the view lookup for both menu items and tabs is performed in a backend service, so there should be no problem with ServiceLoader there. However, while that might fit the `ViewDescriptor` lookup, it still wouldn't help us in the `TabDescriptor` lookup where I must differentiate in which view every tab must appear.

ctron:
So at least 50% would be solved with standard tools then 😉 But I think you could to the rest with the service lookup. You could annotate your tab with e.g a `@View("myView)` annotation. That should do the trick and you would be using 100% Java functionality.

stefanomorson:
A also agree that views can be loaded with service loader. 
However the file is still useful to specify which views should be showed and which not and the same can be said for the tabs within a view. This will give some more flexibility.
However I would prefer to define this in a single JSON file instead of two with a custom format.

lorthirk:
So, I addressed your concerns in my last two commits. Now the `MainViewDescriptor`s are loaded with Java `ServiceLoader` while `TabDescriptor`s associations to tabs are written in a standard JSON file. Such file is a JSON object whose keys are the views fully qualified class names and their related values are arrays containing the fully qualified names of `TabDescriptor` files.

PR #788, opened 2017-08-29
DB migration enhancement

elbert3:
This makes some improvements to the way liquibase is used in Kapua:
- A 'logicalFilePath' is specified for each changeset so that the change can be uniquely identified regardless of the file's actual path.
- The Rest API and Admin console are now configured to (optionally) run the liquibase client at startup, allowing them to be started without requiring the db to be seeded externally.
- The liquibase update is now disabled by default, but can be enabled using the `commons.db.schema.update` system setting.
- All of the SQL scripts have been converted to use liquibase's XML format to make it db agnostic.


stefanomorson:
HI,

Some feedback:

* I see that usually, within the same xml file, the table creation and the seeding are located in two distinct change sets. This is not the case for example in:
  * ./0.2.0/athz_domain_actions.xml
  * ./0.2.0/athz_role.xml. Is this intentional?
* Since the default configuration for the db update on component startup has changed to disabled, demo/develop deployments should be changed to enable the db update on startup:
  * Docker deployment
  * OpenShift deployment
  * Vagrant deployments
Also the documentation and (README) for launching the Docker containers have to be changed consistently

muros-ct:
@elbert3 Could you rebase to develop.
I just can't find a reason why tests are failing in your PR and not on develop.

PR #807, opened 2017-09-12
Added positive value check on BIGINT UNSIGNED fields

### LINKED ISSUE #207: Database schema: General issue with IDs
There is a general issue regarding the ID fields in the database.
The table creation scripts define all IDs as BIGINT(21) UNSIGNED (i.e. **scope_id    BIGINT(21)  UNSIGNED NOT NULL**).
Nevertheless it is possible to create an ID from a negative integer (i.e. **new KapuaEid(BigInteger.valueOf(-1))**) and the database will not complain. Searching the database with such ID is also successful.
This is somewhat counter-intuitive, as the field definition seems to imply that IDs with negative values should be rejected by the database.

Coduz:
Hi all, 

this should fix #207 

Regards, 

_Alberto_


Coduz:
Fixed `CHECK` clause adding `>=0`

Issue #649, opened 2017-06-08
ArgumentValidator regular expressions problem


LeoNerdoG:
While testing IP and local IP regular expressions (listed bellow) in ArgumentValidator I found that tests do not fail properly due to ((.*)$) at the end. Obviously dollar sign ($) marks end of expression but why is there (.*) at the end? 
.* = matches any char zero or more times. 

If regular expressions contain this expression, tests do not fail if IP addresses are e.g. "http://10.10.10.19999999". 

If I remove (.*) then tests pass/fail as expected. 

IP_regexp = "(^(http://)|(https://)|())([01]?\\d\\d?|2[0-4]\\d|25[0-5])\\.([01]?\\d\\d?|2[0-4]\\d|25[0-5])\\.([01]?\\d\\d?|2[0-4]\\d|25[0-5])\\.([01]?\\d\\d?|2[0-4]\\d|25[0-5])((.*)$)";

Local_IP_regexp = "(^(http://)|(https://))((127\\.0\\.0\\.1)|(10(\\.[0-9]+){3,3})|(172(\\.[0-9]+){3,3})|(192\\.168(\\.[0-9]+){2,2}))((.*)$)";

PR #808, opened 2017-09-12
Added test for Local IP and IP address

LeoNerdoG:
Added tests that check if (local) IP is in proper format
and if there are any prohibited strings in it.


riccardomodanese:
fix #649

PR #865, opened 2017-10-13
Splitting of tests into profiles

muros-ct:
Integration and cucumber unit tests are annotated with cucumber tags.
This enables us to run different batches of tests during development and
final testing. Build process is thereby optimized.

By using this tags, multiple sets of tests can be run on Travis CI in parallel.

Changes were made to cucumber feature files and master pom file.

Added jobs to test stage in travis. This jobs run tests in parallel.


muros-ct:
@ctron can you please check how this spliting might effect codecov, as tests are run in multiple paralel jobs.

In genaral I see it runs nice and it helps while runing tests locally as basic run on my machine is now 10 minutes insted of half an hour. Developer can also select set of tests that he/she wants to run and not full integration batch.

Full integration can be run onece a night or even each time by providing -PtestAll

Any suggestions welcome.

ctron:
I don't know how codecov can handle this myself. I guess someone needs to figure that out. Maybe they have some documentation, or doing a simple example.

About this PR I would like to understand:
1) Are all tests run by Travis, or just a subset?
1) Why did the test still fail?
1) How to you control what is executed?

What I don't like, and I think this could simply be dropped, is the mapping profiles to tags 1:1. You can control the cucumber tags using `-Dcucumber.options`. And I think a default plus the "all" profile is fine. But for the rest, I would fall back to the plain properties.

muros-ct:
ad.1.
Yes all tests are run by Travis but in parallel. Each job runs one profile, and that profile can have 1..* cucumber tags. This is why I used profiles.
And yes I could also be done with cucumber.options, and it acutually is in background.

ad.2.
When I first created this branch, tests ware successful each time, but now after couple rebases to develop. Promblems arose again. It has to do something with new tests. I will investigate, because
tests are failing sporadically. It is hard to identify failure, but it will be easier with spliting of tests, because it can be limited to those that are frequently failing – like datastore tests.
I allready did some fixes to those tests but it is in feature branch fix-securityPlugin and I hope it is merged soon. Many tests are allready fixed but not merged and together with this change it will get more and more
complicated to merge and make every thing work again.

ad.3.
Controll of test execution is by maven profiles specified in master pom. Yes it is somehow hidden, but again it should be managed by just couple of “lead” testers, others should just tag the tests properlly.

ctron:
Ok, I understand.

But I still don't see the benefit of re-creating the tags with profiles. I think in travis you can directly put in the tags. Otherwise, adding a new tag, also requires adding a new profile as well. Making things more complicated.

So once we figured out how codecov can handle this, we are good to go from my side.

muros-ct:
Ok, I will change it to cucumber tags and investigate codecov?

PR #898, opened 2017-10-23
Improved exception handling in service unit tests

andrejn-ct:
## A refactoring of the exception handling in service unit tests
The exception handling in most service unit tests was improved.
When an expected exception is caught, the type and message are logged in the cucumber html report. If, on the other hand, an unexpected exception is caught, the exception is rethrown and the full stack trace is logged to help diagnose the underlying problem.

No changes.

### Implementation changes
No changes.

### Test changes
The test steps were modified to better handle exceptions.


Coduz:
Hi @andrejn-ct ,

we need to check if this needs to go under Eclipse Evaluation

Regards, 

_Alberto_

Coduz:
@andrejn-ct Nevermind, it should be ok.

Issue #892, opened 2017-10-19
Cannot Add Credentials to Users


LeoNerdoG:
When creating new user (as kapua-sys or in any other child account) we can add roles, permissions but no credentials. User can click on Credentials tab, but it looks like everything is frozen. Mouse pointer does change from mouse to hand-like pointer, but "add" button (and all others) is unclickable. 
(Also we cannot add Credentials to default users: kapua-broker and kapua-sys). 

Testflow: 
1) Login as kapua-sys
2) Go to Users and try to add Credentials to kapua-broker/kapua-sys or
3) Create new user and try to add Credentials to it. 
**It should not be possible due to bug in Kapua.**

<img width="1440" alt="screen shot 2017-10-19 at 15 22 59" src="https://user-images.githubusercontent.com/15121999/31773097-f0979d82-b4e1-11e7-8b73-628ac34ad204.png">

Browser used:
- Google Chrome (Version 61.0.3163.100 (Official Build) (64-bit))
- Safari (Version 11.0 (12604.1.38.1.7))
- Mozilla Firefox (56.0.1 (64-bit))
- Opera (Version 48.0.2685.39)

ct-goranpalibrk:
I started working on this

lorthirk:
Fixed by c47024a70b51ec15976fd5a0f2c3dfe0f67fa169

Issue #896, opened 2017-10-20
User Filtering Box Reset Button Not Working Properly


LeoNerdoG:
When trying to use filtering box in Users tab, everything is OK until user presses "Reset" button. Then the  "Status" field is left blank (it should be automatically set to "ANY") so user has to manually set this field to "ANY", "ENABLED" or "DISABLED". 

Testflow: 
1) Login as kapua-sys
2) Go to Users tab and create a couple of users
3) Use Filtering box on the right side, after search click "Reset"
4) Status field is blank, but it should be set to "ANY"
**At this point user cannot filter the users, unless it manually sets the "Status" field. **

![screen shot 2017-10-20 at 14 06 04](https://user-images.githubusercontent.com/15121999/31820281-ecdcf01a-b5a0-11e7-8ece-4645e150928d.png)

Browsers used:
- Google Chrome (Version 61.0.3163.100 (Official Build) (64-bit))
- Safari (Version 11.0 (12604.1.38.1.7))
- Mozilla Firefox (56.0.1 (64-bit))
- Opera (Version 48.0.2685.39)

Issue #891, opened 2017-10-19
No Client IP Field in Connections


LeoNerdoG:
Connection filter contains Client ID, Connection Status and Client IP fields, which we can use to filter devices. 
When we look at the columns available in Connections, there is no Client IP field (this way it is much harder to filter devices by their IP). 

Testflow: 
1) Login as kapua-sys
2) Connect several devices to Kapua
3) Go to Connections - there is no Client IP column available. 

![screen shot 2017-10-19 at 14 17 23](https://user-images.githubusercontent.com/15121999/31772693-c4c300b2-b4e0-11e7-80c0-82dabf4fd892.png)

Browser used:
- Google Chrome (Version 61.0.3163.100 (Official Build) (64-bit))
- Safari (Version 11.0 (12604.1.38.1.7))
- Mozilla Firefox (56.0.1 (64-bit))
- Opera (Version 48.0.2685.39)

bojanutkovic:
I started working on this

Issue #889, opened 2017-10-19
Name "Kapua" is Missing in Info tab


LeoNerdoG:
When navigating to About tab, there is only logo of Kapua, but no name. 

<img width="1440" alt="screen shot 2017-10-19 at 14 15 05" src="https://user-images.githubusercontent.com/15121999/31770373-f0380b60-b4d7-11e7-80d8-070b6cd70fed.png">

Browser used:
- Google Chrome (Version 61.0.3163.100 (Official Build) (64-bit))
- Safari (Version 11.0 (12604.1.38.1.7))
- Mozilla Firefox (56.0.1 (64-bit))
- Opera (Version 48.0.2685.39)

PR #911, opened 2017-10-30
Console UI bugs

bojanutkovic:
Tasks: #891, #880, #883, #884,  #889, #896

### DIFF
--- a/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/ui/grid/KapuaGrid.java
+++ b/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/ui/grid/KapuaGrid.java
@@ -37,6 +37,6 @@ public class KapuaGrid<M extends KapuaBaseModel> extends Grid<M> {
         GridView gridView = getView();
         gridView.setAutoFill(true);
         gridView.setForceFit(true);
-        gridView.setSortingEnabled(false);
+        gridView.setSortingEnabled(true);
     }
 }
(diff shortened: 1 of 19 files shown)


lorthirk:
Just out of curiosity, How does this fixes #892? It should be fixed by c47024a70b51ec15976fd5a0f2c3dfe0f67fa169, does something in this PR also address it?

bojanutkovic:
Sorry I made a mistake, that is fixed in "Console review fixes part 2" PR.

lorthirk:
Ok, so I'm editing this one :smile:

console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/ui/grid/KapuaGrid.java
```diff
@@ -37,6 +37,6 @@ public KapuaGrid(ListStore<M> store, ColumnModel cm) {
         GridView gridView = getView();
         gridView.setAutoFill(true);
         gridView.setForceFit(true);
-        gridView.setSortingEnabled(false);
+        gridView.setSortingEnabled(true);
```
lorthirk: Ok for this line, but we probably also want `gridView.setRemoteSorting(true)` so that any sort triggers a new query to the service. Also be sure to test that the behavior is correct in all the grids after this.

lorthirk: Actually, I see that the backend services needs to be changed in order to support remote sorting. I'll do this ASAP.

Issue #885, opened 2017-10-19
Child Account Can Delete Itself


LeoNerdoG:
If a User changes from one child account to another, Kapua GUI is not automatically refreshed (but it should be). So if we are logged in as kapua-sys and we have Child Accounts tab open, we change to kapua-child1 (look at images bellow), we still see child accounts of kapua-sys. Because of this, we can delete kapua-child1 account, though we are signed in.
This is same for all child accounts, not only for kapua-sys <-> kapua-child1. 

Testflow: 
1) Sign in as kapua-sys
2) Go to Child Accounts tab and create new child account (e.g. kapua-child1)
3) Switch to kapua-child1 account.
4) Delete kapua-child1 account
After we delete the account, we can still click through its tabs (we can also add new Jobs?), but if we look at the child accounts available (last screenshot), there is no kapua-child1 account. 

<img width="1440" alt="screen shot 2017-10-19 at 08 33 36" src="https://user-images.githubusercontent.com/15121999/31757971-c38a6742-b4ab-11e7-9487-3a991e6cbbd6.png">

<img width="1440" alt="screen shot 2017-10-19 at 08 33 55" src="https://user-images.githubusercontent.com/15121999/31757976-c8cf2ada-b4ab-11e7-9d73-3648b58a2219.png">

<img width="1440" alt="screen shot 2017-10-19 at 09 00 51" src="https://user-images.githubusercontent.com/15121999/31758037-0fcc81a8-b4ac-11e7-95db-fd17fc92f1c2.png">

Browsers used:
- Google Chrome (Version 61.0.3163.100 (Official Build) (64-bit))
- Safari (Version 11.0 (12604.1.38.1.7))
- Mozilla Firefox (56.0.1 (64-bit))
- Opera (Version 48.0.2685.39)

lorthirk:
Could you please open a separate issue for "If a User changes from one child account to another, Kapua GUI is not automatically refreshed (but it should be)" as well?

lorthirk:
To be verified again after fixing #894

LeoNerdoG:
I have verified that this issue is resolved via PR #953 .

PR #953, opened 2017-11-10
Fix #894

### LINKED ISSUE #894: Kapua GUI Not Automatically Refreshed After Changing Accounts
This is related to Issue #885 .
If a user is logged in as kapua-sys, creates new child account and then switches to this newly created child account, Kapua GUI does not refresh automatically, but it should. 

Testflow: 
1) Login as kapua-sys
2) Go to Child Accounts, create new child accout (e.g. kapua-child1)
3) switch to kapua-child1 aaccount and observe the Kapua GUI
**We can see, that GUI does not refresh itself, as show on screenshots bellow.**

Screenshot 1: we are logged in as kapua-sys user

Screenshot 2: We are switching to kapua-child1 account

Screenshot 3: Kakpua GUI does not refresh

Browser Used: 

ct-goranpalibrk:


LeoNerdoG:
Hi @ct-goranpalibrk @lorthirk ,
I am verifying that issues #885 and #894 are solved with this PR.

PR #971, opened 2017-11-14
Fix #862

### LINKED ISSUE #862: rxByteLimit in MessageStoreService Can be Set to Negative Value
When setting child account's rxByteLimit (Total data usage per month),negative number can be entered without any error. 

TestFlow:
1) Log in as kapua-sys
2) Create new child account (e.g. kapua-child1)
3) go to child account's Settings -> MessageStoreService
4) Set rxByteLimit to -10 
**Kapua returns no error (but it should, as data usage cannot be negative).**

Browsers used:

ct-goranpalibrk:


LeoNerdoG:
I confirmed that this PR solves issue #862 so it can be merged.

PR #991, opened 2017-11-16
QA steps moved to main from src - reuse of Cucumber steps

muros-ct:
Cucumber steps for services are moved from test to main, making them
accessible in other modules for creating Cucumber tests.

org.eclipse.kapua:kapua-qa is maven dependency that includes this steps.

Refactored for usage with sql scripts that are part of QA module resources
in maven artifacts and not external files.


muros-ct:
This PR is replaced by #1047 which moves steps into seaprate module kapua-qa-steps.

PR #1047, opened 2017-11-22
QA steps moved to separate module qa-steps

muros-ct:
Cucumber steps for services are moved from test to main, making them
accessible in other modules for creating Cucumber tests. kapua-qa
uses it as a dependency.

org.eclipse.kapua:kapua-qa-steps is maven dependency that includes this steps.

Refactored for usage with sql scripts that are part of QA module resources
in maven artifacts and not external files.


muros-ct:
@Coduz This PR replaces PR #991.
Change is only that it is separate module not inside qa. qa than uses this module.
When we merge this I will remove #991

PR #1065, opened 2017-11-24
Fixed console error report messages 

### LINKED ISSUE #1057: Nonconsistent Error When Trying To Delete a Device
Best way to explain this issue is through an example. 
Basically we need a user that has all the _**read**_ and _**write**_ permissions but no _**delete**_ permissions.
After we login, go to Tags (or Jobs, Groups...it does not really matter) and create a tag (e.g. tag1). After that try to delete tag1. Kapua will return an error: **Subject does not have permission _tag:delete_**, which is ok. 
Now go to Devices tab and create a device (e.g. device1) and after that try to delete it. Kapua will return partial error of previous one, when we tried to delete a tag (screenshot No. 1).

Now repeat upper procedure with job. Go to Jobs and create a job (e.g. job1). Kapua will return information: **Job successfully created**, which is, again, ok. 
Now go to devices and try to again delete device1. This time Kapua will return partial _errror_: Job successfully created (screenshot No. 2).

This happens with all other scenarios: when trying to delete device without **_device:delete_** permissions, Kapua shows wrong error. 

Testflow: 
1. Login as kapua-sys
2. Go to Users, create user (e.g. kapua-user1), 
3. Add read and write, but no delete permissions for all domains
4. Login as kapua-user1 and follow procedure described in example.

Screenshot No. 1
screen shot 2017-11-23 at 14 03 30

Screenshot No. 2
screen shot 2017-11-23 at 14 06 08

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.39)

bojanutkovic:


console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/DeviceDeleteDialog.java
```diff
@@ -58,9 +59,9 @@ public void submit() {
                 new AsyncCallback<Void>() {
 
                     public void onFailure(Throwable caught) {
+                        FailureHandler.handle(caught);
```
Coduz: Handling of the error is managed by the `EntityCRUDToolbar` so is enough to set `exitStatus` and `exitMessage`. 

This call to the `FailureHandler` can be removed

Coduz:
Change the title to a more meaning full one.

This PR fixes #1057

LeoNerdoG:
@bojanutkovic @Coduz ,
I have verified that this PR resolves issue #1057 .

PR #1166, opened 2017-12-11
Basic tests for KapuaDateUtils ISO-8601 support

muros-ct:
This is just porposal how to test ISO compatibility for week of year.

### DIFF
--- a/commons/src/test/resources/kapua-environment-setting.properties
+++ b/commons/src/test/resources/kapua-environment-setting.properties
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2011, 2016 Eurotech and/or its affiliates and others
+# Copyright (c) 2011, 2018 Eurotech and/or its affiliates and others
 #
 # All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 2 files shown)


commons/src/test/resources/kapua-environment-setting.properties
```diff
@@ -66,6 +66,7 @@ commons.locale.zoneId=UTC
 #Locale ID (Default en_US). See {@link DayOfWeek} for more information about allowed settings for this parameter
 commons.locale.id=en_US
 #Local minimal days in first week (Deafault 7). See {@link WeekFields} for more information about allowed settings for this parameter
```
riccardomodanese: We can set this parameter to 4 for now but we should discuss in one of the next scrum about which should be the more appropriate value since, setting to 7 days, all the weeks in the year should have the same length so, ideally, an equally distributed amount of messages in each data store index. 
Can you update also the comment (Default 7 -> Default 4)

muros-ct:
@riccardomodanese what shall we do with this PR. Shall we merge it or we don't need this test and can simply remove it?

Issue #1080, opened 2017-11-28
Multiple Same Name Schedules Can Be Added In Jobs


LeoNerdoG:
When adding schedules for certain job, schedules with same names can be added (look at the screenshot). 

Testflow:
1. Login as kapua-sys
2. Go to jobs, create a job (e.g. job1) 
3. Select job1 and add schedule for it (e.g. schedule1) with all the necessary parameters.
4. Try to create schedule with same name again. 
**At this point Kapua should report an error because there are two (or more) schedules with same name.**

<img width="1438" alt="screen shot 2017-11-28 at 15 32 02" src="https://user-images.githubusercontent.com/15121999/33325159-d2c016ba-d451-11e7-9fe5-d05ac56ed75a.png">

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.39)

LeoNerdoG:
I have verified that this issue is solved by PR #1170 .

PR #1170, opened 2017-12-11
fixed adding two schedules with same name in Jobs

ct-goranpalibrk:

### DIFF
--- a/service/scheduler/api/src/main/java/org/eclipse/kapua/service/scheduler/trigger/TriggerService.java
+++ b/service/scheduler/api/src/main/java/org/eclipse/kapua/service/scheduler/trigger/TriggerService.java
@@ -40,4 +40,6 @@ public interface TriggerService extends KapuaEntityService<Trigger, TriggerCreat
      */
     public TriggerListResult query(KapuaQuery<Trigger> query)
             throws KapuaException;
+
+    public void setQuery(KapuaQuery<Trigger> query) throws KapuaException;
 }
--- a/service/scheduler/api/src/main/java/org/eclipse/kapua/service/scheduler/trigger/TriggerService.java
+++ b/service/scheduler/api/src/main/java/org/eclipse/kapua/service/scheduler/trigger/TriggerService.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017 Eurotech and/or its affiliates and others
+ * Copyright (c) 2017, 2018 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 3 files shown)


LeoNerdoG:
@ct-goranpalibrk @lorthirk 
I have verified, that this PR fixes issue #1080 .

LeoNerdoG:
Hi @ct-goranpalibrk @lorthirk 
I have verified that this PR fixes issue #1080 .

service/scheduler/quartz/src/main/java/org/eclipse/kapua/service/scheduler/trigger/quartz/TriggerServiceImpl.java
```diff
@@ -57,6 +59,7 @@
         implements TriggerService {
 
     private static final Domain SCHEDULER_DOMAIN = new SchedulerDomain();
+    KapuaQuery<Trigger> queryByJobId;
```
Coduz: This MUST be removed. 
Service classes are a singleton. 
Relying on the value set here is not suitable since this a value shared across the whole JVM.

service/scheduler/quartz/src/main/java/org/eclipse/kapua/service/scheduler/trigger/quartz/TriggerServiceImpl.java
```diff
@@ -87,7 +90,14 @@ public Trigger create(TriggerCreator triggerCreator)
         authorizationService.checkPermission(permissionFactory.newPermission(SCHEDULER_DOMAIN, Actions.write, triggerCreator.getScopeId()));
 
         return entityManagerSession.onTransactedInsert(em -> {
-            // Kapua Trigger definition
+
+            KapuaQuery<Trigger> triggerQuery = queryByJobId;
+            TriggerListResult list = query(triggerQuery);
+            for (Trigger trigger : list.getItems()) {
```
Coduz: This is a very inefficient way to do it. 
If someone has 5000 of triggers you'll need to scan them all, which is far more inefficient than just query by name.

service/scheduler/api/src/main/java/org/eclipse/kapua/service/scheduler/trigger/TriggerService.java
```diff
@@ -40,4 +40,6 @@
      */
     public TriggerListResult query(KapuaQuery<Trigger> query)
             throws KapuaException;
+
+    public void setQuery(KapuaQuery<Trigger> query) throws KapuaException;
```
Coduz: This method MUST be removed from the API. 
Service APIs MUST remain clean, avoiding adding 'in-utility' methods.

ct-goranpalibrk:
I have implemented new changes,  methods that you told me to delete are deleted, and all logic in create method in TriggerServiceImpl is done with predicates.

lorthirk:
@ct-goranpalibrk , in your last commit it looks like you're indenting code with two spaces instead of four, like the rest of the Kapua code base. Could you please check your formatter options?

LeoNerdoG:
Hi @ct-goranpalibrk @lorthirk ,
I have reviewed the fix once more and I can confirm, it successfully fixes issue #1080 .
It does not affect any other thing so from my perspective it can be merged.

Issue #1011, opened 2017-11-21
Kapua UI Freezes If Custom Attributes Too Long


LeoNerdoG:
If string entered in any of the Custom Attributes in devices by user is longer than 255 characters, Kapua freezes and needs to be refreshed. 

Testflow: 
1. Login as kapua-sys
2. Go to Devices, add/create a device (e.g. device1) 
3. Select device1, click edit and add string longer than 255 characters in any of Custom Attributes. 
**Kapua freezes and has to be refreshed for normal operation.**

<img width="722" alt="screen shot 2017-11-21 at 12 51 16" src="https://user-images.githubusercontent.com/15121999/33071506-0b40f630-cebc-11e7-9423-94c8dd7f0036.png">

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.34)

LeoNerdoG:
@lorthirk I have verified that this issue is resolved.

Issue #1012, opened 2017-11-21
Kapua UI Freezes If Display Name Too Long


LeoNerdoG:
If string entered in Display Name in devices by user is longer than 255 characters, Kapua freezes and needs to be refreshed.

Testflow:
1. Login as kapua-sys
2. Go to Devices, add/create a device (e.g. device1)
3. Select device1, click edit and add Display Name longer than 255 characters.
**Kapua freezes and has to be refreshed for normal operation.**

<img width="1440" alt="screen shot 2017-11-21 at 13 08 33" src="https://user-images.githubusercontent.com/15121999/33071806-4b1cbb76-cebd-11e7-87cd-b2471761e20d.png">

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.34)

LeoNerdoG:
@lorthirk I have verified that this issue is resolved.

PR #1178, opened 2017-12-11
Fix device edit dialog not freezes when custom attribute is too long

ct-goranpalibrk:
Issue #1011


ct-goranpalibrk:
This PR also resolves issue  #1012

LeoNerdoG:
@ct-goranpalibrk @lorthirk 
I have verified that this PR resolves issues #1011 and #1012 .

PR #1197, opened 2017-12-13
Resource file separator

andrejn-ct:
## Resource file separator
KapuaConfigurableServiceSchemaUtilsWithResources uses a system dependent separator (File.separator) in the resource name instead of the '/'-separated path name the Java docs define ([https://docs.oracle.com/javase/7/docs/api/java/lang/ClassLoader.html#getResource(java.lang.String)]). 
On some systems (Windows) this returns the wrong separator ('\\'), which causes test (and possibly runtime) failures.

No changes.

### Implementation changes
The implementation of KapuaConfigurableServiceSchemaUtilsWithResources was modified to use the correct separator.

### Test changes
No changes.


muros-ct:
I run branch on local - Mac OS based machine and it works.
All tests pass, and this change only influence tests.

Issue #1105, opened 2017-12-01
Non-specific Error When Creating Tag With Non-unique Name


LeoNerdoG:
Kapua returns a non-specific (_**Persistance error**_) error/info message, when user tries to create two tags with same names. 

Testflow:
1. Login as kapua-sys
2. Go to Tags, try to create two tags with same names. 
**Kapua returns an _persistance_ error with non-specific info message (look at the screenshot). Message should imply, what the problem is.**

<img width="1434" alt="screen shot 2017-12-01 at 15 36 44" src="https://user-images.githubusercontent.com/15121999/33487459-92b95c30-d6ad-11e7-96bc-601901feb2b5.png">

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.39)

Issue #1104, opened 2017-12-01
Non-specific Error When Creating Roles With Non-unique Name


LeoNerdoG:
Kapua returns a non-specific (_**Persistance error**_) error/info message, when user tries to create two roles with same names. 

Testflow:
1. Login as kapua-sys
2. Go to Roles, try to create two roles with same names. 
**Kapua returns an _persistance_ error with non-specific info message (look at the screenshot). Message should imply, what the problem is.**

<img width="1438" alt="screen shot 2017-12-01 at 15 34 14" src="https://user-images.githubusercontent.com/15121999/33487256-dc7239f6-d6ac-11e7-89ee-21257ad54c9d.png">

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.39)

Issue #1103, opened 2017-12-01
Non-specific Error When Creating Group With Non-unique Name


LeoNerdoG:
Kapua returns a non-specific (_**Persistance error**_) error/info message, when user tries to create two groups with same names. 

Testflow:
1. Login as kapua-sys
2. Go to Groups, try to create two groups with same names. 
**Kapua returns an _persistance_ error with non-specific info message (look at the screenshot). Message should imply, what the problem is.**

<img width="1439" alt="screen shot 2017-12-01 at 15 31 54" src="https://user-images.githubusercontent.com/15121999/33487149-9c029794-d6ac-11e7-9798-1cab6d34557c.png">

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.39)

LeoNerdoG:
I have verified that this issue is resolved.

PR #1216, opened 2017-12-19
Non-specific Error fixes

### LINKED ISSUE #1102: Non-specific Error When Creating User With Non-unique Name
Best way to demonstrate this error is through an example. 
If end user is logged in as kapua-sys, creates a user (e.g. kapua-user1) and then tries to create a user with the same name in a child account (e.g. kapua-child1), Kapua returns an _**Persistance error**_ with non-specific info message, which can be confusing for end user, especially if there are a lot of users, child accounts,...

Also this error appears if user tries to create two uses with same name within same user account (e.g. two same users in kapua-sys).

Testflow:
1. Login as kapua-sys
2. Go to Users, create a user (e.g. kapua-user1)
3. Go to Child accounts and create a child account (e.g. kapua-child1)
4. Switch to kapua-child1 and go to Users; try to create user with name kapua-user1. 
**Kapua returns an _persistance_ error with non-specific info message (look at the screenshot). Message should imply, what the problem is.**

screen shot 2017-12-01 at 15 14 08

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.39)

bojanutkovic:
#1102, #1103, #1104, #1105, #1106, #1107


LeoNerdoG:
@bojanutkovic @lorthirk 
I have verified that this PR fixes all of the issues listed above.

Issue #1231, opened 2017-12-24
jobs - test scenarios fail creating job items


stefanomorson:
Testflow:
Under org.eclipse.kapua.service.job, run 

> RunTest

in debug mode

Two scenarios are failing:
* Count job items
* Query for job items

muros-ct:
Solved by (re)moving failing unit/crud test.
This test shall be implemented as part of integration QA test by @andrejn-ct.

PR #1245, opened 2018-01-03
Removed a misbehaving test scenario from the Job service unit test suite

andrejn-ct:
## Job service unit tests
A misbehaving unit test was removed. Eventually it will be moved into the Job service integration tests. At the moment is is just commented out to allow the test suite to pass.

No changes.

### Implementation changes
No changes.

### Test changes
A Cucumner test scenario was removed.


muros-ct:
This fixes #1231

PR #1279, opened 2018-01-11
Allow code coverage drop on PR 

Coduz:
This is a tentative change to allow a minimum code coverage drop in PRs.
This is intended to 'solve' a lot of cases where a code change which causes a small drop (like -0.02%) and this is resulting in a PR check failure.


Coduz:
This PR is subsituted by #1282

PR #1321, opened 2018-01-24
connector descriptor fix pluggability

riccardomodanese:
Allow the customization of the message translator chain also from outside Kapua codebase.
So an extension module can plug its own messages types and then the translators.


riccardomodanese:
Old feature no more needed!

Issue #1083, opened 2017-11-29
Very Long Client-ID Name In Data Not Shown Properly


LeoNerdoG:
When Kapua contains a device with very long device-id name (look at the screenshot) its name is not entirely shown  (nor in the _**Available Topics**_ nor in _**Available Metrics For Topic**_). 

Testflow: 
1. Login as kapua-sys
2. Connect a live device to Kapua (e.g. Raspberry Pi)
3. Change Client-ID on the connected device to a very long name.
**Kapua does not show entire name and user cannot do anything to properly show it.**

<img width="1440" alt="screen shot 2017-11-29 at 10 10 38" src="https://user-images.githubusercontent.com/15121999/33369352-8a4d5804-d4f4-11e7-8ae7-e16f6d9fd1b9.png">

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.39)

muros-ct:
This is how very long client-id changes metric display. It is split to multiple lines.
@LeoNerdoG  Is this OK?
If yes, this is allready implemented and can be closed.

![very-long](https://user-images.githubusercontent.com/23033235/35560649-416a578c-05ae-11e8-85ba-36af668039ed.png)

@lorthirk This is solution that we discussed. It seams it works allready.

LeoNerdoG:
@muros-ct @lorthirk I have verified that this is indeed solved, so I am closing.

lorthirk:
Uhm... honestly I would reopen it. It would be better to truncate at the first line instead of wrapping to the new one.

LeoNerdoG:
@muros-ct @lorthirk I have verified that this issue is solved by PR #1337 .

PR #1337, opened 2018-01-30
Fixing long client id name in header with ...

muros-ct:
This fixes issue #1083 where name in header of topic table is truncated to
width of header and in case of long name ended with ...


muros-ct:
@LeoNerdoG please evaluate

LeoNerdoG:
@muros-ct I have verified that this PR solves issue #1083 .

Issue #1315, opened 2018-01-19
Tab Not Highlighted When User Switches Accounts


LeoNerdoG:
If user uses one account and then switches to another (e.g. from kapua-sys to child-account1), the tab, which was active last in kapua-sys (e.g. Devices), is not higlighted, when user changes accounts to child-account1 (although nothing else changed). 

Testflow: 
1. Login as kapua-sys
2. Go to Child accounts, create new account (e.g. child-account1) 
3. Go to Devices and then change to child-account1 account
**At this point Kapua will switch accounts but the _Device_ tab will no longer be highlighted.**

Screeshot No. 1 (User in kapua-sys account, switching to child-account1): 
![screen shot 2018-01-19 at 12 26 43](https://user-images.githubusercontent.com/15121999/35149398-36aca192-fd16-11e7-96ff-5df84d141c84.png)

Screenshot No. 2 (User in child-account1, Devices tab not highlighted):
![screen shot 2018-01-19 at 12 26 53](https://user-images.githubusercontent.com/15121999/35149396-35947bae-fd16-11e7-9c54-532bc409f88d.png)

Browsers used:
- Google Chrome (Version 63.0.3239.132 (Official Build) (64-bit)))
- Safari (Version 11.0.2 (13604.3.5))
- Mozilla Firefox (57.0.4 (64-bit))
- Opera (Version 50.0.2762.45)

stefanomorson:
#### Expected result:
The menu item on the left should be selected

LeoNerdoG:
I have verified that this issue is resolved by PR #1341 .

PR #1341, opened 2018-01-31
fix current view is not highlighted when user is changed

ct-goranpalibrk:


ct-goranpalibrk:
PR for #1315

ct-goranpalibrk:
Header in North View changed.

Issue #1266, opened 2018-01-09
Dialogue Box Closes if User Enters Job With Non-Unique Name


LeoNerdoG:
If user enters non-unique name in Jobs, Kapua does not close the dialogue box, but only shows an error that user has not used a non-unique name. 
But if user enters for example a tag with non-unique name, after clicking **Submit** the dialogue box closes and an error is shown (which is ok).
Problem here is non consistency and for end user is much better, if dialogue box does not close. 

Testflow: 
1. Login as kapua-sys
2. Go to Jobs, add a job (e.g. job1) 
3. Try to add another job, also with name job1
**At this point Kapua will close the dialogue box and show error, instead just showing the error and keeping the dialogue box open.**

Screenshot No. 1 (No dialogue box when job with non-unique name entered):
<img width="1435" alt="screen shot 2018-01-09 at 11 21 27" src="https://user-images.githubusercontent.com/15121999/34721132-dc63de84-f541-11e7-8a9a-fc418cf4559c.png">

Screenshot No. 2 (Dialogue box does not close when tag with non-unique name entered): 
<img width="721" alt="screen shot 2018-01-09 at 11 21 45" src="https://user-images.githubusercontent.com/15121999/34720944-081d6abe-f541-11e7-9090-6e233838b999.png">

Browsers used:
- Google Chrome (Version 63.0.3239.132 (Official Build) (64-bit)))
- Safari (Version 11.0.2 (13604.3.5))
- Mozilla Firefox (57.0.4 (64-bit))
- Opera (Version 50.0.2762.45)

stefanomorson:
#### Expected results:
Same as #1264

LeoNerdoG:
I have verified that this issue has been successfully resolved with PR #1350 .

Issue #1265, opened 2018-01-09
Dialogue Box Closes if User Enters Device With Non-Unique Name


LeoNerdoG:
If user enters non-unique name in Devices, Kapua does not close the dialogue box, but only shows an error that user has not used a non-unique name. 
But if user enters for example a tag with non-unique name, after clicking **Submit** the dialogue box closes and an error is shown (which is ok).
Problem here is non consistency and for end user is much better, if dialogue box does not close. 

Testflow: 
1. Login as kapua-sys
2. Go to Devices, add a device (e.g. device1) 
3. Try to add another device, also with name device1
**At this point Kapua will close the dialogue box and show error, instead just showing the error and keeping the dialogue box open.**

Screenshot No. 1 (No dialogue box when device with non-unique name entered):
<img width="1434" alt="screen shot 2018-01-09 at 11 21 15" src="https://user-images.githubusercontent.com/15121999/34720940-04f69e32-f541-11e7-8a5b-242cae4000df.png">

Screenshot No. 2 (Dialogue box does not close when tag with non-unique name entered): 
<img width="721" alt="screen shot 2018-01-09 at 11 21 45" src="https://user-images.githubusercontent.com/15121999/34720944-081d6abe-f541-11e7-9090-6e233838b999.png">

Browsers used:
- Google Chrome (Version 63.0.3239.132 (Official Build) (64-bit)))
- Safari (Version 11.0.2 (13604.3.5))
- Mozilla Firefox (57.0.4 (64-bit))
- Opera (Version 50.0.2762.45)

stefanomorson:
@LeoNerdoG : maybe there's an inconsistency in the description, did you mean that the device dialog closes while the tag dialog keeps open ?
In any case the issue looks similar to #1264

LeoNerdoG:
Hi @stefanomorson ,
yes, exactly. The tag dialogue box is kept open, if user enters non-unique name and error pops-out. This way user does not have to click "add" again and enter all the info, but only changes the tag name and that's it. 
But the Device dialogue box closes, which is very inconvenient (as it has a lot more fields to fill).

LeoNerdoG:
I have verified that this issue has been successfully resolved with PR #1350 .

PR #1350, opened 2018-02-01
Dialog box fixes

### LINKED ISSUE #1264: Dialogue Box Closes if User With Non-Unique Name Entered
If user enters non-unique name in Users, Kapua does not close the dialogue box, but only shows an error that user has not used a non-unique name. 
But if user enters for example a tag with non-unique name, after clicking **Submit** the dialogue box closes and an error is shown (which is ok).
Problem here is non consistency and for end user is much better, if dialogue box does not close. 

Testflow: 
1. Login as kapua-sys
2. Go to Users, add a job (e.g. user1) 
3. Try to add another user, also with name user1
**At this point Kapua will close the dialogue box and show error, instead just showing the error and keeping the dialogue box open.**

Screenshot No. 1 (No dialogue box when job with non-unique name entered):
screen shot 2018-01-09 at 11 21 36

Screenshot No. 2 (Dialogue box does not close when tag with non-unique name entered): 
screen shot 2018-01-09 at 11 21 45

Browsers used:
- Google Chrome (Version 63.0.3239.132 (Official Build) (64-bit)))
- Safari (Version 11.0.2 (13604.3.5))
- Mozilla Firefox (57.0.4 (64-bit))
- Opera (Version 50.0.2762.45)

ct-ajovanovic:
#1264, #1265, #1266

### DIFF
--- a/console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/DeviceAddDialog.java
+++ b/console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/DeviceAddDialog.java
@@ -275,18 +275,12 @@ public class DeviceAddDialog extends EntityAddEditDialog {
 
             @Override
             public void onFailure(Throwable caught) {
+                FailureHandler.handleFormException(formPanel, caught);
+                status.hide();
+                formPanel.getButtonBar().enable();
                 unmask();
-
                 submitButton.enable();
                 cancelButton.enable();
-                status.hide();
-
-                exitStatus = false;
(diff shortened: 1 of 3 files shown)


LeoNerdoG:
I have verified that this PR successfuelly resolves issues #1264 , #1265 and #1266 .

muros-ct:
@ct-ajovanovic and @bojanutkovic  
Two things:
- `exitStatus = false;` is not set in onFailure header, is there reason not to set it or should also be removed from onSuccess.
- Please update header, there should be
`Copyright (c) 2017, 2018 Eurotech and/or its affiliates and others`

Issue #1206, opened 2017-12-15
Wrong Message When Adding Empty Target List in Jobs 


LeoNerdoG:
When user wants to add a target for a certain job, but selects none of them, Kapua returns _**Targets successfully added**_, although non have been added, because user has not selected any. 

Testflow: 
1. Login as kapua-sys
2. Go to Jobs, add a job. 
3. Try to add _none_ of the devices to the target list
**Although user has not added any devices as targets, Kapua returns info as described above and seen in bellow screenshots.**

Screenshot No. 1 (User selects none of the devices and clicks _Submit_):
<img width="1436" alt="screen shot 2017-12-15 at 10 28 45" src="https://user-images.githubusercontent.com/15121999/34036442-7e53a844-e185-11e7-8474-9b76834af81d.png">

Screenshot No. 2 (Kapua says that devices have been added successfully, although none have  been selected): 
<img width="1437" alt="screen shot 2017-12-15 at 10 28 51" src="https://user-images.githubusercontent.com/15121999/34036447-80b0278e-e185-11e7-86de-08f50e60669d.png">

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0.1 (64-bit))
- Opera (Version 49.0.2725.39)

stefanomorson:
SImilar to #1172

LeoNerdoG:
I have verified that thiss issue has been resolved with PR #1355 .

Issue #1172, opened 2017-12-11
Wrong Message When Adding Same Targets in Jobs


LeoNerdoG:
When user tries to add same devices (targets) twice to a job, Kapua always reports _**Targets Successfully Added**_, altohugh it should report that certain devices have already been added. 

Testflow: 
1. Login as kapua-sys
2. Go to Devices, add two devices (e.g. device1 and device2)
3. Go to Jobs, create a job (e.g. job1) and try to add both devices (targets) to this job. 
**Try to add same devices multiple times. Kapua will always return info: _Targetes Successfully Added_.**

<img width="1439" alt="screen shot 2017-12-11 at 13 31 26" src="https://user-images.githubusercontent.com/15121999/33831330-e995a9e2-de77-11e7-8024-5d2b77cc9adf.png">

Browsers used
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0.1 (64-bit))
- Opera (Version 49.0.2725.39)

stefanomorson:
By design, the functionality adds a new target in the list only if the target it is not already present. The same can be said if a list of targets are selected to be added.

#### Expected results:
If the list of newly created targets is empty the notification message should report that "No new targets have been selected"

LeoNerdoG:
I have verified that this issue is resolved in PR #1355 .

PR #1355, opened 2018-02-02
Fixing Wrong messages when adding targets to Job

muros-ct:
Messages were fixed and new logic added to check if targets are
actually being added.

Fixes issues:
- 1206
- 1172


LeoNerdoG:
I have verified that this PR solves issues #1206 and #1172 .

muros-ct:
@bojanutkovic and @ct-goranpalibrk, can you check this code.

PR #1358, opened 2018-02-05
Feature endpoint info service

### LINKED ISSUE #1342: Implementation of EndpointInfoService
Implementation of `EndpointInfoService` which will be used to store scope-specific endpoints for accounts.

This will replace the current Configurable service option `brokerUri` of the `AccountService` and will allow to define multiple endpoints with description options for each of them.

This service will be managed only by the kapua admin account and will be accessible in read mode to all other accounts.

Coduz:
This issue closes #1342 .

This service has been also integrated with the rest of the component.

Only the root account can manage EndpointInfo entities while all other accounts get the EndpointInfo from their self or the first ancestor that has one, scaling up to the default platform value returned by invoking `SystemUtils.getNodeURI()`.

### DIFF
--- a/console/module/account/pom.xml
+++ b/console/module/account/pom.xml
@@ -11,7 +11,7 @@
         Eurotech - initial API and implementation
 -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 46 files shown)


Coduz:
Failures of Travis CI comes from develop failures which are being fixed by #1379

Issue #1306, opened 2018-01-18
Sort Icon in "Created On" Column in Devices Snapshots is not working


LeoNerdoG:
In _**Created On**_ column there is a triangle icon that is ussually represents button for setting sorting (ascending/descending) but in this case it is unclickable. 

Testflow: 
1. Login as kapua-sys
2. Connect live Kura device to Kapua
3. Go to Devices, select the live device and navigate to Configuration -> Snapshots -> Created On
**Screenshot bellow shows where the unclickable triangle icon is located.**

![triangleinsnapshots](https://user-images.githubusercontent.com/15121999/35089534-84acbe02-fc37-11e7-9fbc-1f17e5ab3ccf.png)

Browsers used:
- Google Chrome (Version 63.0.3239.132 (Official Build) (64-bit)))
- Safari (Version 11.0.2 (13604.3.5))
- Mozilla Firefox (57.0.4 (64-bit))
- Opera (Version 50.0.2762.45)

stefanomorson:
#### Expected results:
by clicking on the sort icon the entries in the list below should be sorted accordingly.

LeoNerdoG:
I have verified that this issue is resolved.

PR #1371, opened 2018-02-06
Set sortable columns in snapshot tabs in devices

ct-goranpalibrk:


LeoNerdoG:
I have verified that this PR successfully solves issue #1306 .

Issue #1191, opened 2017-12-13
Long Username Not Displayed Properly


LeoNerdoG:
If end-user creates kapua user with very long name (256 characters), its name is not properly displayed when loged in as this user (look at the screenshot). 

Testflow: 
1. Login as kapua-sys
2. Go to Users, create new user with very long name (256 characters - look at screenshot) OR
2.1 Create a user with normal, short name (e.g. kapua-user12) and a very long _**Display name**_
3. Logout as kapua-sys and login as user just created
**Look at the screenshot: If username is too long, it is not displayed properly.**

<img width="1437" alt="screen shot 2017-12-13 at 10 14 02" src="https://user-images.githubusercontent.com/15121999/33931970-f2f89c3a-dff1-11e7-998b-fd1da5be77ee.png">

Browsers used
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0.1 (64-bit))
- Opera (Version 49.0.2725.39)

stefanomorson:
#### Expected results:
* When the string <username>@<account-name> exceeds the width of the user dropdown menu the label inside should be truncated and ellipses should be added at the end of the string
* When the mouse is over the label a tip should appear with the full name (<username>@<account-name>)

LeoNerdoG:
I have verified that this issue has been resolved.

lorthirk:
Unfortunately I had to revert #1382 with #1521 because with that implementation the account name would be partially hidden most of the time if it's more than a few characters long, which is a very common scenario, for sure more than the one described in this issue.

LeoNerdoG:
I am reopening as PR #1382 has been reverted.

ct-anaalbic:
Hi @lorthirk 

I have one idea for this issue. Maybe I can add some menu item if username is too long, like on screenshot bellow:

![image](https://user-images.githubusercontent.com/45655642/58020218-cf096e00-7b07-11e9-8121-d879af400aad.png)

When user clicks on this item, he can get dialog with whole username, and also username in label should be truncated and ellipses should be added at the end of the string. What do you think about this solution?

Thank you.

Regards,
Ana

lorthirk:
Honestly... I don't thin I like this. It looks... counterintuitive, I think. @LeoNerdoG what do you think?

LeoNerdoG:
Hej @lorthirk,
[...]

PR #1382, opened 2018-02-09
Fix for display of long user name

muros-ct:
This solves issue wher use name is too long and is not correctyl displayed
in console header.
Name is shortende and ellipsis is added.
When mouse is hoovered over username its full name is displayed in tooltip.

Solution was first tried by setting style on button, but it didn't work, so
I had to resolve to changing console.css for button.

This solves issue 1191.


muros-ct:
@lorthirk Please check if this implemetation is ok, I am not satesfied with console.css change, but I couldn't make it work any other setStyle way.

LeoNerdoG:
I have verified that this PR successfully fixes issue #1191 .

LeoNerdoG:
I have again verified that this PR fixes issue #1191 .

PR #1379, opened 2018-02-08
Fixed GroupService tests

Coduz:
This PR fixes GroupService test which are now failing after 8df63c7372b57b297b347659e16339024167eec1

PR #1389, opened 2018-02-12
Fix service duplicate name check

Coduz:
This PR fixes the strategy of checking done to avoid duplicate names on entities. 
The check made before on the `.update(Entity)` methods was querying the DB filtering only by the name of the entity. This made impossible to update an entity without changing its name, because on update with `name` unchanged, the original record was matched and therefore causing an error. 
Also updated to use `.count(...)` instead of `.query(...)` which is more appropriate for the check, since we are not interested in the content of the matched entities.

Some refactoring of the services has been done on the Services involved in these changes. 

Improved the implementation of the `AndPredicate` implementation (
604c2bf) to allow ease of use when adding multiple clauses.

This PR also replaces #1379 which was the initial work of fixing the test `GroupService.feature` which then became a much bigger work

### DIFF
--- a/commons/src/main/java/org/eclipse/kapua/commons/model/query/predicate/AndPredicate.java
+++ b/commons/src/main/java/org/eclipse/kapua/commons/model/query/predicate/AndPredicate.java
@@ -11,17 +11,19 @@
  *******************************************************************************/
 package org.eclipse.kapua.commons.model.query.predicate;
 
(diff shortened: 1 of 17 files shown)


Coduz:
This last rebase should be final one. Fixed some predicates that had the wrong `Operator` set.
After the build succeeds this work can be merged

Issue #1134, opened 2017-12-06
"Upload and Apply" Button Not Frozen Although Device Disconnected


LeoNerdoG:
When device is offline (look at the screenshot), _**Upload and apply**_ button can still be pressed, user can also select a file and _upload_ it (Kapua returns an error). 

Testflow: 
1. Login as kapua-sys
2. Go to Devices and add a device (e.g. device1) 
3. Click on device1, go to _Snapshots_ tab under _Configuration_ and click _Upload and apply_.
4. Select a file and try to upload it. 
**Kapua will return an error, which is ok, but uploading a file to a offline device should not be possible at all (button should be frozen).**

![screen shot 2017-12-06 at 11 11 31](https://user-images.githubusercontent.com/15121999/33657938-267349c4-da7b-11e7-877e-4d77148a1908.png)

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.39)

LeoNerdoG:
I have verified that this issue is resolved.

PR #1401, opened 2018-02-14
upload and apply button frozen if device is online 

ct-goranpalibrk:


ct-goranpalibrk:
If device is null or ofline upload and apply button now is frozen in Snapshots tab in Devices view.

LeoNerdoG:
I have verified tha this PR successfully fixes issue #1134 .

muros-ct:
Simple one liner solution.
@LeoNerdoG if this solves issue, we can merge.
Thou it might still be a problem in case when device is disconnected and shown as connected in device grid. Is this solved by previous fixes? Please re-test.

LeoNerdoG:
@muros-ct I have retested this and found out that this PR does not solve problem in this specific case (when device is disconnected and shown as connected in device grid). Although this was not one of the scenarios of issue #1134 I think it should also be corrected before merging.

LeoNerdoG:
@muros-ct Goran and I have agreed that we will solve problem, you mentioned, later with a PR, which will solve not only issue with this button but with all others also - becuase problem you described is not directly connected to this button but is more general in its nature.

PR #1407, opened 2018-02-14
Fixed AuthenticationServiceShiroImpl.verifyCredential method

### LINKED ISSUE #1397: AuthenticationService.verifyCredentials(...) invocation causes problem with the session
Invocation of the `AuthenticationService.verifyCredentials(...)` method causes the current session to be messed up. This is likely that the methods calls a regular login and then the user that has invoked has its own session replaced by the one just created by the `verifyCredentials` login.

Coduz:
This PR fixes #1397 .

The Shiro `.login(..)` was invoked from the current `Subject`. 
This was causing the current session to be replaced with the one created from the `.login(...)` with the credentials to verify. 
I've changed to use a `Subject` on the fly, on which call `.login(...)` and then logout to avoid leave opened sessions.

I've also changed the `AuthenticationService.verify(...)` signature because it was set to always return `true` if no exception was risen. It makes more sense to have a `void` return value and throw an exception if validation fails.

### DIFF
--- a/console/module/authentication/src/main/java/org/eclipse/kapua/app/console/module/authentication/server/GwtCredentialServiceImpl.java
+++ b/console/module/authentication/src/main/java/org/eclipse/kapua/app/console/module/authentication/server/GwtCredentialServiceImpl.java
@@ -11,17 +11,13 @@
  *******************************************************************************/
 package org.eclipse.kapua.app.console.module.authentication.server;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.concurrent.Callable;
-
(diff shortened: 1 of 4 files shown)


ctron:
Is that a security issue?

Coduz:
Uhmmm I don't see any. 
What is your concern?

ctron:
No concern, just asking.

Coduz:
Ah ok. 
Well the methods act exactly like the `.login(...)`, but does not interfere with the current session of the user. This is handy when you need to make a "change password" feature. In that case you have a logged user which only needs to verify the `oldPassword` without interfering with the current session. All the lockout policy constraints still apply in case of too many wrong attempts. So all should be fine regarding the security

Issue #1326, opened 2018-01-26
No Client ID Column in Jobs


LeoNerdoG:
When user is adding devices as targets to certain job, a list with devices appears (the _**Add targets**_ dialogue box) where devices can be added according to their Device ID (human readable name). 
But when the device is added to the target list, there is no _**Device ID**_ column, only _**Job Target Id**_, which are some _random_ characters and numbers. 
I suggest that a _**Job Target ID**_ is replaced with _**Device ID**_, because it is more user-friendly and easier to understand and relate to certain device. 

Testflow: 
1. Login as kapua-sys
2. Go to Jobs, create a job (e.g. job1)
3. Select job1 and add a target. 
**When target is added, user cannot see the name, because there is no Device ID oclumn available.**

![screen shot 2018-01-26 at 10 39 19](https://user-images.githubusercontent.com/15121999/35434233-b9007edc-0286-11e8-9f05-e4e27edecf70.png)

Browsers used:
- Google Chrome (Version 63.0.3239.132 (Official Build) (64-bit)))
- Safari (Version 11.0.2 (13604.4.7.1.6))
- Mozilla Firefox (58.0 (64-bit))
- Opera (Version 50.0.2762.67)

lorthirk:
This is similar to the username resolution used in other views. You should somehow query the DeviceRegistryService to obtain the Client ID for that given target.

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1409, opened 2018-02-14
Replaced target id with client id on jobs target list

muros-ct:
Solution to this problem is solved by separate query on device by its
PK and using client id for more readable display of selected target.
Each target has to be queried, which might cause problems in very long
targets lists.

This solves issue #1326.


LeoNerdoG:
@muros-ct I have tested this fix and there is one problem. Regardless of how many devices are added - if user clicks on _**Client ID**_ in Job -> Targets, Job Target window freezes (look at the bottom screenshot) and whole browser window has to be refreshed to continue with normal operations. 
I have added 300 devices at once and it took Kapua 2,5 seconds to add all, so I think this lag is not significant (even for 1000+ devices it will be only a few seconds).

screen shot 2018-02-15 at 10 15 48

muros-ct:
@LeoNerdoG This is because there is still sorting enabled on Client Id. Sorting is not possible, because name of client is retrieved by secondary queries. Sort works only on first query that fills data about targets into grid.

@Coduz This might be a problem for all grids that would fetch data this way, by issuing secondary service query.

muros-ct:
Soved issue for now by removing possibility to sort from Client Id and Error Message column.

LeoNerdoG:
I have verified that there is no more freezing. This can be merged when ready.

Issue #1357, opened 2018-02-05
"Select All Targets" Button in Jobs Does Not Work


LeoNerdoG:
When user selects _**Selected only**_ in Jobs -> Targets -> Add and wants to select all the targets by checking the appropriate field, Kapua does not react (look at the screenshot). 

Testflow: 
1. Login as kapua-sys
2. Go to Devices, add several devices (e.g. dev1, dev2, dev3...)
3. Go to Jobs, add  job (e.g. job1)
4. Select job1 and go to Targets -> Add
5. Select _**Selected only**_ and try to check box for selecting all devices
_**The button, for selecting all devices does not work.**_

![screen shot 2018-02-05 at 09 19 20](https://user-images.githubusercontent.com/15121999/35794507-4da4e0e2-0a56-11e8-80be-ded680d92229.png)

Browsers used:
- Google Chrome (Version 64.0.3282.119 (Official Build) (64-bit)))
- Safari (Version 11.0.3 (13604.3.5))
- Mozilla Firefox (58.0.1  (64-bit))
- Opera (Version 50.0.2762.67)

muros-ct:
@LeoNerdoG Please verify again. Now select / unselect should work.
This one is just for you.

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1426, opened 2018-02-15
Fixing multi select checkbox on selecting Targets for Job

muros-ct:
Changes were needed on EntityGrid because implementation was not
according to GXT rules. Setting of grid was moved before onRender
method.

This was done in a manner that should leave existing usage of EntityGrid
in tact. Test of this fix should be extended to other girds not only
on this single one.

This fixes issue #1357


LeoNerdoG:
I have verified that this PR successfully fixes issue #1357 .

PR #1427, opened 2018-02-16
Feature console flexible permissions

### LINKED ISSUE #897: Allow for permission extensibility on GWT Console
Currently the permissions for the GWT Console are stored as single, hardcoded variables in `GwtAuthorizationServiceImpl`. We need those permissions to be extensible by third party modules, that should be able to introduce even new domains to check authorization against.

Coduz:
This PR closes #897 

This still the first draft. 

Basically on login the user pemissions are loaded into the `GwtSession`. On permission check, each UI component that request it, provides the permission to check. The new `hasPermission(...)` method in `GwtSession` replicates a simplified access check like the one on `AuthorizationService`.
Even if this is a simplified check, it doesn't mean that introduces security vulnerabilities because the authorization checks on the services are still performed. 

Next improvements:

- [x] Make `String` action parameter a GWT class
- [x]  For each modules define specific extensions of `GwtSessionPermission` to allow ease of use in UI checks, clean up the code and make it more maintainable.


Coduz:
A new `hasPermission(...)` has been added to allow each module to provide reference to specific module permissions (like `UserSessionPermission.read()` identifies `user:read` permission) and now all modules are using those when checking permissions. 
This allow a little clarity on the code.

Issue #1351, opened 2018-02-01
Device View Freezes if No Device Selected And "Change Pages Buttons" Pressed


LeoNerdoG:
This issue is connected to #1182 .
If user clicks any of the buttons in Device -> Events (_**First Page**_, _**Previous Page**_, _**Next Page**_, _**Last Page**_), when no device is selected, the lower part of Device view freezes (look at the screenshot). 

Testflow: 
1. Login as kapua-sys
2. Go to Devices
3. Without selecting any device go to _**Events**_ and click any of the buttons described above

![screen shot 2018-02-01 at 14 23 25](https://user-images.githubusercontent.com/15121999/35680779-8dd4d1e8-075b-11e8-81b2-88ae5da9b71a.png)

![screen shot 2018-02-01 at 14 22 39](https://user-images.githubusercontent.com/15121999/35680784-9114a1c6-075b-11e8-8b76-a06b8ad6dc21.png)

Browsers used:
- Google Chrome (Version 64.0.3282.119 (Official Build) (64-bit)))
- Safari (Version 11.0.3 (13604.3.5))
- Mozilla Firefox (58.0.1  (64-bit))
- Opera (Version 50.0.2762.67)

lorthirk:
This should be fixed automatically by #1182

LeoNerdoG:
I have verified that this issue has been resolved with commit [236feba](https://github.com/eclipse/kapua/commit/236feba0b9cdbfdd1c75e5859c42212acd65313b).

PR #1444, opened 2018-02-21
toolbar in Events tab is frozen when device is not selected

ct-goranpalibrk:


LeoNerdoG:
I have verified that this PR successfully fixes issue #1351.

Issue #1132, opened 2017-12-06
Discard Button In Child Accounts Not Working Properly


LeoNerdoG:
When user changes configuration of a child account (e.g. infiniteChildAccounts from False to True) but then reverts the change (as nothing had changed) _**Discard**_ button is non-functional (nothing happens, when pressed). 
From end-user point of view I see two possibilites: 
1. Freeze the button, if at the end configuration is not changed.
2. Revert changes as they were before user started to change them, when _**Discard**_ is presssed.

Testflow: 
1. Login as kapua-sys 
2. Go to Child Accounts, create a child account (e.g. kapua-child1)
3. Select kapua-child1, go to Settings
4. SelectinfiniteChildAccounts and it from **True** to **False** and tehn back to **True** (so effectively no changes are made)
**Discard button remains unfrozen and is not operational (if clicked, it does nothing).**

![screen shot 2017-12-06 at 10 22 27](https://user-images.githubusercontent.com/15121999/33654027-2a6a97fa-da6f-11e7-9a1c-eebd25fcf3a8.png)

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.39)

lorthirk:
Hi @LeoNerdoG ,

I couldn't replicate the issue. Could you please verify?

Thanks!

stefanomorson:
#Expected results:
Same as #1305

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1456, opened 2018-02-26
Discard button in AccountSettings tab in Child Accounts view now works properly

ct-goranpalibrk:


LeoNerdoG:
I have verified that this PR successfully fixes issue #1132.

Issue #1314, opened 2018-01-19
Kapua Reports No Error When Package Cannot Be Installed


LeoNerdoG:
When user tries to install a package and the _**Package Name**_ or _**Package Version**_ fields are longer than 125 characters/symbols, Kapua reports no errors (just info message that request has been sent to device), although in Kura log there is an error message stating that fields mentioned contain too long strings and package cannot be installed. 

Testflow: 
1. Login as kapua-sys
2. Connect a live device to Kapua
3. Go to Devices, select the live device, go to packages and click _**Install**_
4. In _**Package URL**_ put [link to the example publisher](http://download.eclipse.org/kura/releases/3.1.1/org.eclipse.kura.example.publisher_1.0.201.dp)
5. in _**Package Name**_ and _**Package Version**_ put string (<125 charactes)
**After clicking on Submit button, Kapua reports that request has been sent to the device and that the progres can be seen in _In progress_ sub tab, but in Kura log there is an error stating that two of the parameteres are too long.**

Screenshot No. 1 (Info message from Kapua): 
<img width="1438" alt="screen shot 2018-01-19 at 11 25 44" src="https://user-images.githubusercontent.com/15121999/35148187-9decf370-fd11-11e7-9766-863625984dcd.png">

Screenshot No. 2 (Error log from Kura device):
<img width="1418" alt="screen shot 2018-01-19 at 12 11 44" src="https://user-images.githubusercontent.com/15121999/35148277-d43bf25a-fd11-11e7-9d1c-389955a583ea.png">

Browsers used:
- Google Chrome (Version 63.0.3239.132 (Official Build) (64-bit)))
- Safari (Version 11.0.2 (13604.3.5))
- Mozilla Firefox (57.0.4 (64-bit))
- Opera (Version 50.0.2762.45)

stefanomorson:
#### Expected result
We are fixing this as part of an extension. See issue #1322

lorthirk:
@stefanomorson how about impose a character limit in those textboxes while #1322 is fixed? It should be a quick workaround while we implement the actual feature

LeoNerdoG:
I have verified that this issue is resolved (with limit on textboxes).

PR #1467, opened 2018-03-01
Set max length of characters in Install package Dialog #1314

ct-goranpalibrk:


ct-goranpalibrk:
Fix for issue #1314

LeoNerdoG:
I have verified that this PR solves (postpones solution - look at issue comments) issue #1314.

Issue #1035, opened 2017-11-22
Toolbar Of Selected Device Not Frozen As It Should Be


LeoNerdoG:
Problem emerges when creating a user (e.g. kapua-user1) and granting Device permissions. Although we grant _**ALL**_ device permissions and none of device connection, event, lifecycle or management permissions, toolbar of selected device (bottom toolbar) of kapua-user1 is not frozen (screenshot No. 1) until user clicks on one of the devices (screenshot No. 2) and Kapua returns a lot of errors. 
At this point it is OK for toolbar to be frozen, but nevertheless there should be no errors from Kapua. 

Testflow:
1. Login as kapua-sys
2. Go to Users, create a user (e.g. kapua-user1)
3. Select kapua-user1, add _**ALL**_ device permissions
4. Login as kapua-user1 and go to Devices.
**Screnshot No. 1 bellow shows that toolbar is not frozen (but it should be - as shown in screenshot No. 2).**

Screenshot No. 1: 
![screen shot 2017-11-22 at 13 16 49](https://user-images.githubusercontent.com/15121999/33127079-4505ecce-cf87-11e7-905d-7a6de01db2ec.png)

Screenshot No. 2: 
<img width="1439" alt="screen shot 2017-11-22 at 13 05 42" src="https://user-images.githubusercontent.com/15121999/33127056-2e5722f4-cf87-11e7-9ebf-6698d7eecac6.png">

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.39)

lorthirk:
We should check the various `*TabDescriptor` `isEnabled()` method

LeoNerdoG:
I have verified that this issue has been resolved.

Issue #1222, opened 2017-12-20
Commands Can Be Executed Although No Device Selected


LeoNerdoG:
When user uses only keyboard for navigating through kapua, it can _execute_ commands, althoguh no device is selected (look at the screenshots).

Testflow: 
1. Login as kapua-sys
2. With keyboard keys (_**TAB**_ and _**arrow**_ keys) navigate to _**Command**_ tab of device tab (no device should be selected)
3. With _**TAB**_ key navigate to **Execute** tab and enter a command (e.g. _**ls -l**_ or **sudo shutdown**). 
**At this point Kapua chages info window from _No Device selected_ to _Executing command_.**

Screenshot No. 1 (Entering the command):
![screen shot 2017-12-20 at 13 59 50](https://user-images.githubusercontent.com/15121999/34212896-0466e4de-e59e-11e7-8f15-4df71a4981a7.png)

Screenshot No. 2 (Executing the command): 
![screen shot 2017-12-20 at 14 00 02](https://user-images.githubusercontent.com/15121999/34212918-1415bae0-e59e-11e7-9c8f-5acb8c719c10.png)

Screenshot No. 3 (Kapua reports that it is _executing the command_ ):
![screen shot 2017-12-20 at 14 00 12](https://user-images.githubusercontent.com/15121999/34212939-20bba8b8-e59e-11e7-9226-54c2e8f930e9.png)

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0.1 (64-bit))
- Opera (Version 49.0.2725.39)

lorthirk:
We should prevent keyboard based navigation in disabled components. Perhaps disable the whole panel instead of only masking it?

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1469, opened 2018-03-02
Tabs in Devices are frozen when device is not selected

ct-goranpalibrk:


LeoNerdoG:
I have verified that this PR fixes issue #1035.

console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/history/DeviceTabHistory.java
```diff
@@ -97,6 +97,7 @@ public DeviceTabHistory(GwtSession currentSession) {
         super(currentSession, DEVICES_MSGS.tabHistory(), new KapuaIcon(IconSet.HISTORY));
         this.currentSession = currentSession;
         initialized = false;
+        setEnabled(false);
```
lorthirk: I don't know if disabling this tab once a Device is deselected is a good idea. Since Events tab doesn't require a communication with the Device itself, and it's only a registry operation, we can just keep it enabled just like any other tab that contains a "slave" grid and shows a "No items" message when no items are selected in a "master" grid.

console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/tag/DeviceTabTags.java
```diff
@@ -29,6 +29,7 @@
     public DeviceTabTags(GwtSession currentSession) {
         super(currentSession, DEVICES_MSGS.tabTagsTitle(), new KapuaIcon(IconSet.TAGS));
         deviceTagGrid = new DeviceTagGrid(null, currentSession, selectedEntity);
+        setEnabled(false);
```
lorthirk: What I'm thinking for the Events tab also stands true for this one. I wouldn't disable it.

console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/tag/DeviceTabTags.java
```diff
@@ -41,6 +42,7 @@ protected void onRender(Element parent, int index) {
     @Override
     public void setEntity(GwtDevice selectedDevice) {
         super.setEntity(selectedDevice);
+        setEnabled(selectedDevice != null);
```
lorthirk: The same should happen when deselecting a device; there's no need to disable this tab.

ct-goranpalibrk:
This PR also should fix issues #1314 (max number of characters changed to 125)  and #1222

LeoNerdoG:
I have verified that this PR successfully fixes issue #1035. There is still problem in Jobs, but after Description tab is added, it should be ok.

lorthirk:
Jobs Description tab has been merged in develop with commit 841cce66b

lorthirk:
I would like to test this before merging to see the actual effect from a UX perspective

LeoNerdoG:
@lorthirk @ct-goranpalibrk I have merged current develop branch with this one and now everything works fine, so it can be merged when all checks complete.

PR #1476, opened 2018-03-06
User Tests Fixes

lorthirk:
Since commit 3e99d6febc8a2475c1a48657faaee9371a0ae73a removed `@Inject` notation from `UserServiceImpl`, tests are now failing. This PR fixes mocked service location in `UserServiceSteps` so that they are all passing again. Probably we want to plan a final refactoring of `@Inject` everywhere since this notation has been deprecated.


muros-ct:
This fixes issue wiht UserService tests failure.
It can be merged.
I will open new issue for review and cleanup of Injection in other service implementations and tests.

Issue #1368, opened 2018-02-06
 No Description When Editing Device


LeoNerdoG:
There is no description in dialogue box when editing device (look at the screenshot). 

Testflow: 
1. Login as kapua-sys
2. Go to Devices 
3. Add a device (e.g. device1)
4. Select device1 and click _**edit**_
**Look at the screenshot bellow - there is no text in dialogue box (but there is in all the others).**

![screen shot 2018-02-06 at 12 57 59](https://user-images.githubusercontent.com/15121999/35858682-a898394e-0b3d-11e8-9da1-14c51da51ed3.png)

Browsers used:
- Google Chrome (Version 64.0.3282.119 (Official Build) (64-bit)))
- Safari (Version 11.0.3 (13604.3.5))
- Mozilla Firefox (58.0.1  (64-bit))
- Opera (Version 50.0.2762.67)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1479, opened 2018-03-06
Added description when editing device

ct-ajovanovic:
Solves issue: No Description When Editing Device #1368


muros-ct:
@elbert3 can you check this. You left FIXME with empty message.
Is there any reason for that, and is this fix ok?

Issue #1110, opened 2017-12-01
Non-specific Error When Adding Same Role To User


LeoNerdoG:
Kapua returns a non-specific (_**Persistance error**_) error/info message, when user tries to add same role to user twice. 

Testflow:
1. Login as kapua-sys
2. Go to Roles, create a role (e.g. role1)
3. Go to users, create a user and try to add same role to it twice.
**Kapua returns an _persistance_ error with non-specific info message (look at the screenshot). Message should imply, what the problem is.**

<img width="1436" alt="screen shot 2017-12-01 at 15 54 15" src="https://user-images.githubusercontent.com/15121999/33488111-e5302398-d6af-11e7-9e71-268fea87cc5e.png">

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.39)

LeoNerdoG:
The described error has sligthly changed and now says that _**an entity with same unique value already exists**_. The first part is still problematic ( _**Role creation failed:....**_) as user is not crating roles, but just removing it from user. 
The message should be: _**Error when associationg role: An entity with the same value for field already exists.**_

Screenshot: 
![screen shot 2018-02-12 at 13 18 33](https://user-images.githubusercontent.com/15121999/36096656-8af21fb8-0ff7-11e8-9962-bdad6583f180.png)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1480, opened 2018-03-06
Fixing issue with adding duplicate role to user

muros-ct:
Before role is added to user, check is made if user already has
this role and if it does, it reports duplicate role error and not
generic duplicate entity error during persistence operation.

This solves issue #1110.


LeoNerdoG:
I have verified that this PR successfully resolves issue #1110.

Issue #1137, opened 2017-12-06
No Error Message For Non-Saved Changes When Leaving Devices Tab


LeoNerdoG:
When user adds or connects a device and then changes its configuration - if changes are not saved when selecting some other parameter, Kapua returns an error aksing user if save or discard changes (look at the screenshot).
However, if user selects any other _main_ tab (Tags, Roles, Child Accounts...) Kapua does not ask user for saving/discarding changes. 

Testflow: 
1. Login as kapua-sys 
2. Connect live Kura device to Kapua (e.g. Raspberry pi)
3. Go to Devices, select live device, go to Configuration.
4. Change some Configuration parameters (ClockService...) but do not save them
5. Go to Roles/Groups/Tags/...
**Kapua does not warn user that there are unsaved changes, but instead it discards them automatically.**

Screenshot (This kind of error should be reported, if user changed configuration and left Devices tab without saving them):
<img width="1434" alt="screen shot 2017-12-06 at 12 24 16" src="https://user-images.githubusercontent.com/15121999/33659742-94e40546-da81-11e7-959a-1a1426c02844.png">

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.39)

lorthirk:
The best option would be to add the same listener we now have on service change to tab change as well

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1493, opened 2018-03-08
Fixing issue with leaving dirty input forms

muros-ct:
There are forms on UI that can be left in dirty state by using main
menu - West menu.
This was solved by adding dirty form flag on gwtSession object and
when user uses West menu it is notified that changes will be lost and
can decide if this is ok or not.

Changes were required to WestNavigationView and that could break unexpected
functionality of UI. It was tested but not overall UI tested.

This fixes issue #1137.

### DIFF
--- a/console/core/src/main/java/org/eclipse/kapua/app/console/core/client/WestNavigationView.java
+++ b/console/core/src/main/java/org/eclipse/kapua/app/console/core/client/WestNavigationView.java
@@ -19,12 +19,15 @@ import com.extjs.gxt.ui.client.data.ModelData;
 import com.extjs.gxt.ui.client.event.ComponentEvent;
 import com.extjs.gxt.ui.client.event.Events;
 import com.extjs.gxt.ui.client.event.Listener;
+import com.extjs.gxt.ui.client.event.MessageBoxEvent;
 import com.extjs.gxt.ui.client.event.SelectionChangedEvent;
 import com.extjs.gxt.ui.client.event.SelectionChangedListener;
 import com.extjs.gxt.ui.client.store.ListStore;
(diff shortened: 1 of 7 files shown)


muros-ct:
@lorthirk Please take a look as I am changing gwtSession and WestNavigationView.
@LeoNerdoG will check functionality and see if I broke something.

LeoNerdoG:
@muros-ct 
I have tested this fix and found a problem with some fields (I am using Raspberry Pi 3 with Kura 3.1.1).
I still haven't finished testing, but if I change for example **Clock.ntp.host** field, Kapua does not open the confirmaiton field, but if I change **Clock.ntp.port** field, the confirmation window pops-up. 
Can you please check this?

muros-ct:
It probably has to do with field loosing focus and because field doesn't have validator, field is not marked as changed. If field is changed and moved to other field it still works, or if field such as radio is changed.
I will look intoi it and try to solve it if at all possible.

muros-ct:
I pushed changes that solve issue with some field changes not invoking confirmation dialog.
I fixed on TextField and TextArea (available on Camel rules configuration).
@LeoNerdoG please evaluate.
@lorthirk Please see changes on KapuaTextField and added KapuaTextArea.

muros-ct:
I looked at TriggerField implementation that is using triggerBlur.
mimicing prevents onFocus to trigger muliple focusEventPreview.

LeoNerdoG:
I have verified again and now also the password fields are OK. You can merge when ready.

muros-ct:
@lorthirk Is it ok if I merge?

lorthirk:
Ok for me. @Coduz do you agree?

PR #1494, opened 2018-03-08
Removed not used DeviceConnectionSummary class

lorthirk:
Hi all,

This PR removes `DeviceConnectionSummary` since it's not used in the code.


Coduz:
Fixed title

Issue #1256, opened 2018-01-09
Page Toolbar Freezes, if Upper Refresh Button in Devices Clicked Too Fast


LeoNerdoG:
If upper **refresh** button in Devices (look at the screenshot - green circle) is pressed too quickly (e.g. holding _**enter**_ key when refresh selected), the page toolbar, used for navigationg through pages freezes. User cannot manually enter page number, but can only press buttons for changing pages. 
Also the **refresh** button in this toolbar is frozen. 

Testflow: 
1. Login as kapua-sys
2. Go to Devices, add several devices (e.g. device1, device2,...)
3. Navigate to upper **refresh** button and press enter key on keyboard (so that Kapua starts refreshing page)
**When refreshing is complete, look at the bottom page toolbar. Number box and refresh button are frozen.**

![screen shot 2018-01-09 at 10 21 02](https://user-images.githubusercontent.com/15121999/34714417-b1de0350-f529-11e7-9b58-5a4a56e3769f.png)

Browsers used:
- Google Chrome (Version 63.0.3239.132 (Official Build) (64-bit)))
- Safari (Version 11.0.2 (13604.3.5))
- Mozilla Firefox (57.0.4 (64-bit))
- Opera (Version 50.0.2762.45)

LeoNerdoG:
I have verified that this issue has been resolved.

Issue #1259, opened 2018-01-09
Page Toolbar Freezes, if Upper Refresh Button in Roles Clicked Too Fast


LeoNerdoG:
If upper **refresh** button in Roles (look at the screenshot - green circle) is pressed too quickly (e.g. holding _**enter**_ key when refresh selected), the page toolbar, used for navigationg through pages freezes. User cannot manually enter page number, but can only press buttons for changing pages. 
Also the **refresh** button in this toolbar is frozen. 

Testflow: 
1. Login as kapua-sys
2. Go to Roles, add several roles (e.g. role1, role2,...)
3. Navigate to upper **refresh** button and press enter key on keyboard (so that Kapua starts refreshing page)
**When refreshing is complete, look at the bottom page toolbar. Number box and refresh button are frozen.**

![screen shot 2018-01-09 at 10 28 29](https://user-images.githubusercontent.com/15121999/34714639-5c44ad4e-f52a-11e7-927c-96075b4db4c9.png)

Browsers used:
- Google Chrome (Version 63.0.3239.132 (Official Build) (64-bit)))
- Safari (Version 11.0.2 (13604.3.5))
- Mozilla Firefox (57.0.4 (64-bit))
- Opera (Version 50.0.2762.45)

muros-ct:
Same as #1256

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1496, opened 2018-03-08
Fixing issue with refresh button on entity grids

muros-ct:
This solves issues that are present on all entity grids.
Now when refresh button is clicked and refresh method called on EntityGrid,
that button is disabled and enabled back only when load is finished or
exception is raised.

This solves issues:
- 1256
- 1257
- 1258
- 1259
- 1260
- 1261
- 1262
- 1263


LeoNerdoG:
I have verified that this PR successfully fixes issues #1256,#1257,#1258,#1259,#1260,#1261,#1262,#1263.

Issue #1129, opened 2017-12-05
Exporting Device Events To .csv Does Not Work


LeoNerdoG:
When certain device has more than two page of entries (500+ events), export to .csv file does not work. Instead of downloading a .csv file, a new browser window opens, showing entries in plain text (secreenshot No. 2).

Testflow: 
1. Login as kapua-sys
2. Connect a live Kura device to Kapua
3. Create 500+ events with this device
4. Try to download event log as .csv file
**At this point Kapua only opens new window without downloading the .csv file.**

Screenshot No. 1 (trying to export event log into .csv file):
<img width="1437" alt="screen shot 2017-12-05 at 11 28 34" src="https://user-images.githubusercontent.com/15121999/33603140-de351d14-d9b1-11e7-95d2-2b6b9529e826.png">

Screenshot No. 2 (a new browser window opens with event log in plain text)
<img width="1436" alt="screen shot 2017-12-05 at 11 28 42" src="https://user-images.githubusercontent.com/15121999/33603143-dfc1a1a2-d9b1-11e7-8c8d-867a4c3c8521.png">

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.39)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1503, opened 2018-03-13
Disable export button

bojanutkovic:
#1129 
Disable export button if device isn't selected


LeoNerdoG:
This PR can be closed as PR #1505 is solving problem with enabled Events tab when no device is selected.

Coduz:
Hi @LeoNerdoG ,

This PR will be closed since it does not solve #1129 and the way of disabling the button is not correct. 
Moreover, as you said, the disabling of the button when no device is selected if fixed in #1505 . 

Can you please  confirm this, so we can reject this PR?

LeoNerdoG:
@Coduz yes, from my standpoint it can be closed. 

Regards, 

Leonardo

Issue #1086, opened 2017-11-29
No Jobs Filter


LeoNerdoG:
It would be very handy if Jobs had their own filter like Access Groups, Tags, Users, Roles and Child Accounts have them. 
I guess that users will have several jobs in the job list so I think it would be a nice feature. 

<img width="1440" alt="screen shot 2017-11-29 at 14 41 54" src="https://user-images.githubusercontent.com/15121999/33377857-3d8d368c-d513-11e7-9246-e692d44a82f8.png">

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.39)

LeoNerdoG:
I have verified that this issue is resolved.

PR #1504, opened 2018-03-13
Added filter panel to Batch Job grid

muros-ct:
Standard filter panel was added to Batch Job grid.
Filter includes just job name.

Additionally EntityCRUDToolbar was fixed. As Jobs grid had additional
button in Start that was wrongly positioned when filter button was added.
This is now fixed and also closes FIXME comment in same part of code.

This fixes issue #1086.


muros-ct:
@lorthirk Can you check EntityCRUDToolbar changes. It solves issue as it is. I can not find other quick fix. Any suggestion.

lorthirk:
Looks good to me. I'm working in a refactor of the whole toolbar to make it as standard as possible, and I took a similar approach.

LeoNerdoG:
I have verified that this PR successfully fixes issue #1086.

PR #1505, opened 2018-03-13
tags and events tab disabled when device is not selected in Device view

ct-goranpalibrk:

### DIFF
--- a/console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/history/DeviceTabHistory.java
+++ b/console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/history/DeviceTabHistory.java
@@ -61,6 +61,7 @@ import org.eclipse.kapua.app.console.module.api.shared.model.session.GwtSession;
 import org.eclipse.kapua.app.console.module.device.client.messages.ConsoleDeviceMessages;
 import org.eclipse.kapua.app.console.module.device.shared.model.GwtDevice;
 import org.eclipse.kapua.app.console.module.device.shared.model.GwtDeviceEvent;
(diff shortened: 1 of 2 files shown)


LeoNerdoG:
Hi @ct-goranpalibrk 
I have tested this, but now all tabs (except Description tab) are disabled, although device that is  _Disconnected_ or _Unknown_ selected. Can you please check this? 
If live device is selected or no device is selected, everything works OK.

LeoNerdoG:
I have retested with the newest changes and it is ok. It can be mreged when ready.

console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/history/DeviceTabHistory.java
```diff
@@ -103,11 +102,7 @@ public DeviceTabHistory(GwtSession currentSession) {
     @Override
     public void setEntity(GwtDevice gwtDevice) {
         super.setEntity(gwtDevice);
-        setEnabled(gwtDevice != null &&
-                gwtDevice.isOnline() &&
-                currentSession.hasPermission(DeviceManagementSessionPermission.read()) &&
-                (gwtDevice.hasApplication(GwtDevice.GwtDeviceApplication.APP_DEPLOY_V1) || (gwtDevice.hasApplication(GwtDevice.GwtDeviceApplication.APP_DEPLOY_V2))));
-
+        setEnabled(gwtDevice != null);
```
muros-ct: IMHO I don't like this kind of coding, especially as one line before gwtDevice is set.
It would be much nicer if both lines would be enclosed in if and setEnabled(true).

Issue #1254, opened 2018-01-05
Unknown Settings Column in About Tab


LeoNerdoG:
When entering Kapua and using only keyboard to navigate through its tabs, an unknown and unclickable_**Settings**_ column appears at the bottom (look at the screenshot). If user after that uses mouse, to click another tab (e.g. _**Users**_) and then returns, the column disappears. 

Testflow: 
_**Using only keyborad!**_
1. Login as kapua-sys
2. With arrow keys and Tab navigate to _**About**_ tab. 
_**At this point an unknown column appears as seen on screenshot bellow and described above.**_

![screen shot 2018-01-05 at 09 36 56](https://user-images.githubusercontent.com/15121999/34601662-51665a7e-f1fd-11e7-819f-7b6e5b2e01dc.png)

Browsers used:
- Google Chrome (Version 63.0.3202.84 (Official Build) (64-bit))
- Safari (Version 11.0.2 (13604.3.5))
- Mozilla Firefox (57.0.3 (64-bit))
- Opera (Version 49.0.2725.64)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1510, opened 2018-03-14
Fixing wrong redraw on fast menu change

muros-ct:
When moving through menu items with cursor this is very fast and
causes settings which is redrawn asynchronously to redraw when About is
already redrawn as next item. This causes display of Settings header in
footer over the copyright.

I added checking in Settings redraw if selection is still on settings and
if not redraw of settings is skipped.

This fixes issue #1254.


LeoNerdoG:
I have verified that this PR successfully fixes issue #1254.

Issue #1075, opened 2017-11-27
Error Starting Job Although User Has Proper Permissions


LeoNerdoG:
When adding _**ALL**_ Job permissions to user, jobs can be successfully executed, but if we add _**read**_, _**write**_ and _**delete**_ job permissions, jobs cannot be executed (look at the screenshot bellow)

Testflow:
1. Login as kapua-sys
2. Go to Users, create a user (e.g. kapua-user1)
3. Select kapua-user1, add read, write and delete job permissions
4. Login as kapua-user1, go to Jobs, create and start the job
**At this point Kapua shows user needs _jobs:execute_ permissions, which cannot be added.**

<img width="1428" alt="screen shot 2017-11-27 at 15 05 51" src="https://user-images.githubusercontent.com/15121999/33270578-d061a92e-d384-11e7-8083-c844d368ee14.png">

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.39)

bojanutkovic:
Here we need to add execute action in ATHZ_DOMAIN_ACTIONS table for a job. After that we can grant execute permission to user and user will able to start a job.

LeoNerdoG:
I have verified that this issue ahs been resolved.

PR #1513, opened 2018-03-15
Execute permission added to Job Service

muros-ct:
Job domain was lacking execute permission which could be assigned to user.
Now (1.0.0 liquibase change) it has it and can be assigned to user and
that user can execute job.

This fixes issue #1075.


muros-ct:
@Coduz Can you please verify this change. Did I correctly use liquibase xml.

LeoNerdoG:
I have verified that this PR successfully adds execute permission and fixes issue  #1075.

PR #1515, opened 2018-03-15
Added Endpoint Info view

### LINKED ISSUE #1440: Implement a way to define the Default EndpointInfo of the deployment
Currently, managing of this property can be done only through  the REST API or by specifying the hostname via Java properties on startup. The last one interferes with other usages of that value. 

We need a clean way to set it for a brand new deployment

Coduz:
This PR closes #1440 

I created a view for the definition of the endpoints of the system.

Since Kapua can work without any definition of those, I didn't added documentation for installing Kapua.


console/module/api/src/main/resources/org/eclipse/kapua/app/console/module/api/client/messages/ValidationMessages.properties
```diff
@@ -70,7 +70,7 @@ OVER_RULE_LIMIT=The maximum number of rules allowed for this account has been re
 SUBJECT_UNAUTHORIZED=User doesn't have permission to perform this action.
 ENTITY_ALREADY_EXISTS=An entity with the same unique value already exists.
 UNABLE_TO_PARSE_CRON_EXPRESSION=Unable to parse cron expression. {0}
-ENTITY_ALREADY_EXIST_IN_ANOTHER_ACCOUNT= {0} already exists in another account.
+ENTITY_ALREADY_EXIST_IN_ANOTHER_ACCOUNT=git st{0} already exists in another account.
```
lorthirk: I smell a terminal leak here... 😄

LeoNerdoG:
@lorthirk @Coduz I am on it! But it will take some time as this is a whole new tab. Nevertheless I think I will be able to finish today.

LeoNerdoG:
Hey @Coduz I have started testing Endpoints and already found some issues: 

Domain fields: 
 - The first field name should be Schema instead of Domain
 - Any symbol can be entered intoboth Domain fileds. Shouldn't they be limited to a-zA-Z0-9 and some special symbols (e.g. /,:,;,-,_,'',(),@?#[],$*~!)?

Port field:
 - If entering name manually: "-" and "." can be entered (although only numbers should pe permitted)
 - If copy-pasting: any symobl can be entered through copy-paste
 - Error for first field (Max length for this field is 64) is shown after user wants to edit a certain endpoint and not at creation.
 - If Port field has a limit, why not limit to 5 digits instead of 6 (max vlaue is 65536)

General: 
 - Endpoints with same name can be created multiple times
 - The name of endpoint on top of the window not OK when editing Endpoint
 - The name of endpoint on top of the window not OK  when deleting Endpoint
 - List of Endpoints is the same throughout the accounts (if endpoint is created with kapua-sys@kapua-sys, it is also visible in all of its child acounts)
 - Cannot see Settings if created even one endpoint (Kapua returns an error saying "An internal error occurred: Unable to build URI for EndpointInfo: xxxxxxxxxxx")

LeoNerdoG:
@Coduz I have finished testing and these are the rest of my findings: 

General:
- Entity and Endpoint instead Entity and Tag in Description info of an Endpoint
- Should there be an Endpoints settings (like Tags and Roles...) when creating child account?
- Kapua returns an error saying "An internal error occurred: Unable to build URI for EndpointInfo: Nm4WkJ3OzpU" when user clicks a child account in Child Accounts
- When editing Endpoint the Text should be: "Edit Endpoints parameters" and not "Edit Endpoint's name", because user is not editing just its name. 

Permissions:
- Permissions for EndpointInfo do not work (user can see Endpoints if has ALL ALL ALL permissions. In every other case, Endpoints are not visible).

Issue #846, opened 2017-10-09
Problem With Selecting Metrics


LeoNerdoG:
I have noticed that when selecting metrics from Metric list, if you miss click one metric the other ones get deselected and the miss clicked one gets selected (look at the images below). 

**First the metric.string, metric.byte and metric. long were selected:** 

<img width="1440" alt="screen shot 2017-10-09 at 15 40 45" src="https://user-images.githubusercontent.com/15121999/31341282-549102be-ad09-11e7-8dcb-48bc2b34a6d3.png">

**And after that I miss clicked on metric.char and all other ones got deselected:** 

<img width="1440" alt="screen shot 2017-10-09 at 15 40 54" src="https://user-images.githubusercontent.com/15121999/31341287-59dc8b3a-ad09-11e7-9a78-2f9524e1eff1.png">

This has been tested on macOS High Sierra (10.12.6) and web browsers: 
 - Google Chrome (Version 61.0.3163.100 (Official Build) (64-bit))
 - Safari (Version 11.0 (12604.1.38.1.7))
 - Mozilla Firefox (56.0.1 (64-bit))

stefanomorson:
@LeoNerdoG, @lorthirk: I was eventually able to reproduce the behaviour described. The issue happens when the click is made out of the checkbox area but within the area of the metric.char item. Is this the way to reproduce it ?

LeoNerdoG:
Hi @stefanomorson ,
yes, this is the right way to reproduce the issue.

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1519, opened 2018-03-16
Fix for selection issue on Metrics grid

muros-ct:
Solved selection of check boxes and columns, now it works with multiple
select, without deselecting when column and not checkbox is selected.

Problem is hidden in order of method calls and modified selection listener
that was already used in Job target selection.

This fixes issue #846


LeoNerdoG:
I have verified that this PR successfully resolves issue #846.

PR #1525, opened 2018-03-20
Toolbar buttons access fixes

### LINKED ISSUE #1432: Add access checks in all UI element
Currently there are many UI components that are missing access checks to determine whether or nor they are enabled or not. 

Each tab/grid/button should take into consideration usage of `GwtSession.hasPermission()` with their respective permission to determine if it its status is disabled/enabled.

bojanutkovic:
#1432


LeoNerdoG:
I have teset this and found a couple of issues: 
1. When sepparatelly added READ, WRITE and DELETE Device Permissions, all the buttons are still disabled. 
2.  In Tags, Roles and Groups - if user has WRITE permissions, **edit** and **delete** buttons do not get disasbled when a Tag/Role/Group is deselected.
@bojanutkovic can you please recheck?

LeoNerdoG:
I have verified that commit 2a7540a resolved two issues I reported earlier. This PR can be merged.

Issue #1002, opened 2017-11-17
"ALL" Permissions Can Be Added to User Multiple Times


LeoNerdoG:
When creating a user and adding Permissions to it, ALL permissions can be added multiple times. 

Testflow: 
1. Login as kapua-sys
2. Go to Users, create new user (e.g. kapua-user1) 
3. Select kapua-user1, go to Permissions tab and add ALL permissions multiple times.

<img width="722" alt="screen shot 2017-11-17 at 13 45 11" src="https://user-images.githubusercontent.com/15121999/32948149-97cb5c7e-cb9e-11e7-99c2-aea156c18f34.png">

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.34)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1526, opened 2018-03-20
Fix adding same Permission to User

muros-ct:
Before permission is added to user it is checked if it already exists and
if so it displays error message explaining that.
Query is made on access permission with same attributes as permission is
created with.

This fixes issue #1002


LeoNerdoG:
I have verified that this PR successfully fixes issue #1002.

Issue #1225, opened 2017-12-21
Any Symbol Can Be Entered in Page Number Fields


LeoNerdoG:
In page number fields, which serve user to switch between pages, any symbol and also any number of symbols can be enteres, which can lead to Kapua crash. 

Testflow: 
1. Login as kapua-sys
2. Select any number enetring field and try to enter any number of all kinds of symbols
**If number exceeds several milion, Kapua freezes.**

Screenshot No. 1 (Error, when too big number entered in Child Accounts page field):
![screen shot 2017-12-21 at 09 15 45](https://user-images.githubusercontent.com/15121999/34248818-acd1836a-e637-11e7-87d4-f68fae638dcb.png)

Screenshot No. 2 (Error, when too big number entered in Roles page field):
![screen shot 2017-12-21 at 09 25 00](https://user-images.githubusercontent.com/15121999/34248822-b0ab1168-e637-11e7-94d4-fe19defea975.png)

Screenshot No. 3 (Kapua's GUI can also crash):
<img width="1439" alt="screen shot 2017-12-21 at 09 27 55" src="https://user-images.githubusercontent.com/15121999/34247361-7bcd83e0-e632-11e7-96d6-8a56f53578da.png">

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0.1 (64-bit))
- Opera (Version 49.0.2725.39)

lorthirk:
The TextBox that holds the current page is currently built-in in GXT PagingToolBar. We would need to find a safe way to replace it with a custom TextBox that can only holds a specified amount of characters.

muros-ct:
@lorthirk I tried to play arround with this paging toolbar and noticed that we already have similar KapuaPagingToolbar.
Can you take a look at this branch if this could be solution.

https://github.com/eclipse/kapua/tree/fix-pagingToolbarText

LeoNerdoG:
I have verified that this issue hasn been resolved. The field number still allows entering any symbol, but as this field is limited to five characters and enetering non-number returns a page number, this is resolved.

PR #1529, opened 2018-03-21
Proposal for fixing long text in paging toolbar

muros-ct:
This is possible solution to paging toolbar, where user can enter arbitrarily long
text for page number.

We already have KapuaPagingToolbar where I tested it and it works.

This fixes issue #1225.

### DIFF
--- a/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/ui/widget/KapuaPagingToolBar.java
+++ b/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/ui/widget/KapuaPagingToolBar.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017 Eurotech and/or its affiliates and others
+ * Copyright (c) 2017, 2018 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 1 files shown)


LeoNerdoG:
I have verified that this PR successfully resolves issue #1225.

console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/ui/widget/KapuaPagingToolBar.java
```diff
@@ -19,8 +19,17 @@
 
     private static final ConsoleMessages MSGS = GWT.create(ConsoleMessages.class);
 
+    /**
+     * Max length of user input field for page number.
+     */
+    public static final String MAX_TEXT_LEN = "4";
```
lorthirk: Wouldn't it be safer to use an `Integer` here, and use its `toString()` in the `setAttibute()` later on?

muros-ct: Shure, I didn't know from which side to go.

lorthirk: Well, at least using `Integer` you can be certain that only an integer will be written in the variable :smile:

muros-ct:
@lorthirk I changed to integer.

Issue #1528, opened 2018-03-21
Wrong Domain Name For Broker, EndpointInfo and Stream Permissions


LeoNerdoG:
When user adds permissions for broker (connect), EndpointInfo (read, write, delete) and Stream (read writ delete) their Domains are labeled as ALL instead of Broker, EndpointInfo and Stream (look at the screenshot). 
Because of this, user cannot add EndpointInfo and Stream permissions to a user , as they have same label ("ALL"). 

Testflow: 
1. Login as kapua-sys
2. Go to Users, create a user (e.g. kapua-user1)
3. Select kapua-user1 and try to add Broker Connect permissions, and EndpointInfo/Stream read, write and delete permissions sepparatelly. 
**As seen on the screenshot, kapua labels all three of them as "ALL" instead of proper domain names.**

Screenshot: 
![screen shot 2018-03-21 at 08 14 45](https://user-images.githubusercontent.com/15121999/37698131-5cd3816a-2ce1-11e8-80ab-bde900c82923.png)

Browsers used: 
- Google Chrome (Version 65.0.3325.162 (Official Build) (64-bit)
- Safari (Version 11.0.3 (13604.3.5))
- Mozilla Firefox (58.0.2  (64-bit))
- Opera (Version 51.0.2830.26)

Coduz:
In PR #1515  I've also changed the code to throw an exception instead of returning `null`.

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1530, opened 2018-03-21
Fix adding broker permissions

muros-ct:
Broker permission was not added because mapping was not available for
broker service.
Workaround for problem with domain mapping in case of non-existent service
in console is provided.

This fixes issue #1528


muros-ct:
@Coduz I rewrote the fix. I couldn't find cleaner solution.
Problem is that if service is not available in console it can't be used in assigning permisions.
Not to change too much code or break something else I used this sollution with annonymous inner implementation of Domain in case of non-existent service.
It works!

LeoNerdoG:
I have verified that new fixes successfully resolve problems described in issue #1528.

Issue #975, opened 2017-11-15
About Tab Divided in Two Non-arrangeable Pieces


LeoNerdoG:
Kapua's about tab is somehow divided in two pieces that cannot be rearranged (look at the screenshot). 

Testflow: 
1) Login as kapua-sys
2) Go to About tab

<img width="1439" alt="screen shot 2017-11-15 at 14 18 42" src="https://user-images.githubusercontent.com/15121999/32838384-05abbade-ca11-11e7-86fd-f50a986b2929.png">

Browsers used:
- Google Chrome (Version 61.0.3163.100 (Official Build) (64-bit))
- Safari (Version 11.0 (12604.1.38.1.7))
- Mozilla Firefox (56.0.1 (64-bit))
- Opera (Version 48.0.2685.39)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1535, opened 2018-03-22
Fix about tab

ct-goranpalibrk:
Hi all, 
This PR implements resize option in About tab, issue #975 .


LeoNerdoG:
I have verified that this PR successfully resolves issue #975.

Issue #1019, opened 2017-11-22
User Gets Logged Out If Browsers Refresh Button Is Pressed


LeoNerdoG:
If certain user does not have _**user**_ permissions it gets logged out if browsers _**refresh**_ button is pressed. Screenshot bellow shows how each of the permissions should be added. 

Testflow: 
1. Login as kapua-sys
2. Go to Users, create a user (e.g. kapua-user1)
3. Select kapua-user1, add all permissions to it (one by on) excepr _**user**_ permissions
4. Login as kapua-user1 and try to refresh the page. User gets logged out. 

<img width="722" alt="screen shot 2017-11-22 at 09 23 23" src="https://user-images.githubusercontent.com/15121999/33116862-ad238652-cf66-11e7-9488-64814d74b334.png">

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.39)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1538, opened 2018-03-23
Fix logout error is user refreshes browser

muros-ct:
Error was caused by exception raised when user was looked up in
security filter. I moved user lookup latter in the code where it is
really needed.
This resolves this issue. If issue will reemerge it might be necessary to
move user lookup code into privileged block.

This fixes issue #1019


LeoNerdoG:
I have verified that this PR sucessfully fixes issue #1019.

Issue #1073, opened 2017-11-27
Child Account Can Be Deleted Without Delete Permissions


LeoNerdoG:
User can delete child account if he has **_read and write permissions_** (without delete permissions). 

Testflow:
1. Login as kapua-sys
2. Go to Users, create a user (e.g. kapua-user1)
3. Select kapua-user1, add read or **_read and write Account permissions_**
4. Login as kapua-user1, go to Child Accounts and select one of child accounts and try to delete it.
**At this point Kapua should return an error that user has no _Account:delete_ permissions, but instead of that child account is deleted.**

Output when we delete child account without proper permissions:
<img width="1438" alt="screen shot 2017-11-27 at 13 29 28" src="https://user-images.githubusercontent.com/15121999/33266750-2b9a66ea-d377-11e7-83ee-d73bcf6a99cc.png">

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.39)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1539, opened 2018-03-23
Fix issue with account deletion permission

muros-ct:
This issue was already resolved but I found new one. User was not able
to add Account.
Solution is based on implementation that disables delete button is user
has no privilege for deletion. Wrong permission was connected to Add button.

This fixes issue #1073


muros-ct:
This issue is resolved, but it brings another issue into perspective.
If user has account:write permission in current account and creates account, that accoutn can not be edited unless permission account:write is forwardable.
Should we provide this explicitly as if we give write it has to be write for sub accounts not just parent accoutn, where we create sub-accounts.

Yes it sounds confusing :)

PR #1540, opened 2018-03-23
scroll-bar added in About view

ct-goranpalibrk:


LeoNerdoG:
I have verified that this PR resolves issue in About tab when browser window is too narrow and the text on the rigth is not entirely visible - a scroll bar is added.

PR #1541, opened 2018-03-23
Refactor Predicates in API

### LINKED ISSUE #1346: Promote query *Entity*Predicates classes from implementation to api module
We need to promote the *entity*Predicates classes from implementations to API.

These classes contains references to the queryable attributes of a entity.

This will avoid required internal dependency on console and rest api modules, which will be provided at runtime or assembly time.

lorthirk:
Hi all,

This PR refactors some `Predicate`s, still in internal modules, in the related API modules; moreover, where needed, some hardcoded Predicate reference were turned in appropriate `Predicate` member and some now unneeded dependencies were removed.

Fixes #1346


commons/src/main/java/org/eclipse/kapua/commons/configuration/ServiceConfigPredicates.java
```diff
@@ -9,16 +9,11 @@
  * Contributors:
  *     Eurotech - initial API and implementation
  *******************************************************************************/
-package org.eclipse.kapua.service.user.internal;
+package org.eclipse.kapua.commons.configuration;
 
 import org.eclipse.kapua.model.KapuaUpdatableEntityPredicates;
 
-public interface UserPredicates extends KapuaUpdatableEntityPredicates {
+public interface ServiceConfigPredicates extends KapuaUpdatableEntityPredicates {
```
Coduz: NAME predicate has been removed but this was not changed to `extends` from `KapuaNamedEntityPredicates`

Issue #1143, opened 2017-12-07
Bottom Window Can Cover Refresh Buttons in Data Tab


LeoNerdoG:
In Data tab (_**By Topic**_ sub-tab), user can lift bottom window up and therefore almost cover the _**refresh**_ buttons, _**Available Topics**_ and _**Available Metrics For Topic**_ windows (look at the screenshot). 

Testflow: 
1. Login as kapua-sys
2. Go to Data, try to lift bottom windows up as much as possible. 
**Kapua should not let user to lift bottom window so much that upper windows and buttons get covered.**

![screen shot 2017-12-07 at 13 15 14](https://user-images.githubusercontent.com/15121999/33715233-279e7ca6-db52-11e7-8247-ca9f5761dbe8.png)

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.39)

lorthirk:
This issues only occurs with a really short window. Can you confirm?

LeoNerdoG:
Hi @lorthirk ,
I have again tried with Chrome (63.0.3239.132), Firefox (57.0.4 (64-bit)), Safari (11.0.2) and Opera (50.0.2762.58) and it does not matter how big the browser window is (narrow, short, full-screen, windowed) - the issue persists.

lorthirk:
We should set a minimum height for the upper part

LeoNerdoG:
I have verified that this issue has bee resolved.

PR #1548, opened 2018-03-28
Data tab's bottom window fix

ct-ajovanovic:
Solves issue: Bottom Window Can Cover Refresh Buttons in Data Tab #1143 
- set maxSize for resultsLayout
- changed bottom margin for queryButtonLayout to prevent overlapping on small windows


LeoNerdoG:
I have verified that thisPR successfully fixes issue #1143 for every reasonable and useful window size.

muros-ct:
@lorthirk 
What do you think about this approach, using Window.getClientHeight()/2
It works until window is resized.
If we would use this, should we put it into a common helper method that would receive just percentage of window size?
Is there an option to call re-setting of max size each time Window gets resized?

muros-ct:
This solves single issue but not the overall problem with resizing browser to very small size.
It is still waiting for clever solution :)

PR #1560, opened 2018-04-04
Feature that checks size of browser window

muros-ct:
This feature is added to console when console is created. Listener is
registered that reacts on browser window resizing.
Minimum window width and height are set and if window is too small UI
is blocked with dialog that informs user of that issue.

This could solve multiple issues that are based on UI being to small to
work correctly or issues when layout of UI changes when browser is resized
and is smaller than minSize set for UI components.

Min size is now set to 900 x 530 based on empirical UI usage observations.
It could be made even less if scroll bars would be introducet to tabbed
panels.

### DIFF
--- a/console/core/src/main/java/org/eclipse/kapua/app/console/core/client/InfoPopup.java
+++ b/console/core/src/main/java/org/eclipse/kapua/app/console/core/client/InfoPopup.java
@@ -1,0 +1,54 @@
+/*******************************************************************************
+ * Copyright (c) 2018 Eurotech and/or its affiliates and others
+ *
+ * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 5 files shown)


muros-ct:
@lorthirk Leo checked it from functional standpoint and defined 900 x 530 min size for console.
We could make it smaller but in some cases it would still break usage of certain features.

What do you think about this approach to minimum requirements in general?

LeoNerdoG:
Hi @lorthirk @muros-ct,
as Uros already stated, I have checked this and from user standpoint it is OK, so this can be merged when ready.

muros-ct:
This is an example when window gets too small for normal usage of UI.

too_small

Issue #1534, opened 2018-03-21
Data export to CSV or Excel fails when query results more than 10000 rows


tovantti:
Related to #1155 that limit data in table view to 10000 rows, data export to CSV or Excel fails when query results more than 10000 rows.

Similar 10000 row hard limit doesn't exist in DataExporterServlet

Affected version 0.3.2

Should be fixed by #1378

lorthirk:
I write a proposal in #1566. While this is not a final solution, it should enable at least exporting in XLS or CSV queries with more than 10000 results. However, I may add a larger, configurable limit for the export before the PR gets merged.

LeoNerdoG:
Hi @tovantti,
as this has been fixed - cna you please close this issue? I have checked the PR and also the Develop branch and it is OK. Now user gets proper files without errors when there are more than 10k results.

Regards,

Leonardo

PR #1566, opened 2018-04-04
Fix Message Export with queries having more than 10000 results

lorthirk:
Hi all,

This PR aims at fixing #1534, so that the `DataExportServlet` doesn't crash anymore due to Elasticsearch result set limit to 10000 rows. However, in order to achieve this, we have to specify data sorting as Timestamp Ascendent, ignoring any other sort setting selected in the Data View.

While I was there I also fixed a couple of issues in the Data View (default sorting was not visible, sort by device id was not working) and refactored a bit exception displayed from REST APIs.


LeoNerdoG:
@lorthirk I have checked this PR and it solves issue #1534. It can be merged when ready.

PR #1570, opened 2018-04-05
Convert Account.childAccounts to List to keep children sorting

### LINKED ISSUE #866: Kapua Child Account List Sequence
When adding a lot of child accounts to certain parent account(e.g. more than 100 or 1000) the list of them is not in ascending/descending order of creation but rather in a numerical way, which can be confusing for user (look at the picture bellow). I have created just a few child accounts, but if there were more, it would get a lot harder to find the desired child account. 

TestFlow: 
1) Log in as kapua-sys
2) Create several child accounts (see picture bellow)
3) Go to Switch account option and try to switch account.
Look at the list of child accounts that Kapua GUI produces. 

screen shot 2017-10-16 at 08 11 38

Browsers used:

lorthirk:
Hi all,

this PR refactors `Account.getChildAccounts` to return a `List` instead of a `Set` so that Child Accounts are consistently returned when clicking on the Account change menu.

Fixes #866

### DIFF
--- a/console/module/account/src/main/java/org/eclipse/kapua/app/console/module/account/shared/model/GwtAccount.java
+++ b/console/module/account/src/main/java/org/eclipse/kapua/app/console/module/account/shared/model/GwtAccount.java
@@ -12,7 +12,7 @@
 package org.eclipse.kapua.app.console.module.account.shared.model;
 
 import java.io.Serializable;
-import java.util.Set;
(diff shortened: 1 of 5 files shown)


LeoNerdoG:
I have verified that this PR fixes inconsistency of Child account list which is one part of issue #866. It can be merged when ready.

Issue #1559, opened 2018-04-03
No "Granted/Assigned Devices" Tab In Tags


LeoNerdoG:
As there are _**Granted Users**_ tab in Roles, which contain all the users that certain role is added to and in _**Assigned Devices**_ in Access Groups, I think there should also be a _**Granted/Assigned Devices**_ tab in **_Tags_** which would contain all the devices that have a certain tag. 

![screen shot 2018-04-03 at 10 37 46](https://user-images.githubusercontent.com/15121999/38238706-0c8d52ea-372b-11e8-99a2-69c451e754c9.png)

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0 (64-bit))
- Opera (Version 49.0.2725.39)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1571, opened 2018-04-09
Adding Assigned Devices Tab to Tags

muros-ct:
New implementation of Tab for devices that have certain Tag assigned.
Similar to implementation of Device Groups and Users in group.


LeoNerdoG:
I have verified that this PR successfully resolves issue #1559.

PR #1576, opened 2018-04-10
Added reset job target step index

### LINKED ISSUE #1568: Implement a feature to reset the JobTarget status before processing
This issue track the feature of resetting the status of a JobTarget to a defined `stepIndex` and from there restart the processing.

This feature is required when a sequence of step needs to be run from the beginning in order to succeed (i.e a command that opens a connection to a server).

This feature can be further improved with a "roll back" action for the target in case of an exception.

Coduz:
This PR closes #1568 .

I added a new field to the `JobStartOptions` to specify from which step the processing should start and made changes to the `DefaultTargetReader` to add `KapuaQueryPredicates` accordingly.

This PR is currently based on the PR #1569, so need to wait the #1569 to be merged into `develop` and then rebased onto `develop` excluding all commits but the last one.

### DIFF
--- a/job-engine/api/src/main/java/org/eclipse/kapua/job/engine/JobStartOptions.java
+++ b/job-engine/api/src/main/java/org/eclipse/kapua/job/engine/JobStartOptions.java
@@ -43,8 +43,7 @@ public interface JobStartOptions extends KapuaSerializable {
     @XmlTransient
     void addTargetIdToSublist(KapuaId targetId);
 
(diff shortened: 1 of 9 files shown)


Coduz:
While writing this feature I've thought that we may want want a functionality to allow some rollback operations when the step processing fails.

I'll open an issue on this, just to track this proposal

Coduz:
Rebased on `develop` after merging #1569 .

Issue #1572, opened 2018-04-09
Kapua Freezes if Tag Deleted From Device


LeoNerdoG:
Kapua freezes when users deletes a tag from device (look at the screenshot).

Testflow: 
1. Login as kapua-sys
2. Go to Tags, create a tag (e.g. tag1)
3. Go to devices, add a device (e.g. device1)
4. Select device1 and add tag1 to it. 
5. Delete the tag1 from device1.
**Kapua freeses when user deletes tag from device.**

<img width="1332" alt="screen shot 2018-04-09 at 14 17 50" src="https://user-images.githubusercontent.com/15121999/38497256-cc7d7baa-3c00-11e8-9227-a3d6792e9ebc.png">

Browsers used: 
- Google Chrome (Version 65.0.3325.162 (Official Build) (64-bit)
- Safari (Version 11.0.3 (13604.3.5))
- Mozilla Firefox (58.0.2  (64-bit))
- Opera (Version 51.0.2830.26)

LeoNerdoG:
As (strictly speaking) this issue has been resolved, I am closing it. For things that are not resolved new issue has been opened (#1586).

PR #1578, opened 2018-04-10
Kapua Freezes if Tag Deleted From Device fix

ct-ajovanovic:
Solves issue: #1572


LeoNerdoG:
@ct-ajovanovic,
I ahve tested this and found out that if you change tags in one device (e.g. device1) - adding or removing - and then want to do the same for another device (e.g. device2), Kapua does not react to these changes. User has to select another service (e.g. Roles, Tags, Users...) and then come back and change Tag parameters for Device2. 
But after the tag paramteres of device2 are changed, user cannot change device1 Tag parameters. 

Can you please check where the problem is? Because currently in Develop branch user cannot delete a tag, this cannot be reproduced there.

Coduz:
Sorry, mistakenly merged 😅

LeoNerdoG:
Ok, no problem, I have already opened a new issue, so I think this one can stay like this :)

PR #1580, opened 2018-04-10
Proposal for Entity Grid auto refresh

muros-ct:
This is an example of automatic refresh of Entity grids.
Example is based on DeviceGrid, where it is useful to have dynamic
refresh of device status.

Implementation is based on client side timer that executes pull for
data and checks if data that is displayed in grid has changed. If data
changed it issues refresh of grid, if not it stays as it is. This prevents
unnecessary blinking of UI.

Data change is checked on check of optlock fields sum of displayed records.

### DIFF
--- a/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/ui/grid/EntityGrid.java
+++ b/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/ui/grid/EntityGrid.java
@@ -54,7 +54,12 @@ public abstract class EntityGrid<M extends GwtEntityModel> extends ContentPanel 
     /**
      * Used for checks if data on grid query has changed from last refresh.
      */
-    protected int lastOptlockSum;
+    protected boolean lastChanged;
+    /**
+     * List of entity changed flags.
+     */
+    protected int[] lastChange = new int[ENTITY_PAGE_SIZE];
+
 
     private EntityCRUDToolbar<M> entityCRUDToolbar;
     private boolean entityGridConfigured;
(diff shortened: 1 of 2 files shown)


LeoNerdoG:
Hi @muros-ct,
I have started testing this and found a case that his does not work as it should. 
The tesflow is bellow: 
1. Login as kapua-sys
2. Go to Devices, add 100+ devices (so you have at least two pages of devices)
3. Go to page 1 and connect a live device with name, that is on the second page (e.g. device99)
**Regardless of which page we are on, Device view blinks, when connection status of the device changes - but GUI should only blink, if the change is on current page/view (so if we are on page 1 and the change has happened on page 2, there should be no blinking).**

LeoNerdoG:
@muros-ct,
I have found another issue. If user manually clicks _**Refresh**_ button before the device is _**automatically**_ refreshed, GUI gets refreshed every 15 seconds...
Testflow: 
1. Login as kapua-sys
2. Go to Devices, add a live Kura device to Kapua (e.g. device0)
3. Disconnect the device (so that Conncetion Status socket icon turns yellow).
4. In Kura, click _**Connect**_ and before socket icon turns green in Kapua, press _**Refresh**_ button (next to add, edit and delete buttons on top). 
_**Kapua will first manually refresh the page, after that the "automatic" refresh should happen. But after that, Kapua will not stop refreshing page every 15 seconds.**_

LeoNerdoG:
@muros-ct,
There is another problem with this automatic refresh.
If a user is changing configuration of (what seems like live device) but device disconnects in the proecss, all the changes are lost. Testflow is bellow.

1. Login as kapua-sys
2. Connect a live Kura device to Kapua (e.g. device0) 
3. Go to Kapua -> Devices, select device0 and go to its Configuration. 
4. Change some parameteres but don't save them
5. Go to Kura GUI and Disconnect the device from kapua
**After this, all unsaved changes will be lost, which is, from user standpoint, not ok.**

LeoNerdoG:
@muros-ct,
if there is a connected (live Kura) device in Kapua and if user deletes this device from Device list and after that adds it to the list again with the same name, this device's connection status is marked as "Unknown" instead Connected.
Testflow: 
1. Login as kapua-sys
2. Connect a live Kura device to Kapua (e.g. device0)
3. Go to Devices and delet device0
4. Add a new device named device0
**After the device0 is added, its conenction status in Device is set to "Unknown", although it is properly connected to Kapua.**

muros-ct:
I did couple of changes to code, so that refresh is working properly.
But... Refresh is still causing problems with editing records while being refreshed.
UI would need much deeper refactorig to accomplish real automatic refresh functionality.

Closing this PR, but leaving branch for future reference.

Issue #1276, opened 2018-01-11
Child Account Refresh Button Not Highlighted When Pressed


LeoNerdoG:
When user uses bottom refresh button in Child Account  (for refreshing certain child account settings) it does not stay highlighted after user presses it.

Testflow: 
1. Login as kapua-sys
2. Go to Child Accounts, create new child account (e.g. kapua-child1)
3. Select created child account and navigate to its Settings tab 
4. Select **Refresh** button (look at the screenshot) and press it multiple times.
**After first press, the refresh button is not highlighted any more.**

Screenshot No. 1 (When user navigates to **Refresh** button it gets highlighted):
![screen shot 2018-01-11 at 08 05 13](https://user-images.githubusercontent.com/15121999/34812631-d5e1911c-f6a6-11e7-9469-26eb6f70aaf3.png)

Screenshot No. 2 (After **Refresh** is pressed for the first time, it is not highlighted any more ):
![screen shot 2018-01-11 at 08 05 31](https://user-images.githubusercontent.com/15121999/34812636-d81b31fe-f6a6-11e7-8a3c-bca71c740756.png)

Screenshot No. 3 (How upper refresh button stays highlighted after first click):
![screen shot 2018-01-11 at 08 18 45](https://user-images.githubusercontent.com/15121999/34812901-0811c0e8-f6a8-11e7-8a66-c6c9c8ec41d1.png)

Browsers used:
- Google Chrome (Version 63.0.3239.132 (Official Build) (64-bit)))
- Safari (Version 11.0.2 (13604.3.5))
- Mozilla Firefox (57.0.4 (64-bit))
- Opera (Version 50.0.2762.45)

LeoNerdoG:
Actually this behaviour is as intended and should not be fixed. The real problematic behaiour is actually for those buttons that stay blue if they are not pressed. 
Bellow is the list of buttons and their behaiours. It is stated, which one should be fixed. 
@ct-ajovanovic has already started working on this.

Detailed description:
After user presses buttons listed bellow, they stay coloured (blue background). They should really behave as refresh button under Devices -> Events or Devices -> Bundles (blue background when they are pressed otherwise default barckground color).

problematic buttons (Connections):
 - Search and Reset buttons in Connections Filter

Problematic buttons (Devices):
 - Device Packages refresh buttonqw	
 - Submit button when adding tags (if parameters not OK, submit button stays coloured)
 - Export Data button when not clicked on triangle icon
 - Refresh button in Device Packages
[...]

PR #1604, opened 2018-04-17
Fixing button behaviour for webkit and gecko browsers - same look

muros-ct:
Changed behaviour of buttons so that they behave same on webkit based and
mozilla browser.

Buttons in mozilla are changed so that they don't have border when they are in
focus. Also mozilla buttons are now blue when in focus, same as with chrome
and other webkit browsers.

Filter button is now not blue when not selected.

This fixes issue #1276


LeoNerdoG:
@muros-ct,
I have found only one issue and that is with the **Reset** button in Devices -> Command. 
For now, this button does not stay coloured after press (as **Reset** button in filter panel), but looses focus. 
Everything else is OK.

LeoNerdoG:
I have verified that with latest changes issue #1276 is resolved with this PR.

Issue #1598, opened 2018-04-17
Problem With Adding Targets When Device Do Not Have Display Name


LeoNerdoG:
If devices have no **Display Name** parameter, there are only empty spaces when sorting by **Display Name** in Jobs -> Targets -> Add (look at the screenshot). 
This can simpy be avoided if **Display Name** parameter was mandatory field at device creation.

Testflow: 
1. Login as kapua-sys
2. Go to Child Accounts, create a child account (e.g. kapua-sys-master)
3. Set all the kapua-sys-master account settings to Infinite (Users, Tags, Groups, Devices...)
4. Within kapua-sys-master create a user (e.g. kapua-sys-master) and give it ALL:ALL:ALL permissions (with Forwardable set to True)
5. Login as kapua-sys-master user and go to Devices
6. Adde 10 or more devices (e.g. device0, device1, device2,...)
7. Go to Jobs, create a job (e.g. job1) and go to its Targets, click **Add**.
**When user clicks on **Dispaly Name** column, to sort by this parameter, Kapua succesfully sorts devices, but because the names are empty, it is not OK.**

Screenshot:
<img width="1438" alt="screen shot 2018-04-17 at 13 46 07" src="https://user-images.githubusercontent.com/15121999/38868651-0cba3382-4248-11e8-95d2-df3d61469726.png">

Browsers used: 
- Google Chrome (Version 65.0.3325.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (59.0.2  (64-bit))
- Opera (Version 52.0.2871.64)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1614, opened 2018-04-19
Set Device Display Name as a mandatory field at device creation

ct-ajovanovic:
Solves: Problem With Adding Targets When Device Do Not Have Display Name #1598


LeoNerdoG:
I have verified that this PR successfully resolves issue #1598. It can be merged when ready.

Issue #935, opened 2017-11-08
TimeStamp Time And DateRange Time Not Same


LeoNerdoG:
When setting DateRange in Data, Timestamp time is not the same as DateRange time, because Timestamp time is in Zulu time (look at the screenshot bellow). I know that Timestamp has "Z" in the end, but nevertheless this can be confusing for end users. 

Testflow: 
1. Login as kapua-sys
2. Connect a device with installed ExamplePublisher to Kapua
3. Go to Data and query the results. 

![screen shot 2017-11-08 at 09 06 13](https://user-images.githubusercontent.com/15121999/32538409-359f3d16-c466-11e7-819d-b75839cf0c0b.png)

Browsers used: 
- Google Chrome (Version 62.0.3202.75 (Official Build) (64-bit))
- Safari (Version 11.0.1 (12604.3.5.1.1))
- Mozilla Firefox (56.0.2 (64-bit))
- Opera (Version 48.0.2685.52)

ct-goranpalibrk:
@Coduz @lorthirk @muros-ct @LeoNerdoG In which way we could solve this, because we have several options: 
1) UTC/Zulu time (like we have in Results table in Data view)
2) Local time(as in Data Range)
Because this problem is not only here, should we implement this fix everywhere or in Data view only for now?

Coduz:
It would be possible to have it as `yyyy-mm-ddThh:mm:ss.sss+hh:mm` ? i.e. `2018-01-01T00:11:22+02:00`

LeoNerdoG:
I have verified that this issue has been resolved.

LeoNerdoG:
I have verified that this issue has been resolved.

Issue #49, opened 2016-10-24
Device event date bug


Coduz:
Hi all, 

while working on issue #47 I discovered that the broker deployend into develop vagrant machine stores device events two hours before the actual time. 

Then by looking at the SQL db table I saw that actually seems that the console stores events in UTC+2 rather than use UTC+0. 

We need to find from where this timezone issue generates.

Regards, 

_Alberto_

Coduz:
Hi all, 

after some debugging and try&error strategy with @lorthirk, we discover a funny thing that is what's causing the problem. 

For some (unknown) reasons, timestamp timezone does not work properly. 
Using console launched from Eclipse locally was causing all timestamps retrieved by the H2, that have not being inserted by the console itself, to be wrong. 
It has to be something related to the "?gwt.codesvr=127.0.0.1:9997" parameter that is used on debug mode. 
From the following screenshot it is clear that with the same Browser and same session there are different behaviour if that query parameter is used.

![screen shot 2016-10-26 at 5 42 28 pm](https://cloud.githubusercontent.com/assets/3942036/19735586/20589862-9ba4-11e6-9f71-b6750025059a.png)

I was not able to found an actual configuration/setup in order to have all situation working. The best set up that I managed to have is to have that query param removed and "-Duser.timezone=UTC" added to the launch configurations of console.

Anyway the deployment in demo and develop machine works fine.

I'll close the issue. If someone knows the reason please comment below 😄 

Regards, 

_Alberto_

elbert3:
Perhaps this explains it: http://stackoverflow.com/questions/22953276/gwt-behaves-differently-when-url-has-gwt-codesvr

Coduz:
Hi @elbert3 

Good spot on this one! 
I'll see if I can make some changes in order to always have dates working as expected.

Thanks!

_Alberto_

stefanomorson:
@Coduz : can you update the status of this issue ?

lorthirk:
I think we should remove this issue from the milestone, since it's not strictly related to development but it's rather something about the debug environment. What do you think?

Coduz:
I think that this issue can be closed. 
It is something that affects Console only when running from IDE.

PR #1636, opened 2018-04-20
Fix timezone in data view

ct-goranpalibrk:
Fix date-time format in all views and grids and Description tabs.

### DIFF
--- a/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/util/DateUtils.java
+++ b/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/util/DateUtils.java
@@ -13,6 +13,9 @@ package org.eclipse.kapua.app.console.module.api.client.util;
 
 import java.util.Date;
 
+import org.eclipse.kapua.app.console.module.api.client.messages.ConsoleMessages;
+
+import com.google.gwt.core.client.GWT;
 import com.google.gwt.i18n.client.DateTimeFormat;
 
 public class DateUtils {
@@ -20,7 +23,7 @@ public class DateUtils {
 //    private static final ConsoleMessages MSGS = GWT.create(ConsoleMessages.class);
 
     private static final String NONE = "None";
-    private static final String FORMAT = "dd MMM yyyy HH:mm:ss ZZZZ";
+    private static final ConsoleMessages MSGS = GWT.create(ConsoleMessages.class);
 
     private DateUtils() {
     }
@@ -33,7 +36,7 @@ public class DateUtils {
             return NONE;
         }
 
-        return DateTimeFormat.getFormat(FORMAT).format(d);
+        return DateTimeFormat.getFormat(MSGS.formatDateTime()).format(d);
     }
 
     public static int getYear(Date date) {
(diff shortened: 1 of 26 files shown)


lorthirk:
Can you please link relevant issues?

ct-goranpalibrk:
This PR is for issue #935

Coduz:
Uhm it seems that with this PR all the dates. Am I seeing right?

ct-goranpalibrk:
Yes you are right. But @LeoNerdoG  has something for discussion and he will write it in Monday morning..

LeoNerdoG:
Hi @ct-goranpalibrk @lorthirk @ct-goranpalibrk,
last week Goran and I had some problems with displaying correct time in the _**Description**_ tab. Where I had GMT time (instead of GMT + 02:00), Goran had the proper time and we could not figure out why (the only difference was that I was running Docker container, whereas Goran was not).
Today, when I again tested this branch, this issue does not manifest itself any more and all the other dates are also OK, so this PR can be merged, as it succesfully resolves issue #935.

lorthirk:
What environment was Goran using? If running from Eclipse, this could be related to #49.

Anyway, let me review the code before merging, since I expected a change only in Data view according to the issue, not in every view. While it's good to be consistent I would like to check that everything is in place first.

ct-goranpalibrk:
@lorthirk I am using eclipse and GWT plugin, and at my side everything was okay, and Leonardo says that everything is ok now also at his side, so I think that is no problems with environment.

lorthirk:
Yes, and I think it's because of the enforcing of the UTC time zone being removed.

ct-goranpalibrk:
Yes, I removed that. If that's not okay, I will revert that.

console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/util/DateUtils.java
```diff
@@ -13,15 +13,14 @@
...
 
-import com.google.gwt.i18n.client.TimeZone;
-
 import com.google.gwt.i18n.client.DateTimeFormat;
 
 public class DateUtils {
 
 //    private static final ConsoleMessages MSGS = GWT.create(ConsoleMessages.class);
 
     private static final String NONE = "None";
```
Coduz: Can you please localize this by using the Console Messages?

ct-goranpalibrk:
@lorthirk About your proposal for changing format in `EEE dd MMM yyyy HH:mm:ss Z`, that format will produce result like this 
onez

I think that is little bit more user-friendly this format `EEE dd MMM yyyy HH:mm:ss ZZZZ` and it displays time like this: 
fourz
But if you don't thinks so, I will keep this your suggestion.

lorthirk:
I think we can avoid the `GMT`. It doesn't add any valuable information to the timezone indication, and there are already a lot of informations packed in a small string, so sparing some of them should lead to a more clear information. Please use `Z`.

lorthirk:
After a quick chat with @LeoNerdoG we agreed to use `EEE dd MMM yyyy HH:mm:ss ZZZZ` as the format.

ct-goranpalibrk:
Format EE dd MMM yyyy HH:mm:ss ZZZZ is now added.

console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/util/DateUtils.java
lorthirk: @ct-goranpalibrk,

@Coduz meant to localize line 22 (`NONE`), not 23 (`FORMAT`). I fixed this in 551cde5.

LeoNerdoG:
@lorthirk I have verified that this PR successfully resolves issue with time zones and all the changes, that have been made, are OK. This can be merged when ready.

Issue #1601, opened 2018-04-17
Save And Discard Buttons in Settings Although Nothing Can be Changed or Saved


LeoNerdoG:
In Settings Tab (on main menu on the left) in Account Settings there are **Save** and **Discard** buttons, that should not be there, as user cannot change its settings (this can only do his parent). 
I think, they should be removed. 

Testflow: 
1. Login as kapua-sys
2. Go to Settings -> Account Settings
**Observe the Save and Discard buttons - they should nt be there, as nothing can be changed from inside of an account.**

Screenshot: 
![screen shot 2018-04-17 at 14 36 52](https://user-images.githubusercontent.com/15121999/38869979-b5c161b8-424c-11e8-9431-752e0c96c9cb.png)

Browsers used: 
- Google Chrome (Version 65.0.3325.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (59.0.2  (64-bit))
- Opera (Version 52.0.2871.64)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1639, opened 2018-04-23
Added methods to remove unnecessary buttons on Settings tab

ct-ajovanovic:
Solves issue: Save And Discard Buttons in Settings Although Nothing Can be Changed or Saved #1601


LeoNerdoG:
I have verified that this PR successfully resolves issue #1601.

console/module/account/src/main/java/org/eclipse/kapua/app/console/module/account/client/AccountConfigComponents.java
```diff
@@ -94,6 +94,7 @@
     private AccountConfigPanel devConfPanel;
     private BorderLayoutData centerData;
     private AccountDetailsTabDescription accountDetailsTabDescription;
+    private SeparatorToolItem separatorToolItem = new SeparatorToolItem();
```
muros-ct: I don't like it here too much. But it is ok.

Issue #1621, opened 2018-04-20
No Device Status Field When Adding device


LeoNerdoG:
When user manually adds a device, there is no "Device status" field, although this field is at device "Edit". 

Testflow: 
1. Login as kapua-sys
2. Go to Devices, click "Add"
**For now there are:
 - Client ID
 - Display Name
 - Access Group
 - Custom Attributes (#1, #2, #3, #4, #5).
There should also be "Device Status" field as this field is in "Edit" window.**

Screenshot No. 1 ("Add" window **without** "Device status" field):
![screen shot 2018-04-20 at 10 08 33](https://user-images.githubusercontent.com/15121999/39038069-cef8335e-4482-11e8-8ea6-3e56171bfc73.png)

Screenshot No. 2 ("Edit" window **with** "Device status" field):
![screen shot 2018-04-20 at 10 08 44](https://user-images.githubusercontent.com/15121999/39038072-d04d7f02-4482-11e8-909e-98129e9c89dd.png)

Browsers used: 
- Google Chrome (Version 65.0.3325.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (59.0.2  (64-bit))
- Opera (Version 52.0.2871.64)

ct-ajovanovic:
Hi @lorthirk 
In **createBody()** method of **DeviceAddDialog.java**  class I found this: 
``` java
      // hide fields used for edit
        clientIdLabel.hide();
        statusCombo.hide();
```
So, the Device Status field is hidden in the add dialog? 
Is this on purpose? And should it be changed? Or not?

Thanks
_Aleksandra_

lorthirk:
I'd say that it shouldn't be hidden. Feel free to proceed with the implementation.

LeoNerdoG:
I have verified that this issue has been successfully resolved.

PR #1645, opened 2018-04-25
Enabled setting device status at device creation

ct-ajovanovic:
The device status is now a mandatory field at the DeviceAddDialog. It's set
using GwtDeviceCreator class, which now has additional deviceStatus
field. There was also a change on the service side in the createDevice
method of GwtDeviceServiceImpl class when setting the status. The
dialogDeviceAddInfoMessage was also updated.


LeoNerdoG:
I have verified that this PR successfully resolves issue #1621. @lorthirk can you please check and merge when all the tests complete?

Issue #1633, opened 2018-04-20
Filter Panel Names Not Unified


LeoNerdoG:
Filter Panel names should be unified as follows: Connections/Devices/Jobs/Tags/Users/Roles/Access Groups/Accounts Filter (singular form) or Connection/Device/Job/Tag/User/Role/Access Group/Account Filter (plural form), because now some are in sigular form (Conneciton, Device, Job,User, Role, Account) and all others in plural form. 

Also Account filter sohuld be renamed to Child Accounts Filter and Job filter to Batch Jobs filter (or Access Groups Filter should be changed to just Groups Filter). 

Testflow:
1. Login as kapua-sys
2. Go to Connections/Devices/Jobs/Tags/Users/Roles/Access Groups/Accounts and observe Filter panel names. 
**Observe the filter panel names and change them all to singular/plural form.**

Browsers used: 
- Google Chrome (Version 65.0.3325.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (59.0.2  (64-bit))
- Opera (Version 52.0.2871.64)

LeoNerdoG:
I have verified that this issue is resolved.

PR #1656, opened 2018-04-26
Re fix issue 1633

mgogic:
Fix for singular filter names
#1633


LeoNerdoG:
I have verified that this PR successfully resolves issue #1633 (Filter names are not in singular form). It can be merged when ready.

Issue #1629, opened 2018-04-20
Inconsistency in Description Tabs


LeoNerdoG:
There are several inconsistencies in **_Description_** tabs. For example, **Tags** and **Access Groups** have _Entity_ and _Tag_ / _Access Group_ fields, but Roles, Users and Jobs have only _Role_, _User_ and _Job_ fields. 

I propose that there is an _**Entity**_ field and item _**Name**_ field (as is currently in Tags and Groups) in every Description tab (with some Exceptions, such as Devices and Connections).

Also in Endpoints, there should be _**Entity**_ and _**Endpoint**_ and not _**Tag**_.

Testflow: 
1. Login as kapua-sys
2. Go to a service (Devices, Tags, Jobs,...) and create an item (so Description tab will be populated with data). 
3. Check every Description tab and observe similarities/differences (as described above). 

Screenshot No. 1 (Example of Role Description tab):
![screen shot 2018-04-20 at 12 56 41](https://user-images.githubusercontent.com/15121999/39047458-35e70218-449a-11e8-945d-6cefc058ed9f.png)

Screenshot No. 2 (Exaxmple of Access Groups Description tab):
![screen shot 2018-04-20 at 12 56 48](https://user-images.githubusercontent.com/15121999/39047462-37f686b4-449a-11e8-9ddc-936c49ebfb20.png)

Browsers used: 
- Google Chrome (Version 65.0.3325.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (59.0.2  (64-bit))
- Opera (Version 52.0.2871.64)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1657, opened 2018-04-26
inconsistency in description tabs

ct-goranpalibrk:


LeoNerdoG:
@ct-goranpalibrk 
I have verified that this PR successfully resolves issue #1629. But I have found out that there is no consistency with _**Created On**_ and _**Created By**_ options. Can you please put Created On first and Created by second? 
@lorthirk is this OK - for Created On to be first? I haven't found any other requirements for this field.

Regards,
Leonardo

Issue #1652, opened 2018-04-26
No Job Name in Start Confirmation Title Window


LeoNerdoG:
When user starts a job, Kapua opens a confimration window, where user can Submit or Cancel his decision. But in title of this window there is not job's name, but some string (I guess from a DB). 
For end-user clarification there should be job's name in the title of this box. 

Testflow: 
1. Login as kapua-sys or any other user that has Device and Job permissions)
2. Go to Devices, add a device (e.g. device0)
3. Go to Jobs, create new job (e.g. job1) and add device0 as Target to this job
4. Go to Steps and add several Bundle Start/Bundle Stop steps (with Bundle ID that does not exist)
5. Press Start 
**After pressing the _Start_ button, Kapua will ask you, if you are sure you want to start the job, but in the title of the window is not job's name, but some random string (I gusee from DB).**

Screenshot:
![screen shot 2018-04-26 at 10 19 05](https://user-images.githubusercontent.com/15121999/39295198-0aef3aa0-493e-11e8-983d-a376434c9516.png)

Browsers used: 
- Google Chrome (Version 65.0.3325.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (59.0.2  (64-bit))
- Opera (Version 52.0.2871.64)

LeoNerdoG:
I have verified that this issue has been successfully resolved.

PR #1658, opened 2018-04-26
Fix issue 1652

mgogic:
#1652


LeoNerdoG:
I have verified that this PR successfully resolves issue #1652.

Issue #1648, opened 2018-04-25
Non Descriptive Error Messages in Jobs


LeoNerdoG:
If user wants to start (or stop) a job, that has no targets or no steps defined, Kapua returns an error (which is OK), but the message is quite non descriptive (it looks like a message direct from the database - look at the screenshots).

Testflow: 
1. Login as kapua-sys (or any other user that has Job and Device permissions)
2. Go to Jobs, create a job (e.g. job1)
3. Select job1 and add a target, but do not add any steps
3.1. Start job1 and observe the error message
4. Select job 1 and add a step to it, but do not add any targets 
4. Start job1 and observe the error message
**Error message should be more descriptive about the error.**

Screenshot (job without added step - same happens if job does not have target configured):
<img width="1440" alt="screen shot 2018-04-25 at 14 57 46" src="https://user-images.githubusercontent.com/15121999/39247238-22ef8b00-4899-11e8-8580-b8e516b416a0.png">

Browsers used: 
- Google Chrome (Version 65.0.3325.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (59.0.2  (64-bit))
- Opera (Version 52.0.2871.64)

LeoNerdoG:
I have verified that this issue has been successfully resolved.

PR #1674, opened 2018-05-09
More descriptive error message when starting Jobs

muros-ct:
By changing exception that is thrown when job doesn't have target or
steps, error message is more descriptive.

This fixes issue #1648


LeoNerdoG:
I have verified that this PR successfully resolves issue #1648.

Issue #1596, opened 2018-04-16
No EndpointInfo Tab Although User Has ALL:ALL:ALL Permissions


LeoNerdoG:
Although user has ALL:ALL:ALL and Forwardable set to True permissions, when logged in as this user, there is no EndpointInfo tab in the main menu. 

Testflow: 
1. Login as kapua-sys
2. Go to Child Accounts, create a child account (e.g. kapua-sys-master)
3. Set all the kapua-sys-master account settings to Infinite (Users, Tags, Groups, Devices...)
4. Within kapua-sys-master create a user (e.g. kapua-sys-master) and give it ALL:ALL:ALL permissions (with Forwardable set to True)
5. Login as kapua-sys-master user
**Look at the main menu on the left - there is no EndpointInfo tab, although user has ALL permissions.**

Browsers used: 
- Google Chrome (Version 65.0.3325.162 (Official Build) (64-bit)
- Safari (Version 11.0.3 (13604.3.5))
- Mozilla Firefox (58.0.2  (64-bit))
- Opera (Version 51.0.2830.26)

Coduz:
Hi @LeoNerdoG ,

this is by design. 

Permission to manage those endpoint is `endpoint_info:read:*:*` By using console you cannot specify `*` as the `targetScopeId` of the permission. 

Available endpoints are still available through the Settings view

Regards, 

_Alberto_

muros-ct:
@LeoNerdoG what is solution to this issue? Not and issue, won't fix?

LeoNerdoG:
@muros-ct I am testing it right now and communicationg with @Coduz so it will be closed soon. I think this is not an issue, but I am performing some additional tests.

PR #1681, opened 2018-05-10
Making Endpoints visible to subaccounts of kapua-sys

muros-ct:
Because permission *:*:ALL was required for Endpoints it didn't work
for subaccounts. Change was made to EndpointViewDescriptor to require
read instead of readAll.

This fixes issue #1596


muros-ct:
@Coduz can you check this is OK from usage standpoint.
Because now any user with Endpoint : read permissions can see endpoints of whole system.

LeoNerdoG:
@Coduz 
I have a question regarding Endpoints. This PR successfully solves problem with Endpoints not visible in child accounts, but buttons (add, edit, delete) are disabled. I have tried setting the user ALL:ALL:ALL permissions and Enpoints:read/write/delete but nothing works. 
So here is the question: Should "sub-user" be able to add/modify/delete endpoints or not?

muros-ct:
This is not an issue and is being closed.
Endpoints can be seen in Settings tab for all accounts and users.

Issue #1586, opened 2018-04-12
Problems With Adding And Deleting Tags From Devices


LeoNerdoG:
The problem here is, that if you remove tags in one device (e.g. device1), and then want toadd or remove tags for another device (e.g. device2), Kapua does not react to these changes. User has to select another service (e.g. Roles, Tags, Users...) and then come back and change Tag parameters for Device2. 
But after the tag paramteres of device2 are changed, user cannot change device1 Tag parameters. 
This is also mentioned in comment in PR #1578.

Testflow: 
1. Login as kapua-sys
2. Go to Tags, create a tag (e.g. tag1)
3. Go to devices, add two devices (e.g. device1 and device2)
4. Select device1 and add tag1 to it. 
5. Delete the tag1 from device1.
6. Select device2 and try to add a tag to it (or try to delete it, if there are any tags)
**At this point Kapua will not let user to edit device's Tags (look at the description above).**

Browsers used: 
- Google Chrome (Version 65.0.3325.162 (Official Build) (64-bit)
- Safari (Version 11.0.3 (13604.3.5))
- Mozilla Firefox (58.0.2  (64-bit))
- Opera (Version 51.0.2830.26)

PR #1683, opened 2018-05-10
Fix problem with adding and deleting tags

bojanutkovic:
Fix issue #1586


LeoNerdoG:
I have verified that this PR successfully resolves issue #1586. It can be merged when all tests complete.

Issue #1611, opened 2018-04-18
Item List Not Automatically Refreshed When Deleting All Items On One Page


LeoNerdoG:
If user, for example, has two pages of child accounts (e.g. 103 child accounts in total) and switches to page 2, deletes all the items on that page, the following happens: 
1. The paging toolbar shows "Page 1 of 1"
2. Buttons for switching pages are disabled
3. No items are displayed on the list

Testflow: 
1. Login as kapua-sys
2. Go to Child Accounts, create a child account (e.g. kapua-sys-master)
3. Set all the kapua-sys-master account settings to Infinite (Users, Tags, Groups, Devices...)
4. Within kapua-sys-master create a user (e.g. kapua-sys-master) and give it ALL:ALL:ALL permissions (with Forwardable set to True)
5. Login as kapua-sys-master, go to Child Accounts and addd 100+ child accounts (103 or 104 will be enaugh)
6. Switch to page 2 (two) and delete all the items on this page. 
**Result is as described above - when user deletes all the items on page 2, Kapua does not go to page 1 and item list is not refreshed.**

This happens in: 
 - Batch Jobs
 - Tags
 - Users
 - Roles
 - Access Groups
 - Child Accounts
 - Endpoints

Browsers used: 
- Google Chrome (Version 65.0.3325.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (59.0.2  (64-bit))
- Opera (Version 52.0.2871.64)

PR #1689, opened 2018-05-14
Grid refresh after delete

muros-ct:
Grid was not refreshing correctly if last item was deleted on last page
of grid. Problem was that count of items was not placed correctly on most
of grids except device grid.

All grids are now fixed.

This fixes issue #1611


muros-ct:
@Coduz fixed int handling and check isEmpty.
@LeoNerdoG please re-evaluate.

Issue #1595, opened 2018-04-16
User Cannot Edit Live/Offline/Missing Device's Parameters


LeoNerdoG:
If user has connected a live Kura device to Kapua and wants to change its e.g. Access Groups parameter, Kapua does not accept changes when clicking on _Submit_ button (button can be clicked but nothing happens).
this also happens if device is offline or missing, but not if it is manually added (Device connection status: Unknown). 

Testflow: 
1. Login as kapua-sys
2. Connect a live Kura device to Kapua (e.g. device0)
3. Go to Devices and select device0, click _Edit_ and try to change something (e.g. Custom Attributes, Groups, Device Status...)
**After the changes are made and user clicks on _Submit_ button, nothing happens. **

Browsers used: 
- Google Chrome (Version 65.0.3325.162 (Official Build) (64-bit)
- Safari (Version 11.0.3 (13604.3.5))
- Mozilla Firefox (58.0.2  (64-bit))
- Opera (Version 51.0.2830.26)

LeoNerdoG:
I have verified that this issue has been resolved.

LeoNerdoG:
After consultation with @lorthirk we should proceed as following: 
1.Device's _**Client ID**_ field should be at least one character long (and not three, as it has been for now)
2. Error message for empty field should be changed from _**"Name must be at least 3 characters..."**_ to _**"Name must be at least 1 character long..."**_.

PR #1693, opened 2018-05-15
Validation of client-id to one character

muros-ct:
Validation of client-id field was changed from 3 to one character.
Tooltip and validation message was also changed accordingly.

This fixes issue #1595


LeoNerdoG:
I have verified that this PR successfully resolves issue #1595.

Issue #1673, opened 2018-05-09
"Device Not Connected" Error If User has Only Device Permissions


LeoNerdoG:
If user has all the proper permissions for viewing device properties (Device:ALL, Device_event:ALL, Device_lifecycle:ALL, Device_Management:ALL) Kapua returns error as shown bellow on screenshot. 

Testflow: 
1. Login as kapua-sys 
2. Go to Child Accounts, create an account (e.g. kapua-child0)
3. Select kapua-child0 and add a user with all the permissions described above and Forwardable set to True (e.g. kapua-user0), also set the user Broker_connect permission
4. Login as kapua-user0 and go to Devices, connect a live Kura device to this user
**When trying to see device's options in Configuration or Assets, kapua returns an error as seen bellow. Also executing commands is not possible (but different error pops-up.**

Screenshot: 
<img width="1284" alt="screen shot 2018-05-09 at 09 27 01" src="https://user-images.githubusercontent.com/15121999/39801467-93ed9be4-536b-11e8-8863-8fad635342ee.png">

Browsers used: 
- Google Chrome (Version 66.0.3359.139 (Official Build) (64-bit))
- Safari (Version 11.1 (13605.1.33.1.2))
- Mozilla Firefox (59.0.2  (64-bit))
- Opera (Version 52.0.2871.99)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1694, opened 2018-05-15
Device Call account find wrapped in privileged action

muros-ct:
Device call actions require account id, this needs to be found by
account name. This requires account service to perform action, but user
otherwise doesn't need account permissions.

That is why this single account operation was wrapped in doPriviledged.

This solves issue #1673


muros-ct:
@Coduz can you please check if this is ok. It regards security priviledges, so it is allways better to check twice.

LeoNerdoG:
I have verified that this PR successfully resolves issue #1673. It can be merged.

Issue #1624, opened 2018-04-20
"Download" And "Rollback To" Buttons Not Disabled Although no Snapshot Selected


LeoNerdoG:
Although user has not selected any of the snapshots in Device -> Configuration -> Snapshots, the "Download" and "Rollback to" buttons are not disabled and can be clicked (when clicked, nothing happens). 
If, for example, user selects and then immediatelly deselects a snapshot - Download and Rollback buttons get disabled.

Testflow: 
1. Login as kapua-sys
2. Go to Devices, connect a live Kura device to Kapua (e.g. device0) 
3. Select device0 and go to Configuration -> Snapshots 
**Download and Rollback to buttons are not disabled although no snapshots are selected.**

Screenshot:
![screen shot 2018-04-20 at 11 38 16](https://user-images.githubusercontent.com/15121999/39044023-45dff310-448f-11e8-8d13-2fae65c19b0c.png)

Browsers used: 
- Google Chrome (Version 65.0.3325.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (59.0.2  (64-bit))
- Opera (Version 52.0.2871.64)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1696, opened 2018-05-15
Fixed Download and Rollback buttons not disabled althought no snpashot selected

mgogic:
Issue 1624
Signed-off-by: mgogic <milan.gogic@comtrade.com>


LeoNerdoG:
I have verified that this PR successfully resolves issue #1624. It can be merged when all tests complete.

console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/configuration/DeviceConfigSnapshots.java
```diff
@@ -378,6 +378,8 @@ public void refresh() {
             } else {
                 toolBar.enable();
                 refreshButton.enable();
+                downloadButton.setEnabled(false);
+                rollbackButton.setEnabled(false);
```
muros-ct: Why not using .disable method as everywhere else?

PR #1698, opened 2018-05-16
Fix sonar issues usage of map entry set

Coduz:
This PR fixes other issue reported by Sonar analisys.

More info at: https://sonar.eclipse.org/drilldown/issues/144827

Rules fixed are on commit message

### DIFF
--- a/commons/src/main/java/org/eclipse/kapua/commons/event/ServiceEventScope.java
+++ b/commons/src/main/java/org/eclipse/kapua/commons/event/ServiceEventScope.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017 Eurotech and/or its affiliates and others
+ * Copyright (c) 2017, 2018 Eurotech and/or its affiliates and others
  *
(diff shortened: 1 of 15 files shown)


commons/src/main/java/org/eclipse/kapua/commons/event/ServiceEventScope.java
```diff
@@ -91,6 +92,10 @@ public static void end() {
             eventContextThdLocal.set(null);
         }
 
-        eventStack.pop();
+        if (eventStack != null) {
+            eventStack.pop();
+        } else {
+            throw KapuaRuntimeException.internalError("Event stack should be 'null'");
```
riccardomodanese: shouldn't be 'null' :)

Coduz: Fixed 👍

Coduz:
After
screen shot 2018-05-17 at 9 30 36 am

Coduz:
Difference over the day:
screen shot 2018-05-17 at 9 30 36 am

Issue #1631, opened 2018-04-20
Kapua Reports Error in Access Group Field When Editing Device 


LeoNerdoG:
If user has several devices (e.g. 30) and starts to edit their parameters one by one, Kapua at some devices eventually (after 20+ consecutive edits) starts to show error by Acccess Group field - as there would be an empty field or some other error. 
This is not right, because when user creates (or connects) a device, it is automatically  added to a group, so when editing a device there shoud be no errors. 

Testflow: 
1. Login as kapua-sys
2. Go to Child Accounts, create a child account (e.g. kapua-sys-master)
3. Set all the kapua-sys-master account settings to Infinite (Users, Tags, Groups, Devices...)
4. Within kapua-sys-master create a user (e.g. kapua-sys-master) and give it ALL:ALL:ALL permissions (with Forwardable set to True)
5. Login as kapua-sys-master user and go to Devices
6. Add several devices (more than 20, e.g. device0, device1,...) and add them all to "No group" group. 
7. Go to Groups, create a group (e.g. group0)
8. Go back to devices, start changing the device Access Groups from "No group" to group0, one by one. 
**At some point, when editing a device, Kapua will report an error (red exclmation mark on the right side) saying, that there are errors in this field, although there aren't any.**

Screenshot:
![screen shot 2018-04-20 at 14 01 14](https://user-images.githubusercontent.com/15121999/39049790-34d67d78-44a3-11e8-8341-61da128658a0.png)

Browsers used: 
- Google Chrome (Version 65.0.3325.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (59.0.2  (64-bit))
- Opera (Version 52.0.2871.64)

muros-ct:
@LeoNerdoG Having trouble to reproduce. Is this still an Issue? I inspected the code that validates field of group and it only has validation of empty field that gets validated only when value is entered.
I tried not to fill this field as part of dialog display and it didn't report that error, the field was just empty.

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1700, opened 2018-05-16
Group field not validated correctly on edit

muros-ct:
This removes mandatory group field on edit dialog for device. This removes
condition that resulted in false negative validation of group field.
As group field is mandatory, that means that device can not be created
without group, it is at least No Group. That means that when user edits
device, it will always have group field set and it doesn't have to check
for empty field.
This should resolve this odd issue. I hope.

This solves issue #1631


Coduz:
Hi Uros. 
Isn't this caused by the fact that we are loading all the groups on DeviceAddDialog class asynchronously and in the DeviceEditDialog we are doing the same for one single group? So what's happening is that with the list of all groups is loaded after the `populateDialog(...)` does the `setValue`?

LeoNerdoG:
From end-user point of view, this PR successfully resolves issue #1631.

muros-ct:
@Coduz I tryed to mimic that behaviour, but I couldn't reproduce it. Because loading is done async but validator should work, as value is allready filled. I even tried not to fill value int group and it still didn't report validation error.

Issue #1638, opened 2018-04-23
No Tooltip For Long Tags in Devices -> Add Tag


LeoNerdoG:
If certain tag is longer than ~ 30 characters (depends on width of characters) then Tag name is not fully shown in **Devices -> Tags -> Add tag** (which is ok). The problem is if there are two similar tags, that are both long and differ only in the end part of tag. 
As for now, Kapua does not show full tag name, if mouse hovered over the shortened tag name but would be verey useful for end-user if it would.

Testflow: 
1. Login as kapua-sys
2. Go to Tags and add a tag with a very long name. 
3. Go to Device, create a device (e.g. device1) and the long tag previously created
**When user hovers over the tag with long name, only part of it is shown. It would be easier for user if a tooltip would show with full tag name (especially if there are a lot of similar long name tags)**

Screenshot: 
<img width="1127" alt="screen shot 2018-04-23 at 08 07 34" src="https://user-images.githubusercontent.com/15121999/39109442-448e50f2-46cd-11e8-9258-97d849844dfb.png">

Browsers used: 
- Google Chrome (Version 65.0.3325.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (59.0.2  (64-bit))
- Opera (Version 52.0.2871.64)

LeoNerdoG:
I have verified that this issue has been successfully resolved.

PR #1707, opened 2018-05-17
Added tooltip on combo box for adding tags to device

muros-ct:
Template for combo box item was added, and this template includes
title attribute that displays tooltip on selected drop down item.
This is issue, when name of item is clipped and it is hard for user
to distinguish items with similar begining.

This solves issue #1638


LeoNerdoG:
I have verified that this PR successfully resolves issue #1638. It can be merged when all tests complete.

Issue #1630, opened 2018-04-20
Unknown Command Error Should be Handled More Gracefully


LeoNerdoG:
If user enters random symbols into **_Execute_** field in Devices -> Command, Kapua returns an error (which is OK), but the text in error box is non-descriptive. 

Can we handle this kind of error more gracefully? 

Testflow: 
1. Login as kapua-sys
2. Connect a live Kura device to Kapua (e.g. device0)
3. Select device0 and go to Command and in _**Execute**_ field enter some random symbols (e.g. 'TYHJ'"YTGHBNı–”’Ó).
4. Press _**Execute**_ and observe the error box. 

Screenshot: 
![screen shot 2018-04-20 at 13 44 56](https://user-images.githubusercontent.com/15121999/39049331-5c712f60-44a1-11e8-8af8-caf63f86d0c0.png)

Browsers used: 
- Google Chrome (Version 65.0.3325.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (59.0.2  (64-bit))
- Opera (Version 52.0.2871.64)

LeoNerdoG:
I have verified that this issue has been successfully resolved.

PR #1708, opened 2018-05-17
Trimmed error output when executing command on kura device

muros-ct:
Error is returned in block of html code. This html was already parsed
to certain degree. I added removal of stack trace, so that only error is
reported to user.

I also changed header of error message from file upload error to command
execution error. I expect this was copy/paste error.

This solves issue #1630


LeoNerdoG:
I have verified that this PR successfully resolves issue #1630. It can be merged when all tests complete.

Issue #1600, opened 2018-04-17
User Display Name Not Automatically Refreshed


LeoNerdoG:
When user **Display Name** is changed, it is not automatically changed in the upper right corner (**username**@account). 

Testflow: 
1. Login as kapua-sys
2. Go to Child Accounts, create a child account (e.g. kapua-sys-master)
3. Set all the kapua-sys-master account settings to Infinite (Users, Tags, Groups, Devices...)
4. Within kapua-sys-master create a user (e.g. kapua-sys-master) and give it ALL:ALL:ALL permissions (with Forwardable set to True)
5. Login as kapua-sys-master, go to Users and change its Display Name
**Look at upper rgith corner, where **Display Name** of a user is shown - it is not changed until browsers refresh button is pressed.**

Browsers used: 
- Google Chrome (Version 65.0.3325.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (59.0.2  (64-bit))
- Opera (Version 52.0.2871.64)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1711, opened 2018-05-17
Update user display name automatically

bojanutkovic:
#1600 

### DIFF
--- a/console/core/src/main/java/org/eclipse/kapua/app/console/core/client/NorthView.java
+++ b/console/core/src/main/java/org/eclipse/kapua/app/console/core/client/NorthView.java
@@ -121,6 +121,10 @@ public class NorthView extends LayoutContainer {
                 new TableData(Style.HorizontalAlignment.RIGHT,
                         Style.VerticalAlignment.MIDDLE));
 
+        ReceiveMessageHandler receiveMessageHandler = new ReceiveMessageHandler();
+        receiveMessageHandler.getChecker().addMessageReceivedEventHandler(receiveMessageHandler);
+        receiveMessageHandler.setView(this);
+
         add(panel);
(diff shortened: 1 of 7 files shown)


LeoNerdoG:
Hi @bojanutkovic,
I was unable to test this, as there was an issue in develop branch, "preventing" end-user to change any of the parameters of the kapua user (Kapua reports error: "An illegal null value was provided for the argument user.name"), and becasue you did not rebase to latests develop, the issue persists in this PR.
Can you please reabase to develop and push changes again?

bojanutkovic:
@LeoNerdoG  
Sorry about that, please try again.

LeoNerdoG:
@bojanutkovic 
I have found an issue with your solution. It looks like Kapua changes the displayed name in top right corner even if you change display name of user, that is not logged in (example: I am logged in as kapua-sys, but I change display name of kapau-broker user. When I click Submit, the diplayed name in the top right corner will change, although it should not, because I am not logged in as kapua-broker). 

Can you please check this? 
Here is a screenshot. The displayed name should be "3" as I was logged in as kapua-sys, but it is "5" (the displayed name I last changed):
screen shot 2018-05-18 at 10 45 33

muros-ct:
@lorthirk can you please check this implementation of events and their handling.
@LeoNerdoG is checking UX.

LeoNerdoG:
After the last commit everything is OK on the GUI side. After @lorthirk confirms that implementation is good, it can be merged.

Coduz:
This PR will be closed.

we don't want to commit to the usage of events in GWT right now. Although this is a good idea, it would be better to have a plan of migration to use GWT events in all views/components.

Regards, 

_Alberto_

Issue #1628, opened 2018-04-20
Page Toolbar in Data View Gets Grayed Out if Query Clicked Multiple Times


LeoNerdoG:
If user clicks **Query** button multiple times (to refresh the data results list) paging toolbar gets grayed out (looks like it is disabled, but buttons are functional). 

Edit: 
Also the paging toolbar in Devices -> Events is grayed out by default.

Testflow:
1. Login as kapua-sys
2. Connect a live Kura device to Kapua (e.g. device1) with installed Example Publisher Bundle
3. In Kapua, go to Data (By Topic or By Device), selected device1, some metrics and click **Query** multiple times (press and hold Enter key)
**Kapua will properly refresh the Data list, but after that the paging toolbar buttons will get greyed out (disabled - unclickable).**

Screenshot No. 1 (Paging toolbar before multiple clicks):
![screen shot 2018-04-20 at 12 18 56](https://user-images.githubusercontent.com/15121999/39046042-0b14a6b2-4495-11e8-8c7b-26fbe870e783.png)

Screenshot No. 2 (Paging toolbar after multiple clicks):
![screen shot 2018-04-20 at 12 19 12](https://user-images.githubusercontent.com/15121999/39046045-0da28ebc-4495-11e8-9de4-dcfa270d26d9.png)

Browsers used: 
- Google Chrome (Version 65.0.3325.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (59.0.2  (64-bit))
- Opera (Version 52.0.2871.64)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1716, opened 2018-05-18
Query button is disabled during query operation

muros-ct:
Issue was showing if Query button was pressed so fast that query operation
didn't finish before next was triggered. That confused paging toolbar.
Solution was to disable button during query operation, so that user is
not able to trigger too many queries.

This solves issue #1628


LeoNerdoG:
I have verified that this PR successfully resolves issue #1628. It can be merged when all tests complete.

Issue #1634, opened 2018-04-20
Timeout Field in Job Steps Error Message Contains Invalid Value


LeoNerdoG:
Error message in Job Timeout Field in Steps say, that maximum value that can be entered is 9223372036854776000, but actual max value is 9223372036854776832. 
Error message hase to be changed. 

Testflow: 
1. Login as kapua-sys
2. Go to Batch Jobs and add a job (e.g. job1)
3. Select job1 and add a step with timeout value 9223372036854776832. 
**Kapua does not report any error, although value is too large.**

Screenshot No. 1 (Timetout value that definetelly exceeds limit):
![screen shot 2018-04-20 at 14 41 29](https://user-images.githubusercontent.com/15121999/39051389-d0227a52-44a8-11e8-9033-b5730dac6442.png)

Screensoht No. 2 (kapua should return an error for this value but it does not):
![screen shot 2018-04-20 at 14 41 35](https://user-images.githubusercontent.com/15121999/39051393-d2324e58-44a8-11e8-8fd4-bc5bdd956f4c.png)

Browsers used: 
- Google Chrome (Version 65.0.3325.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (59.0.2  (64-bit))
- Opera (Version 52.0.2871.64)

ct-ajovanovic:
Hi @Coduz @lorthirk 
Can you take a look at this?
It seems that the **maxValue** for this field is not rendered properly. It's set in the _JobStepAddDialog_ class like this:
```  
              numberField.setData(PROPERTY_NAME, property.getPropertyName());
                if (propertyType.equals(Long.class.getName())) {
                    numberField.setMaxValue(Long.MAX_VALUE);
                } else if (propertyType.equals(Integer.class.getName())) {
                    numberField.setMaxValue(Integer.MAX_VALUE);
                } else if (propertyType.equals(Float.class.getName())) {
                    numberField.setMaxValue(Float.MAX_VALUE);
                } else if (propertyType.equals(Double.class.getName())) {
                    numberField.setMaxValue(Double.MAX_VALUE);
                }
```
In the text of the issue we can see the problem when the Long.MAX_VALUE value is used.  It's true value is  922337203685477**5807**, in the error message its rounded to 922337203685477**6000**, but the actual rendered max value is 922337203685477**6832**. 
I found that similar thing happens with Integer.MAX_VALUE as well. Float and Double values should be checked as well.

[...]

PR #1723, opened 2018-05-21
Long (integer) value validation in NumberField

muros-ct:
While validating Long that is 64 bit values for integers there is wrong
validation of max value. Value is not validated correctly because
value is converted to Double before validating. This problem occurs because
Double can have mantisa which is 2^52 (signed). This brings validation to
that value.
To solve issue, max value is not set to Long.MAX_VALUE but to
4.503.599.627.370.496. If this is not good enough, NumberField provided
by GXT can not be used, as its validator uses Double for storing its values.

Other types should function correctly, as there is no limitation with conversion
to Double.

This solves issue #1634


LeoNerdoG:
I have verified that this PR successfully solves issue #1634. It can be merged when all tests pass.

Issue #1219, opened 2017-12-19
Generic Driver Name in Device Configuration


LeoNerdoG:
If user adds a drivers to the Kura device (e.g. driver1, driver2, driver3...), they are all named DummyDriver (as this is the name of the package, from which they are created) which is very confusing for end user. 

Testflow: 
1. Login as kapua-sys
2. Go to Devices, add a live Kura device to Kapua which has several drivers (e.g. DummyDrivers named driver1, driver2, driver3,...)
3. Select the connected device and go to its Configuration. 
**Kapua shows that device has a lot of _DummyDrivers_ (which is a generic name of the package) instead of the _ServicePID_ name, which would be much more suitable.**

Screenshot No. 1 (Dummy driver in Kura - its name is driver123):
![screen shot 2017-12-19 at 10 37 21](https://user-images.githubusercontent.com/15121999/34150828-dbafece2-e4a9-11e7-82eb-9b32e5aeaf1b.png)

Screenshot No. 2 (In Kapua, generic name of the package is displayed instead of ServicePID name):
![screen shot 2017-12-19 at 10 37 02](https://user-images.githubusercontent.com/15121999/34150846-e1546a24-e4a9-11e7-8231-412c5235dfa5.png)

Browsers used:
- Google Chrome (Version 62.0.3202.94 (Official Build) (64-bit))
- Safari (Version 11.0.1 (13604.3.5))
- Mozilla Firefox (57.0.1 (64-bit))
- Opera (Version 49.0.2725.39)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1726, opened 2018-05-21
Generic Driver Name in Device Configuration fix

ct-ajovanovic:
Changed the findDeviceConfigurations() method so that ServicePID is
shown instead of generic name for drivers and assets.

### DIFF
--- a/console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/server/GwtDeviceManagementServiceImpl.java
+++ b/console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/server/GwtDeviceManagementServiceImpl.java
@@ -298,8 +298,7 @@ public class GwtDeviceManagementServiceImpl extends KapuaRemoteServiceServlet im
                         gwtConfig.setId(config.getId());
                         if(config.getId().indexOf('.') == -1){
                             gwtConfig.setName(config.getId());
-                        }
-                        else{
+                        } else {
                             gwtConfig.setName(ocd.getName());
                         }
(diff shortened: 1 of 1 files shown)


ct-ajovanovic:
Solves issue: #1219

LeoNerdoG:
I have verified that this PR successfully fixes issue #1219. It can be merged.

console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/server/GwtDeviceManagementServiceImpl.java
```diff
@@ -296,7 +296,12 @@ public int compare(DeviceComponentConfiguration arg0, DeviceComponentConfigurati
                     if (ocd != null) {
                         GwtConfigComponent gwtConfig = new GwtConfigComponent();
                         gwtConfig.setId(config.getId());
-                        gwtConfig.setName(ocd.getName());
+                        if(config.getId().indexOf('.') == -1){
```
Coduz: Just a small suggestion, no change required :) 

Instead of using `aString.indexOf('x') == 1` you can use `aString.contains('x')` which is implemented using the `indexOf(...)` (e.g. ` return indexOf(s.toString()) > -1`). 

The benefit is that the code is slightly more readable :)

ct-ajovanovic: I will keep that in mind. Thanks for the suggestion!

Issue #1725, opened 2018-05-21
Job Step Window Freezes if User Edits Job Step


LeoNerdoG:
If user adds step to a job and afterwards wants to edit it, Kapua Step Window is not accessible (look at the screenshot).

Testflow:
1. Login as kapua-sys or any other user with properpermissions
2. Go to Jobs, add a job (e.g. job1)
3. Select job1 and go to its Steps, add a step (e.g. step1)
4. After adding step1, select it and click Edit.
**After doing this, Kapua freezes (look at the screenshot).**

Screenshot:
![screen shot 2018-05-21 at 15 17 33](https://user-images.githubusercontent.com/15121999/40309521-1801d2cc-5d0a-11e8-873f-3fec639e533c.png)

Browsers used: 
- Google Chrome (Version 65.0.3325.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (59.0.2  (64-bit))
- Opera (Version 52.0.2871.64)

Coduz:
I have some ideas about the problem on this. 

I've made a change where I removed the calculation of the next step index for a new step. 
Now the JobStepService supports automatic generation if no `stepIndex` is given. 

At the time I tested Add button but never tried Edit button.

My change is likely to have caused this error.

Sorry 😅

LeoNerdoG:
@Coduz 
Ok, no problem :) 
@muros-ct is fixing this right now, so this should be resolved quick.

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1730, opened 2018-05-22
Hidden problem in NumberField

muros-ct:
When adding steps to job, there is list of dynamic Job step properties.
In case that property is of type Long, NumberField is used for entry
field e.g. Timeout, display of that value didn't work, thou in DB it was
correct.
While inspecting code it might never worked at all.
I did change for Long, but other changes might be needed for other types,
if this was not related only to NumberField component.

This solves issue #1725


LeoNerdoG:
@muros-ct 
I have checked this PR and unfortunatelly the problem persists. Can you please recheck?

LeoNerdoG:
I have vrified that after last commit everyithing is OK and issue #1725 is successfully resolved. This PR can be merged when all tests complete.

Issue #1713, opened 2018-05-18
Kapua Returns an Error in Settings, if "+" Sign in Domain Name Field


LeoNerdoG:
If user enters at least one "+" sign in Endpoint Domain name (here Kapua does not report any problems), Kapua reports an error, when user navigates to Settings on the left side. 

Testflow: 
1. Login as kapua-sys
2. Go to Endpoints
3. Add a Endpoint with "+" sign in its Domain name (e.g. "super+mario.com")
4. Click Submit and then go to Settings. 
**At this point Kapua returns an error as shown on screenshot bellow.**

![screen shot 2018-05-17 at 14 56 05](https://user-images.githubusercontent.com/15121999/40179276-39d4f7c2-59e4-11e8-8ee3-bc925825211c.png)

Browsers used: 
- Google Chrome (Version 66.0.3359.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (60.0.1  (64-bit))
- Opera (Version 53.0.2907.57)

LeoNerdoG:
@Coduz 
As you can see from @ct-ajovanovic commit above, she removed the plus sign from allowed characters. 
I have reviewed section 3.2.2 of [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt), section 3.5 of [RFC 1034](https://tools.ietf.org/html/rfc1034) and section 2.1 of [RFC 1123](https://tools.ietf.org/pdf/rfc1123.pdf) and I did not found that plus sign is forbidden character, but not preferred / not assumed  for host names (the preferred ones are digits 0-9, characters a-z and A-Z (upper case "ignored") and also hyphen ("-")). 

So my question is if this is a valid solution? As I have never noticed plus signs in URLs I think this is OK. 

Also when looking at the **REGEX** `GwtEndpointValidationRegex.java`, I have noticed that:
1. User can enter multiple "." (dot signs) in the **Domain Name** field and the result is the same as described in original issue.
2. If user enters minus sign on the end (e.g. "abc-.com") same thing happens as described in original issue.
3. User can enter only "abc." and Kapua does not report any kind of error, but it should? As this is not a valid address?
@ct-ajovanovic can you fix this with next commit?

LeoNerdoG:
I have verified that this issue has been resolved.

Coduz:
Just saw this issue and the merged PR. 
Now is not possibile to use `localhost` or use any direct IP. 
[...]

PR #1750, opened 2018-05-29
Changed the URI_DNS regex

ct-ajovanovic:
Solves issue: #1713
Changed the Domain name regex in the GwtEndpointValidationRegex class.


LeoNerdoG:
I have verified that this PR successfully resolves issue #1713. @Coduz can you please check if REGEX is OK? @ct-ajovanovic and I have reviewed the documents explaining how this field should look like and from user standpoint it looks ok. As I see it, this can be merged when all tests pass.

muros-ct:
Great work on that RegEx.

Coduz:
Just saw this issue and the merged PR. 
Now is not possibile to use `localhost` or use any IP as `DomainName`. 
This is preventing any test on a local machine.

Issue #1751, opened 2018-05-29
Sorting by Display Name For Devices is Not Working


LeoNerdoG:
Sorting by **Display Name** for Devices does not work (see list bellow).

Devices -> Display Name
Jobs -> Targets -> Selected Only
Tags -> Assigned Devices 
Access Groups -> Assigned Devices 

I will recheck Kapua for this kind of error and later add more issues into this, if I find them. 

**Testflow:**
1. Login as kapua-sys or any other user with proper priviliges
2. Go to Tags and add a tag (e.g. tag1)
3. Go to Access Groups and add a group (e.g. group1)
4. Go to Devices, add several devices with not so similar names (e.g. "kkj090d9c", "nmncmnc",...)
5. Add tag1 to each device and also add all devices to group1
**When all the devices are added, try to sort them by Display Name. Also go to Tags/Groups -> Assigned Devices and try to sort them there by Display Name.**

**Expected behavior:**
 - Devices should sort by Device Display Name, as "Display Name" column can be clicked.

**Screenshot:**
![screen shot 2018-05-29 at 10 18 06](https://user-images.githubusercontent.com/15121999/40647441-28354b60-632c-11e8-8d24-98ef1eb22b3f.png)

**Version of Kapua**
 - 1.0.0 SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Openshift (in its variants) 
[ ] Others

**Main component affected**
[x] Console 
[ ] REST API
[ ] Message Broker
[ ] - Others

**Browsers used:** 
- Google Chrome (Version 66.0.3359.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (60.0.1  (64-bit))
- Opera (Version 52.0.2871.64)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1752, opened 2018-05-29
fix sorting on Display name in Devices

ct-goranpalibrk:

Added null check for sort fields and removed some part of code because this parts with sort fields are enough and because in all other converters is done like that

This PR fixes issue #1751 

Like i Wrote I have inserted some null checks an in them I gave values to sort fields like in all other converters and after that  deleted some part of code which was unnecessary because this my piece of code works all about sorting.


LeoNerdoG:
I hav everified that this PR successfully resolves issue #1751. It can be merged when all tests complete.

Issue #1724, opened 2018-05-21
Kapua Turns to Page One When Adding or Deleting an Item


LeoNerdoG:
If user is on page 2 or higher and tries to add, edit or delete an item, Kapua succesfully completes the job, but turns to page one of that list (which is very inconvenient).

Testflow: 
1. Login as kapua-sys or any other user with proper permissions
2. Go to Devices/Batch Jobs/Tags/Users/Roles/Access Groups/Child Aaccounts/Endpoints
3. Add 100+ items, so you have more than one page of items
4. Turn to page two or higher and try to add, edit or delete an item
**Kapua successfully completes the operation, but changes to page one.**

Browsers used: 
- Google Chrome (Version 66.0.3359.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (60.0.1  (64-bit))
- Opera (Version 53.0.2907.57)

LeoNerdoG:
I have verified that this issue has been successfully resolved.

PR #1755, opened 2018-05-30
Changed EntityGrid's refresh() method

ct-ajovanovic:
Solves issue: #1724
Changed the way entityLoader's properties (offset and limit) are set in the refresh method. They are reused after refresh.


LeoNerdoG:
I have verified that this PR successfully resolves issue #1724. It can be merged when all tests complete.

ct-ajovanovic:
Hi @Coduz 
I think it would be the best for _setReuseLoadConfig(true)_  property to stay in the refresh() method.  If this property is set after the loaders initialization, it will possibly cause problems with refresh(query) method where offset and limit for the loader should be set differently.

Issue #1597, opened 2018-04-17
Targets Deselected When Pressing Refresh, ClientID or Display Name


LeoNerdoG:
When user is adding targets to certain Job - if the **Refresh** button is pressed or column header DeviceID or Display Name, all the selected targets are deselected and only last one remains selected. 

Testflow: 
 1. Login as kapua-sys
 2. Go to Devices and add some devices (e.g. dev0, dev1, dev2,...)
 3. Go to Jobs and create a ob (e.g. job1) 
 4. Select job1 and add targets. 
 **When target dialogue box is open, select some devices and after that press _Refresh_ button, or try to sort Targets by Device ID or Display Name - kapua will deselect all except last selected target.**

Screenshot No. 1 (Selecting devices): 
<img width="1440" alt="screen shot 2018-04-17 at 13 34 47" src="https://user-images.githubusercontent.com/15121999/38867644-e0e41d16-4244-11e8-8efe-66052ec52f17.png">

Screenshot No. 2 (Situation after clicking Refresh, DeviceID or DisplayName):
<img width="1440" alt="screen shot 2018-04-17 at 13 34 47" src="https://user-images.githubusercontent.com/15121999/38867641-df6a0400-4244-11e8-9c4c-61c46332fe01.png">

Browsers used: 
- Google Chrome (Version 65.0.3325.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (59.0.2  (64-bit))
- Opera (Version 52.0.2871.64)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1758, opened 2018-05-30
fixes all problems with selecting and deselecting targets on refresh …

ct-goranpalibrk:
…and sorting

Brief description of the PR.
Added some new methods for Refresh button with another listener(also new made) in constructor, then hide that button in all other views, also added some if's is GwtKapuaDeviceModelConverter for sorting.

**Related Issue**
This PR fixes issue #1597 

New Refresh button which not only refresh grid, also deselect all checked items if they are checked. Also when User wants to sort columns, all selected items become unchecked.

**Screenshots**
No screenshots.

No description.


LeoNerdoG:
I have verified that this PR successfully resolves issue 1597. It can be merged.

Issue #1756, opened 2018-05-30
Quatation Marks Problem In Jobs 


LeoNerdoG:
This issue is connected to #1717.
If user enters quatation marks in Job Display Name or Device Display Name, every time Job Name field is modified, it gets more "& amp;quot;" for double quatation marks and "& amp;#39;" for single quatation marks. 
For Device Display Name this manifests indirectly - through Job Targets. If user creates a device with display name "!"%&%&"%"""""'(('&!", Kapua will show this as described above, but only in Job -> Targets -> Display Name).

**Testflow:**
1. Login as kapua-sys or any other user with proper priviliges (Users:ALL, Jobs:ALL)
2. Go to Devices, add a device with quatation marks in Display Name.
3. Go to Jobs an add a job (e.g. job1)
4. Select job1 and add a step with quatation marks in its Name (e.g. """"). After that select this step and click Edit - Submit - Edit - Submit. 
5. Select job1 and go to Targets. Select "Selected Only" and observe Device Display Names. Instead of quotes there are &quot;&...
**Observe, how every time the job Step Nme fields get longer because of additional symbols entered automatically by Kapua and how Display Name of a device is displayed in Devices and how in Jobs -> Targets.**

**Expected behavior**
job step name and device display name should be displayed "as-is" or quatation amrsk should not be allowed. 

**Screenshot No.1 (Quatation marks problem in Job Targets):**
![screen shot 2018-05-30 at 13 50 44](https://user-images.githubusercontent.com/15121999/40720061-cc7f8bf6-6415-11e8-964e-ed8f0e7bbd3e.png)

**Screenshot No. 2 (Quatation mark problem in Job Steps):**
![screen shot 2018-05-30 at 13 50 34](https://user-images.githubusercontent.com/15121999/40720059-cadf9caa-6415-11e8-937c-3457b79ef482.png)

**Version of Kapua**
 - 1.0.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Openshift (in its variants) 
[ ] Others

**Main component affected**
[x] Console 
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 66.0.3359.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (60.0.1  (64-bit))
- Opera (Version 53.0.2907.57)

LeoNerdoG:
[...]

PR #1762, opened 2018-06-04
Special characters problem in Jobs fix

ct-ajovanovic:
Solves issue: #1756
Added method to get unescaped value of JobStepName propery in GwtJobStep class, used in GwtJobStep update method. Made change in JobStepEditDialog, so that so that value of jobStepName is unescaped when populating the dialog. Updated JobTargetGrid so that the values on the grid and tooltips are unescaped as well.


LeoNerdoG:
I have verified that this PR successfully resolves issue #1756. It can be merged when all tests complete.

Issue #1736, opened 2018-05-24
User Can Disable Itself


LeoNerdoG:
This is similar to issue #1445 but in this case I have found, that user can disable itself, but continue working until logout. After that Kapua reports that "user has been locked" and login is disabled (this last part is OK).

I suggest that we prevent user from disabling itself as this is done for kapua-sys user. 

Testflow: 
1. Login as kapua-sys or any other user with proper permissions
2. Go to Users, create a user (e.g. user0)
3. Login as user0 and go to Users
4. Select user0 and click Edit, under **Status** select **Disabled**
**Althoguh user is now disable, all the operations can be perfomerd normally (until logout).**

Screenshot: 
<img width="1439" alt="screen shot 2018-05-24 at 13 41 20" src="https://user-images.githubusercontent.com/15121999/40483544-a898b94a-5f58-11e8-8874-c23abc413b7f.png">

Browsers used: 
- Google Chrome (Version 66.0.3359.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (60.0.1  (64-bit))
- Opera (Version 53.0.2907.57)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1765, opened 2018-06-05
user can not disable himself

ct-goranpalibrk:

Brief description of the PR.
Added check if User which is edited is the current User and if it is then I added check if User status is disabled and if it is then throw exception.

**Related Issue**
This PR fixes issue #1736 

I think that everything is self-explained

**Screenshots**
/

/


LeoNerdoG:
I have verified that this PR successfully resolves issue #1736. It can be merged when all tests complete.

PR #1766, opened 2018-06-05
Fixed handling of selected item on multi-selectable grids

### LINKED ISSUE #1597: Targets Deselected When Pressing Refresh, ClientID or Display Name

Coduz:
Brief description of the PR.
Managed properly selected item before and after load.

**Related Issue**
This PR fixes #1597 

The `KapuaLoadListener` was designed to work with only single selected items. So the problem was that the `KapuaLoadListener` was keeping the selection of the last selected item, instead of all of them. 

Changed the single `selectedEntity` into a `List` of `selectedItems` so all items are tracked. 
Added a parameter which can configure if "remember selected item" feature is disabled or not.

**Screenshots**
_None_

Reordered a bit field in classes and added some javadocs

### DIFF
--- a/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/ui/grid/EntityGrid.java
+++ b/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/ui/grid/EntityGrid.java
@@ -47,22 +47,28 @@ public abstract class EntityGrid<M extends GwtEntityModel> extends ContentPanel 
     protected GwtSession currentSession;
     private AbstractEntityView<M> parentEntityView;
 
-    private EntityCRUDToolbar<M> entityCRUDToolbar;
-    private boolean entityGridConfigured;
-    private SelectionMode selectionMode = SelectionMode.SINGLE;
+    protected EntityCRUDToolbar<M> entityCRUDToolbar;
     protected KapuaGrid<M> entityGrid;
-    protected BasePagingLoader<PagingLoadResult<M>> entityLoader;
-    protected ListStore<M> entityStore;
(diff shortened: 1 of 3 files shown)


LeoNerdoG:
Hey @Coduz 
I have tested this branch and it is still not OK, because if user selects multiple devices and then tries to sort by Client ID or Display Name, only the last selected device (last two or three - it is random) stay(s) selected, all others are deselected. 
Also when refresh is clicked, nothing is deselected (as it should be?). Can you please check this again?

Coduz:
@LeoNerdoG I'll check for the case of sorting. The "keep selected" is now configurable. I'll change the parameter to disable this.

PR #1767, opened 2018-06-06
Fix for Job service unit tests

andrejn-ct:
The unit tests for the Job service are now correctly performed.

**Related Issue**
None

The setup phase for the Job service unit tests is enhanced to provide the actual implementation of the Job Engine service. This eliminates the null pointer exception that was previously thrown and the tests now successfully pass. Also, all the Job related tables are dropped between test scenarios, to make the scenarios independent.

**Screenshots**
None

None


Coduz:
Confirmed that this PR fixed tests.
https://hudson.eclipse.org/kapua/job/develop/

PR #1787, opened 2018-06-08
User Status validation in form when disabling self

### LINKED ISSUE #1777: User Status Field Not Coloured in Red When User Wants to Disable Itself
As the title of the issue implies, user can set its STATUS field to _DISABLED_, but Kapua returns an error in the bottom right corner (which is OK).The problem is, that field _STATUS_ is not coloured in red, so it si immediatelly visible, what the problem is. See testflow and screenshots for more information.

**Testflow:**
2. Go to Users, create a user (e.g. kapua-user0)
3. Add ALL ALL ALL permissions to kapua-user0
4. Login as kapua-user0 and go to Users
5. Select kapua-user0 (user that is currently logged in) and click Edit
6. Try to change status to DISABLED and submit changes
**Observe the error on the right (which is OK), but the filed, that is causing this error (STATUS) is not coloured in red.**

**Expected behavior**
STATUS field should be coloured in red.

**Screenshots**
screen shot 2018-06-07 at 17 07 12

**Version of Kapua**
1.0.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Others

**Main component affected**
[x] Console 
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 66.0.3359.181 (Official Build) (64-bit)
- Mozilla Firefox (60.0.1  (64-bit))
- Opera (Version 53.0.2907.57)

lorthirk:
Brief description of the PR.
The User Status field is now marked as invalid if a user tries to disable himself

**Related Issue**
This PR fixes #1777 

**Screenshots**
schermata 2018-06-08 alle 17 21 35

### DIFF
--- a/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/server/util/KapuaExceptionHandler.java
+++ b/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/server/util/KapuaExceptionHandler.java
@@ -15,6 +15,7 @@ import org.eclipse.kapua.KapuaEntityNotFoundException;
 import org.eclipse.kapua.KapuaEntityUniquenessException;
 import org.eclipse.kapua.KapuaErrorCodes;
 import org.eclipse.kapua.KapuaException;
+import org.eclipse.kapua.KapuaIllegalArgumentException;
 import org.eclipse.kapua.KapuaRuntimeException;
 import org.eclipse.kapua.KapuaUnauthenticatedException;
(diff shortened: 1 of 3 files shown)


LeoNerdoG:
@lorthirk just one question: for now, Kapua returns error messages in bottom right corner and coloures  problematic fields in red, if there are any problems. 
But here there is no error in bottom right corner, just  STATUS field is coloured in red. Is this OK?

PR #1825, opened 2018-06-25
EclipseInt - add initial implementation of Hono device registry servi…

dejanb:
…ces using Kapua


riccardomodanese:
Hi @dejanb and @stefanomorson , 
   I was thinking about how to proceed with the integration between the code of this branch and this pr.
I said in one of the last Kapua scum that we can integrate this pr into that branch but I'm not so sure now.
Referring to the slides projected in the last Kapua scrum the code of this branch is related to the "1.1" scenario and the pr is related to the "after 1.1" scenario.
So I was thinking if it's better to move the pr in a different branch since it doesn't need any code from this branch.
For the name I can suggest something like feature-eclipseiotHonoRegistry or something more general if this branch will own the whole code changes needed for the "after 1.1" scenario.

What do you think?

riccardomodanese:
I just rebased the feature-eclipseIot branch to develop

dejanb:
@riccardomodanese sounds good to me. I was thinking originally to have it all in one branch as they are both related to the integration effort, but they don't have to be.

PR #1835, opened 2018-06-26
Changed Bundle start/stop error handling

### LINKED ISSUE #1805: Error At Unsuccessfull Bundle Start Should Be Handled More Gracefully
When user wants to start a bundle on a live device from Devices -> Bundles -> Start and something goes wrong, Kapua returns an error, that is totally undescriptive (look at the screenshot). I think this error should be handled more gracefully. 

**Testflow:**
1) Login as kapua-sys or any other user that has proper permissions
2) Connect a live Kura device to Kapua (e.g. device0)
3) Go to Devices and selct device0
4) Go to its Bundles and select a inactive bundle (in this example _org.eclipse.kura.sun.misc_ is used)
5) Press Start and observe the consequences
**Kapua returns an error (which is OK, if something goes wrong), but this error should be more descriptive, handled more gracefully.**

**Expected behavior**
Error/error message should be handled more gracefully.

**Screenshot:**
screen shot 2018-06-13 at 10 38 48

**Version of Kapua**
1.0.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Others

**Main component affected**
[x] Console
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 66.0.3359.181 (Official Build) (64-bit))
- Mozilla Firefox (60.0.2  (64-bit))
- Opera (Version 53.0.2907.68)

ct-ajovanovic:
Brief description of the PR.
Changed Bundle start/stop error handling

**Related Issue**
This PR fixes #1805 

Method start() in **_DeviceBundleManagementServiceImpl_** class now throws _KapuaException_ with _BUNDLE_START_ERROR_ code, which is then handled in the  _**KapuaExceptionHandler**_. Similar solution is now applied for the stop() method of the same class. New error codes and messages were added to relevant classes.
Changes were also made in the **_DeviceTabBundles_** class , so that in case of an error the grid is no longer frozen and the toolbar and specific buttons are active.

**Screenshots**
None

Removed _BundleStartMenagementException_ and _BundleStartMenagementException_ classes as no longer needed. Instead of them _KapuaException_ with specific error code is thrown and handled. 


LeoNerdoG:
I have verified that this PR resolves issue 1805. @Coduz can you pleas check if the changes are OK and merge when all tests complete?

service/device/bundle/internal/src/main/java/org/eclipse/kapua/service/device/management/bundle/internal/DeviceBundleManagementServiceImpl.java
```diff
@@ -170,7 +169,7 @@ public void start(KapuaId scopeId, KapuaId deviceId, String bundleId, Long timeo
         if (!responseMessage.getResponseCode().isAccepted()) {
             KapuaResponsePayload responsePayload = responseMessage.getPayload();
 
-            throw new BundleStartManagementException(responseMessage.getResponseCode(), responsePayload.getExceptionMessage(), responsePayload.getExceptionStack());
+            throw new KapuaException(BundleManagementResponseErrorCodes.BUNDLE_START_ERROR);
```
Coduz: Why removing Specific exception for a generic one with the same ErrorCode?

service/device/bundle/internal/src/main/java/org/eclipse/kapua/service/device/management/bundle/internal/DeviceBundleManagementServiceImpl.java
```diff
@@ -222,7 +221,7 @@ public void stop(KapuaId scopeId, KapuaId deviceId, String bundleId, Long timeou
         if (!responseMessage.getResponseCode().isAccepted()) {
             KapuaResponsePayload responsePayload = responseMessage.getPayload();
 
-            throw new BundleStopManagementException(responseMessage.getResponseCode(), responsePayload.getExceptionMessage(), responsePayload.getExceptionStack());
+            throw new KapuaException(BundleManagementResponseErrorCodes.BUNDLE_STOP_ERROR);
```
Coduz: Why removing Specific exception for a generic one with the same ErrorCode?

service/device/bundle/internal/src/main/java/org/eclipse/kapua/service/device/management/bundle/internal/DeviceBundleManagementServiceImpl.java
LeoNerdoG: @Coduz The problem is in descriptivness of the error message. Before it was "Error: INTERNAL_ERROR,null,null,[Ljava.lang.Object;@18955d4a" for unsuccessfull bundle start, now it is a generic message saying that bundle could not be started. User can still look in logs for proper error message, but since the error box in bottom right corner is too small for decent message, why wanted to change this to a generic message. Do you think that we should solve this differently? Incorporating a specific error message of error code?

ct-ajovanovic: @Coduz Also there is the question of using this exception from service.device.management.bundle.internal package, which is then accessible to the GwtDeviceMenagementServiceImpl and its startBundle() method.
Should this class have access to exception from internal package? Which is then shown to the user?
it didn't look like a good solution to handle this internal exception in the KapuaExceptionHandler.
What do you think? What would be a better solution?

Coduz: @ct-ajovanovic  yeah! Right, I forgot about the fact that we still have exception classes on the `internal` part of a service... That unfortunately causes this problems... There is already a PR on this topic but there is no due date. Then is ok to leave things like they are! 

I'll approve the PR

Issue #1810, opened 2018-06-14
Unable To Edit Endpoints


LeoNerdoG:
If user wants to edit an endpoint, Kapua starts returning error as if user was trying to create another, identical endpoint. 

**Testflow:**
1) Login as kapua-sys or any other user that has roper permissions
2) Go to Endpoints, create an Endpoint (e.g. Schema: google.com, Domain Name: google.com, Port: 123, Secure: False)
3) Select the newly created endpoint and click **Edit**
4) Witohut changin anything, try to save this endpoint.
**Kapua returns an error, saying that endpoint with same parameteres already exists, although we are only editing the one.**

**Expected behavior**
Kapua should let user to submit changes (even if there are not any) without an error. 

**Screenshot:**
![screen shot 2018-06-14 at 13 16 27](https://user-images.githubusercontent.com/15121999/41409039-09aa8bfa-6fd5-11e8-8d5b-9bc6c3f614da.png)

**Version of Kapua**
1.0.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Openshift (in its variants) 
[ ] Others

**Main component affected**
[x] Console
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 66.0.3359.181 (Official Build) (64-bit))
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (60.0.2  (64-bit))
- Opera (Version 53.0.2907.68)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1841, opened 2018-06-28
fix editing endpoint without changing values

ct-goranpalibrk:

Brief description of the PR.
Removed method for checking if there is an Endpoint with same values from updae method in EndpointInfoServiceImpl.

**Related Issue**
This PR fixes issue #1810 

The problem was when User wants to edit Endpoint with same values, then in update method in service implementation we had method which throws exception when that Endpoint is founded. That method is removed now.

**Screenshots**


LeoNerdoG:
I have verified that this PR successfully solves issue 1810. Although only 8 lines have been deleted, GUI now works as inteded - I haven't found a scenario where this deletion would negatively impact on UX. 
@Coduz Can you please take a look and merge when all tests finish?

PR #1845, opened 2018-06-29
Optional Datastore indexing window by hour, day or week

lorthirk:
This PR introduces two new features in the Datastore:

1- Now the Datastore indexes can have a common prefix, configurable per-deployment via the `datastore.index.prefix` system property. The prefix will be prepended to all the indexes within the deployment
2- `MessageStoreService` now has an option to configure indexes width per week, day or hour, in order to have a control on how many messages will be contained in a single index. Such configuration is available per account among the other services configurations.

All the methods that have an access in the index name have been reworked according the the new implementations


lorthirk:
@Coduz Done!

Coduz:
This PR is waiting for integration test to be added.

muros-ct:
@lorthirk @Coduz PR #1908 is ready.

PR #1849, opened 2018-07-02
Fix verifyCredentials exception

### LINKED ISSUE #1847: User Can Change Password Without Knowing Old Password
When user is chaning its password, he must enter old password an then twice new one. But for now, user can enter anything in "Old Password" field and Kapua will still accept new password.

**Testflow:**
1) Login as kapua-sys
2) Go to Users and create a user (e.g. user1) with password "K4puaS0$tware"
3) Logout and login as user1
4) In upper right corner click on "user1@kapua-sys" and select ""Change Password" 
5) In field "Old Password" enter anything except the real old password
6) In fields "New Password" and "Confirm Password" enter the new password. 
**Kapua will change old password succesfully, although user has enetered wrong old password!**

**Expected behavior**
Kapua should change password only if all the parameters are correct.

**Screenshot No. 1 (Entering the password credentials):**
screen shot 2018-07-02 at 09 38 25

**Screenshot No. 2 (Although old password not correct, Kapua changes it):**
screen shot 2018-07-02 at 09 38 36

**Version of Kapua**
1.0.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Others

**Main component affected**
[x] Console
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 67.0.3396.99 (Official Build) (64-bit))
- Mozilla Firefox (61.0 (64-bit))
- Opera (Version 54.0.2952.41)

lorthirk:
This PR fixes #1847, fixing a wrong return of an exception instead of throwing it in `AuthenticationServiceShiroImpl.verifyCredentials()`.

**Related Issue**
Fixes #1847

### DIFF
--- a/console/module/authentication/src/main/java/org/eclipse/kapua/app/console/module/authentication/server/GwtCredentialServiceImpl.java
+++ b/console/module/authentication/src/main/java/org/eclipse/kapua/app/console/module/authentication/server/GwtCredentialServiceImpl.java
@@ -15,8 +15,8 @@ import com.extjs.gxt.ui.client.data.BasePagingLoadResult;
 import com.extjs.gxt.ui.client.data.PagingLoadConfig;
 import com.extjs.gxt.ui.client.data.PagingLoadResult;
 import org.apache.commons.lang3.StringUtils;
+
 import org.eclipse.kapua.KapuaEntityNotFoundException;
-import org.eclipse.kapua.KapuaException;
 import org.eclipse.kapua.app.console.module.api.client.GwtKapuaException;
 import org.eclipse.kapua.app.console.module.api.server.KapuaRemoteServiceServlet;
 import org.eclipse.kapua.app.console.module.api.server.util.KapuaExceptionHandler;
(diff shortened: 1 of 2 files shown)


LeoNerdoG:
@lorthirk I have checked this and fount out that if user misstypes "old password" three times, Kapua locks user out. 
Can you please check this?

Coduz:
Hi @LeoNerdoG ,

I've had a brief discussion with @lorthirk about his and at first I think that is "ok" in the sense that is it acting as a protection agains "brute force" of the password, even though it is quite annoying.

You think that this is a limiting factor for the functionality? 

Regards, 

_Alberto_

LeoNerdoG:
Hey @Coduz @lorthirk,
I don't think this limits the functionality, because - as you said - this prevents brute forcing the password. 
There is, however, a workaround for this, if user has Credential permissions. If user misstypes his old password more than 3 times, Kapua locks him immediatelly but does not log user out, so user is still functional. 
User can then go to Users and select himself (if he has permissions) -> Credentials -> Unlock. 
So there is a kind of a workaround. 

Nevertheless, if this feature stays, I would recommend to log the user out after three failed attempts and locking him (because Kapua throws "User locked" message, but user is stil functional and this is somehow not logical from user standpoint)? 

If this will be the case, I would add a  message bellow the "old password" field saying: "After three failed attempts user gets locked!" (just like in Users -> Confirm Password field) or something similar, for informing the user about this. 

Regards,

Leonardo

Coduz:
I think that logging out the user makes sense.

The "three" limit is configurable and it is also possible disabling the lockout policy (which is the actual feature name). So to display that message properly you need some logic to determine which is the proper value (if there is one). 
If you want to commit to that is ok, but I would stay simple also because the lockout is an "obvious" thing if you are providing a wrong password.

Regards, 

_Alberto_

Issue #1800, opened 2018-06-13
Wire Graph Configuration Marked as Invalid


LeoNerdoG:
If user connects a live Kura device to Kapua, which contains a wire graph, Kapua will report an error if field value is longer than 255 characters. However, it will permit user to change tabs without asking for saving/discarding changes or informing user that there are fields with errors (as for example in ClockService if you leave clock.ntp.host field empty).

**Testflow:**
1) Login as kapua-sys or any other user that has proper permissions
2) Connect a live Kura device (e.g. device0) to Kapua, which contains a working Wire Graph
3) In Kapua, go to Devices, select device0 and go to Configuration
4) See Wire Service and on the right, the WireGraph field.
**The WireGraph field is coloured in red and Kapua rtports that this field should be limited to 255.**

**Expected Behaviour:**
WireGraph field should not be limited to 255 characters, but something greated (8192 or more), because some graphs can get very complex and so can their descriptions.

**Screenshot:**
![screen shot 2018-06-13 at 08 06 40](https://user-images.githubusercontent.com/15121999/41334065-700fda4a-6ee4-11e8-86b1-cbc92f68f162.png)

**Version of Kapua**
1.0.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Openshift (in its variants) 
[ ] Others

**Main component affected**
[x] Console
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 66.0.3359.181 (Official Build) (64-bit))
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (60.0.2  (64-bit))
- Opera (Version 53.0.2907.68)

muros-ct:
I will put this issue back on hold. This is still an issue as it is not possible to edit long fields that are defined in Kura. One solution to this issue would be to increase length of field to a larger value and not just 255.
Only real solution would be for Kura to provide max length of field in this case text field.

LeoNerdoG:
I have verified that PR #2592 has resolved this issue. The maximum length for this field has been increased to 1M, so this should solve all our problems with the wire graph (at least for now).

PR #1855, opened 2018-07-03
wireGraph configuration not marked as invalid anymore

ct-goranpalibrk:

value for max length in WireGraph field changed

Brief description of the PR.
PR for increasing max length in field for WireGraph input.

**Related Issue**
This PR fixes issue #1800 

Added check if parametar id is "wireGraph" and if it is then max length for field is increased to 32768

**Screenshots**
/

/

### DIFF
--- a/console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/configuration/DeviceConfigPanel.java
+++ b/console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/configuration/DeviceConfigPanel.java
@@ -457,7 +457,7 @@ public class DeviceConfigPanel extends LayoutContainer {
             field.setMaxLength(1);
             field.setValidator((CharValidator) validator);
         }
-        if (validator != null && validator instanceof StringValidator && field.getId().equals("WireGraph")) {
+        if (validator != null && validator instanceof StringValidator && !param.getId().equals("WireGraph")) {
             field.setValidator((StringValidator) validator);
         }
         return field;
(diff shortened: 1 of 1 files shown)


LeoNerdoG:
I have verified that this PR successfully resolves issue _#1800._ @Coduz Can you check if everything is OK and merge when all tests complete?

console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/configuration/DeviceConfigPanel.java
```diff
@@ -454,7 +457,7 @@ public void handleEvent(BaseEvent be) {
             field.setMaxLength(1);
             field.setValidator((CharValidator) validator);
         }
-        if (validator != null && validator instanceof StringValidator) {
+        if (validator != null && validator instanceof StringValidator && field.getId().equals("WireGraph")) {
```
lorthirk: This means that any `StringValidator` instance applied to a fild with id different than `WireGraph` will evaluate to false, and the validator will not be set. Please check this, and fix accordingly.

muros-ct:
@Coduz can you check what are our options here?
This config parameters doesn't have any field that could be used as length validation field.

muros-ct:
I am closing this PR, as solution proposed is not acceptable.
Currently there is no viable solution to this problem, as length of text field is not provided and therefore validation can not be set correctly. Field will stay un-editable in case it is longer than 256 characters. In case of wire graph that is ok, as it is not meant to be editable from Kapua side.

PR #1865, opened 2018-07-10
Duplicate count of batch job properties

muros-ct:
Duplicate records for job schedules caused wrong paging of schedules.
Count method doesn't work and was replaced by using size of returned list of schedules instead.

**Related Issue**
This PR fixes/closes _1760_

Count on properties of job doesn't work as expected. This caused wrong
paging on job schedules. Count method was removed and replaced by list
getSize method. That method already has info about number of records, this
is also more efficient approach.

Count method doesn't work because it does cartesian product on schdl_trigger_properties
multiple times while combining AND predicates. It works on query because
query is using DISTINCT on selected results.

**Screenshots**
Not applicable.

Count method should be investigated, as it doesn't work as expected.

### DIFF
--- a/commons/src/main/java/org/eclipse/kapua/commons/service/internal/ServiceDAO.java
+++ b/commons/src/main/java/org/eclipse/kapua/commons/service/internal/ServiceDAO.java
@@ -433,7 +433,7 @@ public class ServiceDAO {
 
         //
         // SELECT
-        criteriaSelectQuery.select(cb.count(entityRoot));
+        criteriaSelectQuery.select(cb.countDistinct(entityRoot));
 
         //
         // WHERE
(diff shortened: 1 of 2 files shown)


LeoNerdoG:
Hi @muros-ct, 
I have checked this and now user can add only 100 Schedules, after the 100th entry, nothing is added into GUI Schedule list. Looks like the paging is broken.
Can you please check this?

muros-ct:
I see the problem. Find is using only one page of data, and size is then only one page.
I am fixing it.

muros-ct:
@LeoNerdoG please re-test. I changed back to count and changed overall implementation of count to be DISTINCT.

PR #1866, opened 2018-07-11
Error messages for MaxNumberChildItems exceeded

ct-ajovanovic:
Brief description of the PR.
Added error messages for MaxNumberChildItems exceeded

**Related Issue**
This PR fixes/closes **_1853_**

In backend `ServiceImpl` classes relevant for this issue, create() method is updated, so that in case when max number of child items is exceeded  instead of `KapuaIlegalArgumentException` new `KapuaMaxNumberOfItemsReachedException`  is thrown, to which specific String value is passed. This exception is then handled with `KapuaExceptionHandler` where new `GwtKapuaException` is thrown.
Added new error code and message.

**Screenshots**
image

None


LeoNerdoG:
I havee verified that this PR successfully resolves issue _**1853**_. @muros-ct can you please take a look at the fix an merge if everything is OK?

LeoNerdoG:
@ct-ajovanovic @muros-ct,
I have verified that with the newest commit this works as intended so from GUI point of view this can be merged. 
Also the if-nesting has been replaced with more suitable else-if so this should be fine too.

ct-ajovanovic:
Hi @lorthirk 
I made the changes you  suggested. Can you please check if everything is ok now?
Hi @LeoNerdoG  
Can you please test this again?

Thanks
_Aleksandra_

LeoNerdoG:
Hey @ct-ajovanovic @lorthirk,
I have retested this and it still looks ok from GUI side. No issues found.

service/api/src/main/java/org/eclipse/kapua/KapuaNumberOfItemsExceededException.java
```diff
@@ -0,0 +1,32 @@
...
+    private static final long serialVersionUID = 8651132350411186861L;
+    private String argValue;
+
+
+    public KapuaNumberOfItemsExceededException(String argValue) {
+        this(KapuaErrorCodes.MAX_NUMBER_OF_ITEMS_REACHED, argValue);
+        this.argValue = argValue;
+    }
+
+    protected KapuaNumberOfItemsExceededException(KapuaErrorCodes code, String argValue) {
```
lorthirk: I don't think this constructor is required. We probably want to tie the specific error code to every `KapuaNumberOfItemsExceededException` instance

ct-ajovanovic:
@lorthirk @LeoNerdoG 
I removed the constructor in question, and updated the default public one. I also changed the name of the new Exception to `KapuaMaxNumberOfItemsReachedException`, to by more in sync with the new error code.  So now every instance of `KapuaMaxNumberOfItemsReachedException` is tied to `MAX_NUMBER_OF_ITEMS_REACHED` error code.  
Is it ok now?
Thanks
_Aleksandra_

LeoNerdoG:
Fix is still OK. @lorthirk can you check and merge please?

PR #1868, opened 2018-07-11
Fixed domain name / hostname validation for EndpointInfo

### LINKED ISSUE #1861: Endpoint info DNS validation still broken
**Describe the bug**
Endpoint info field `dns` validation still not work properly.

**To Reproduce**
1. Go to console
2. Login and go to Endpoint view
3. Try add a endpoint like `www.test-with-dash.com`

**Expected behavior**
It is marked as invalid domain name, even if is a fully valid DNS name

**Screenshots**
screen shot 2018-07-06 at 2 40 55 pm

**Version of Kapua**
1.0.0-M6

**Type of deployment**
All

**Main component affected**
Console

**Additional context**
https://tools.ietf.org/html/rfc1123

Coduz:
Fixed DNS validation regex.

**Related Issue**
This PR fixes #1861 

We created a regex inspired by the one contained in the [RFC 1123](https://tools.ietf.org/html/rfc1123).

**Screenshots**
screen shot 2018-07-11 at 5 31 23 pm

_None_


LeoNerdoG:
Hi @Coduz,
I have tested this and found an issue - user can now in Domain Name enter any string without TLD (e.g. "emtpy", "nothing", "123",...) and not only "localhost". 
Also if user enters only "1" in Domain Name, Kapua does not react when user clicks on Submit.

Can you please check? 

screen shot 2018-07-12 at 09 05 06

screen shot 2018-07-12 at 09 05 19

Coduz:
Hi @LeoNerdoG ,

we solved the problem of `1` hanging. I didn't changed the check on the length with required at least 3 chars.

I think that this now covers the majority of the cases. Some cases are not covered, (i.e. IP v6 stuff and others) but I think those can be left there until we have some complains from customer or other sources.

Regards,

_Alberto_

LeoNerdoG:
Hi @Coduz,
I have found the following bugs and now you decide if this should be merged or not.
1) I have just now noticed that the error message provided by Schema/Domain Name fields reference the wrong documnet. The message is: "The value must comply with RFC 3968 Section 3.1./3.2.2." but it should be: "...RFC **3986** section..." (numbers 8 and 6 should be switched)
2) Domain field does not accept IP v6 literals which are in square brackets (e.g. [::1]) - but you said thiis will no be covered.
3) Domain field accepts IPv4 forms that have more than 4 digit fields (e.g. 0.0.1.2.3.4.5.) or digits are larger than 255 (e.g. 256.901.932.999)
4) Domain field accepts random strings without TLD (e.g."random")
5) Domain field accepts "localhost" string with a TLD (e.g. "localhost.com")

Coduz:
Hi @LeoNerdoG 

Another version has been pushed.

4. hostname without TLD are valid WINS, I think we are ok with those
5. localhost.com is a valid domain, isn't it? 

Regards, 

_Alberto_

LeoNerdoG:
Hey @Coduz,
I wrote 4) and 5) because of this in RFC 3986:

>  Such a [Domain] name **_consists of a sequence of domain labels separated by ".",_**
   each domain label starting and ending with an alphanumeric character
   and possibly also containing "-" characters.  The rightmost domain
   label of a fully qualified domain name in DNS may be followed by a
   single "." and should be if it is necessary to distinguish between
   the complete domain name and some local domain.

      reg-name    = *( unreserved / pct-encoded / sub-delims )

>   If the URI scheme defines a default for host, then that default
   applies when the host subcomponent is undefined or when the
   registered name is empty (zero length).  For example, the "file" URI
   scheme is defined so that no authority, an empty host, and
   "localhost" all mean the end-user's machine, whereas the "http"
   scheme considers a missing authority or empty host invalid.

I have never used localhost with a TLD, and could not find that **localhost.com** is invalid domain, (except from the part in RFC 3986 that is shown above) so I guess it can stay then?
The sam applies to WINS. 
So if you think this is OK, you can merge. but maybe we can debate about this in a couple of minutes  when meeting starts =)

PR #1952, opened 2018-08-28
Make Certificate Service configurable

elbert3:
This updates the CertificateService API to make it configurable, allowing for example to configure and put limits on the number of certificates used by an account.

**Related Issue**
N/A

N/A

**Screenshots**
N/A

N/A


service/security/certificate/internal/src/main/java/org/eclipse/kapua/service/certificate/internal/CertificateServiceImpl.java
```diff
@@ -154,4 +157,24 @@ public Certificate generate(CertificateGenerator generator) throws KapuaExceptio
...
         throw new UnsupportedOperationException();
     }
+
+    @Override
+    public CertificateDomain getServiceDomain() {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public KapuaTocd getConfigMetadata() throws KapuaException {
```
Coduz: `getConfigMetadata()` must return something, otherwise when fetching available configurations from console form the `Settings` > `AccountSetting` view an error will occur.

This view looks for all configurable services and loads the values per account basis. 

I think that we can make this return a empty mocked up `KapuaTocd`.

service/security/certificate/internal/src/main/java/org/eclipse/kapua/service/certificate/internal/CertificateServiceImpl.java
```diff
@@ -154,4 +157,24 @@ public Certificate generate(CertificateGenerator generator) throws KapuaExceptio
...
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public Map<String, Object> getConfigValues(KapuaId scopeId) throws KapuaException {
```
Coduz: `getConfigMetadata()` must return something, otherwise when fetching available configurations from console form the `Settings` > `AccountSetting` view an error will occur.

This view looks for all configurable services and loads the values per account basis. 

I think that we can make this return a empty map.

service/security/certificate/internal/src/main/java/org/eclipse/kapua/service/certificate/internal/CertificateServiceImpl.java
elbert3: I've updated it to return an empty KapuaTocd for getConfigMetadata() and an empty Map for getConfigValues(), and this seems to fix the error in the console.  We might also want to consider fixing the underlying code to handle this scenario (a service returns null values).

Coduz:
hi @elbert3 ,

with 98c6150 I've fixed the Account Settings view. 

Regards, 

_Alberto_

PR #1975, opened 2018-09-06
fix the api doc can not perform try invocation under HTTPS scheme.

miricy:
Signed-off-by: alva.huang <alva@izhiju.cn>

This PR modifies Swagger UI  can only use HTTP access mode.

 remove the fix HTTP scheme code
`
beanConfig.setSchemes(new String[] { "http" });
`

### DIFF
--- a/rest-api/web/src/main/java/org/eclipse/kapua/app/api/web/RestApisApplication.java
+++ b/rest-api/web/src/main/java/org/eclipse/kapua/app/api/web/RestApisApplication.java
@@ -90,7 +90,6 @@ public class RestApisApplication extends ResourceConfig {
         beanConfig.setVersion(apiVersion);
         beanConfig.setBasePath(basePath);
         beanConfig.setTitle(apiTitle);
-        beanConfig.setSchemes(new String[] { "http" });
         beanConfig.setResourcePackage(apiResourcePackage);
         beanConfig.setScan(true);
     }


miricy:
Hi @lorthirk  if removing the setting altogether，  the swagger ui  can detect automatically.
and the ’api.war‘ can correct run on 'localhost' and the 'https tomcat server'

lorthirk:
The swagger.json URL is dynamically compiled according to the browser URL, as you can see in `rest-api/web/src/main/webapp/doc/index.html`, so your removal should have no effect in this context.

As for the cURL that gets generated, I think there should be a way to select either `http` or `https` by specifying them both with the `setScheme()` method. Just let me do some experiments.

lorthirk:
So, I added `https` to the available schema, but even if I didn't try on an actual HTTPS server it looks like Swagger UI is somehow ignoring the setting. Anyway it seems that on the new Swagger UI interface (that you can reach at `https://editor.swagger.io/`, you have the choice to switch to HTTPS, and since we will be soon updating our version of Swagger UI as well I'd rather add `https` among the available schemas instead of removing it altogether.

miricy:
Hi @lorthirk， I think It best consistent with  the Address bar， No need to choose。

lorthirk:
Yes, I agree. I need to investigate more on Swagger UI to understand whether it's something that can/should be fixed on that side, instead of removing the annotation.

miricy:
Ok， thank you for your reply，and praise your patience.

PR #1984, opened 2018-09-10
Removed Excel export, added Devices Export button

### LINKED ISSUE #1981: Removal of Excel exporting
Removal of the Excel format when exporting data from the web Admin console.

We currently offer both CSV and Excel exporting.

This is a bit redundant, since CSV files can be easily imported in Excel.

By dropping this feature, we also can drop the `org.apache.poi:poi:jar:3.11` dependency. 
@stefanomorson reported difficulties having a later version being approved. 
Dropping this, also will address a reported CVE: see Apache Poi [Home page](http://poi.apache.org/#20+March+2017+-+CVE-2017-5644+-+Possible+DOS+%28Denial+of+Service%29+in+Apache+POI+versions+prior+to+3.15)

_None_

**Additional context**
_None_

lorthirk:
This PR removes Excel export in Data and Device Events views, and also removes the dependency from Apache POI that was needed for Excel exporting.

It also restores the Export button in Devices view.

**Related Issue**
This PR fixes #1981 

I just removed the Excel-related classes, and replaced dropdown buttons with plain buttons.

Also, the servlet for the Device Export button was already present; I only created the new button and pointed the click event to the servlet

**Screenshots**
N/A

N/A


lorthirk:
Hey @LeoNerdoG,

It would be great if you could test the Devices export checking if all the filters are correctly applied when creating the file 😄

LeoNerdoG:
Hey @lorthirk I have tested this. Bellow are my findings for Device view (split in main list view and Event view) and for Data view. 
**_CSV from Device view:_**
1. When downloading Device list, there are no Connection status, no Last Event On and no Last Event Type data in the CSV file (Screenshot No. 1). I don't know if this is OK or not as this may be sesnitive information? 
2. The title of the document consists ID of the child account, which is inpractical for end user. I recommend changing this to child account name (Screenshot No. 2)
3. I would change the first two field names from "id" and "Account" to "Account Id" and "Account Name" for clarification (Screenshot No. 3) 

_**CSV from Device Events:**_
1.There are columns "Created By" and "Device ID" - could you change this to user friendly names? (Screenshot No. 4).
2. Event message is not completely visible (Screenshot No. 5 and No. 6)
3. Title of the document is not descriptive enaugh (in my case device in child0 account was called "aaa" and the title was child0-device-events-4. Look at Screenshot No. 7)


_**CSV from Data View:**_
1. If Date Range is set so nod ata is shown, Export Data button is disabled, although in the same scenario in Devices (if no devices are present) the Export button is enabled. Is this OK? Because although no data is in the list, this empty list can be a evidence, proof of empty Dataview? (Screenshot No. 8)


**Screenshot No. 1:**
screen shot 2018-09-11 at 09 48 03

**Screenshot No. 2:**
screen shot 2018-09-11 at 09 58 13

**Screenshot No. 3:**
screen shot 2018-09-11 at 10 04 03

**Screenshot No. 4:**
screen shot 2018-09-11 at 10 17 38

**Screenshot No. 5 (what is shown in the CVS document):**
screen shot 2018-09-11 at 10 19 03

**Screenshot No. 6 (what should actually be shown in CVS document):**
screen shot 2018-09-11 at 10 19 34

**Screenshot No. 7:**
screen shot 2018-09-11 at 10 44 29

**Screenshot No. 8:**
screen shot 2018-09-11 at 10 58 23

lorthirk:
I addressed all your concerns apart except the following ones:

> **_CSV from Device view:_**
> 1.There are columns "Created By" and "Device ID" - could you change this to user friendly names? (Screenshot No. 4).

The Device ID is... well, the ID  :) there's no "human" version of it; we could put in the Client ID instead but it would be redundant since it would never change; also, the Client ID is in the file name now

> 2. Event message is not completely visible (Screenshot No. 5 and No. 6)

This looks like an Excel issue. Try to import the CSV in Excel specifying all the correct format options and it should appear. Also, in Apple Numbers it gets correctly imported automatically.

LeoNerdoG:
Hey @lorthirk,
I have reviewed the changes and everything looks fine now except the last point: 

> **_CSV from Data View:_**
> 
> If Date Range is set so no data is shown, Export Data button is disabled, although in the same scenario in Devices (if no devices are present) the Export button is enabled. Is this OK? Because although no data is in the list, this empty list can be a evidence, proof of empty Dataview? (Screenshot No. 8)

Also I have another suggestion - the Timestamp column in Data CSV has "(UTC)" added so user knows the exact time when data was published. 
Can you add this also to:
 - Device's Created On column
 - Device's Last Event On column
 - Devices's Event Created On column
 - Devices's Event Received On column
 - Devices's Event Sent On column

I think this would be a good feature for end-user. 
Other than that I think it is OK.

lorthirk:
The Devices Export button should be disalbed when no devices are present. Can you confirm?

LeoNerdoG:
I am confirming that button is disabled when no devices are present in the device list.

Issue #1949, opened 2018-08-28
Stabilize TravisCI build 


Coduz:
**Is your feature request related to a problem? Please describe.**
We need to improve Travis CI builds on PR since they are a bit unstable in the last few weeks.

Main recurring problem is something related with the download of the Maven on the setup phase, which causes the build to time out, lasting more than 49mins

```
2690.07s$ wget -O maven.tar.gz http://www-eu.apache.org/dist/maven/maven-3/3.3.9/binaries/apache-maven-3.3.9-bin.tar.gz || wget -O maven.tar.gz http://www-us.apache.org/dist/maven/maven-3/3.3.9/binaries/apache-maven-3.3.9-bin.tar.gz
--2018-08-27 12:58:10--  http://www-eu.apache.org/dist/maven/maven-3/3.3.9/binaries/apache-maven-3.3.9-bin.tar.gz
Resolving www-eu.apache.org (www-eu.apache.org)... 95.216.24.32, 2a01:4f9:2a:185f::2
Connecting to www-eu.apache.org (www-eu.apache.org)|95.216.24.32|:80... failed: Connection timed out.
Connecting to www-eu.apache.org (www-eu.apache.org)|2a01:4f9:2a:185f::2|:80... failed: Cannot assign requested address.
Retrying.
--2018-08-27 13:00:18--  (try: 2)  http://www-eu.apache.org/dist/maven/maven-3/3.3.9/binaries/apache-maven-3.3.9-bin.tar.gz
Connecting to www-eu.apache.org (www-eu.apache.org)|95.216.24.32|:80... failed: Connection timed out.
Connecting to www-eu.apache.org (www-eu.apache.org)|2a01:4f9:2a:185f::2|:80... failed: Cannot assign requested address.
Retrying.
--2018-08-27 13:02:27--  (try: 3)  http://www-eu.apache.org/dist/maven/maven-3/3.3.9/binaries/apache-maven-3.3.9-bin.tar.gz
Connecting to www-eu.apache.org (www-eu.apache.org)|95.216.24.32|:80... failed: Connection timed out.
Connecting to www-eu.apache.org (www-eu.apache.org)|2a01:4f9:2a:185f::2|:80... failed: Cannot assign requested address.
Retrying.
--2018-08-27 13:04:37--  (try: 4)  http://www-eu.apache.org/dist/maven/maven-3/3.3.9/binaries/apache-maven-3.3.9-bin.tar.gz
[...]

PR #1985, opened 2018-09-11
Merge Travis build steps

lorthirk:
This PR merges the first two Travis build steps making them a single step with two subsequent Maven runs, to make sure that the dependencies built by the first step will be available during the actual build.

**Related Issue**
No related issue

By joining the two step into a single one, we are now safe to assume that the `swagger-ui` and `swagger-ui-lib` will be available in the same local Maven repository that will be used in the second step when running the whole build. 

**Screenshots**
N/A

N/A


Coduz:
This issue closes  #1949.

Issue #1897, opened 2018-07-24
Generic Error When Setting Reserved User In Connections


LeoNerdoG:
If user is trying to set same user as _**Reserved User**_ to two different connetions, Kapua returns a generic error saying: _**"Connection edit failed: An internal error occured: Error during Persistence Operations"**_

**Testflow:**
1) Login as kapua sys or any other user that has proper permissions
2) Connect two live Kura devices to kapua (e.g. device0, device1) or one device twice with two different names
3) Go to Connections and select device0 and set its Reserver User parameter to kapua-broker"
4) Now select device1 and try to set its Reserved User parameter also to kapua-broker
**Kapua will return an error, which is OK, because one user can be Reserved User only for one connection at a time, but Kapua should report an error with a more specific message.**

**Expected behavior**
Kapua should report an error with a more specific message.

**Screenshot:**
![screen shot 2018-07-24 at 13 33 49](https://user-images.githubusercontent.com/15121999/43136150-1124ba08-8f48-11e8-8897-e42e994e5b68.png)

**Version of Kapua**
1.0.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Openshift (in its variants) 
[ ] Others

**Main component affected**
[x] Console (in case of console please report info on which browser you encountered the problem)
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 68.0.3440.106 (Official Build) (64-bit))
- Safari (Version 11.1.2 (13604.3.5))
- Mozilla Firefox (61.0.2 (64-bit))
- Opera (Version 55.0.2994.44)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #1996, opened 2018-09-12
error message when User wants to add reserved user to device

ct-goranpalibrk:

made new exception which will be thrown when User wants to reserve User which is already reserved by another connection

dialog box not hide on failed update

edit dialog not hide on failed edit connections

reserved user field mark as invalid when user is already reserved by another connection

Brief description of the PR.
Changed error message when User wants to add reserved user to a connection but that user is already reserved by another connection. 

**Related Issue**
This PR fixes issue #1897 

Made new exception for reserved User.

**Screenshots**

### DIFF
--- a/service/api/src/main/java/org/eclipse/kapua/ConnectionServiceException.java
+++ b/service/api/src/main/java/org/eclipse/kapua/ConnectionServiceException.java
@@ -1,0 +1,25 @@
+/*******************************************************************************
+ * Copyright (c) 2011, 2018 Eurotech and/or its affiliates and others
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Eurotech - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.kapua;
+
+public class ConnectionServiceException extends KapuaException{
+
+    /**
+     * 
+     */
+    private static final long serialVersionUID = -3657002518916941124L;
+
+    public ConnectionServiceException(KapuaErrorCode code) {
+        super(code);
+    }
+
+}
--- a/service/api/src/main/java/org/eclipse/kapua/KapuaErrorCodes.java
+++ b/service/api/src/main/java/org/eclipse/kapua/KapuaErrorCodes.java
@@ -95,5 +95,10 @@ public enum KapuaErrorCodes implements KapuaErrorCode {
     /**
      * Max number of items is reached
      */
-    MAX_NUMBER_OF_ITEMS_REACHED
+    MAX_NUMBER_OF_ITEMS_REACHED,
+
+    /**
+     * User is already reserved by another connection
+     */
+    USER_ALREADY_RESERVED_BY_ANOTHER_CONNECTION
 }
--- a/service/api/src/main/java/org/eclipse/kapua/KapuaErrorCodes.java
+++ b/service/api/src/main/java/org/eclipse/kapua/KapuaErrorCodes.java
@@ -97,8 +97,4 @@ public enum KapuaErrorCodes implements KapuaErrorCode {
      */
     MAX_NUMBER_OF_ITEMS_REACHED,
 
-    /**
-     * User is already reserved by another connection
-     */
-    USER_ALREADY_RESERVED_BY_ANOTHER_CONNECTION
 }
--- a/service/device/registry/internal/src/main/java/org/eclipse/kapua/service/device/registry/connection/option/internal/DeviceConnectionOptionServiceImpl.java
+++ b/service/device/registry/internal/src/main/java/org/eclipse/kapua/service/device/registry/connection/option/internal/DeviceConnectionOptionServiceImpl.java
@@ -15,6 +15,7 @@ import org.eclipse.kapua.KapuaEntityNotFoundException;
 import org.eclipse.kapua.KapuaException;
 import org.eclipse.kapua.KapuaRuntimeErrorCodes;
 import org.eclipse.kapua.KapuaRuntimeException;
+import org.eclipse.kapua.UserAlreadyReservedException;
 import org.eclipse.kapua.commons.service.internal.AbstractKapuaService;
 import org.eclipse.kapua.commons.util.ArgumentValidator;
 import org.eclipse.kapua.locator.KapuaLocator;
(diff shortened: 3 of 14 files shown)


ct-goranpalibrk:
@Coduz Is this solution okay?

service/device/registry/internal/src/main/java/org/eclipse/kapua/service/device/registry/connection/option/internal/DeviceConnectionOptionServiceImpl.java
```diff
@@ -62,9 +64,16 @@ public DeviceConnectionOption update(DeviceConnectionOption deviceConnectionOpti
         //
         // Check Access
         KapuaLocator locator = KapuaLocator.getInstance();
+        DeviceConnectionOptionQuery query = new DeviceConnectionOptionQueryImpl(deviceConnectionOptions.getScopeId());
         AuthorizationService authorizationService = locator.getService(AuthorizationService.class);
         PermissionFactory permissionFactory = locator.getFactory(PermissionFactory.class);
         authorizationService.checkPermission(permissionFactory.newPermission(DeviceDomains.DEVICE_CONNECTION_DOMAIN, Actions.write, deviceConnectionOptions.getScopeId()));
+        DeviceConnectionOptionListResult list = query(query);
+        for (DeviceConnectionOption deviceConnectionOptionDB : list.getItems()) {
```
lorthirk: Instead of asking for all the reservation for a given scope and then iterate through all of them, I would rather look for the specific association (i.e. the one with the same user id but a different client id): if at least one exists you are already sure you need to throw the exception. The result is the same of your proposal, but it's better optimized since you don't have to iterate and check multiple useless items.

ct-goranpalibrk:
@lorthirk I made some changes, please take a look :)

service/device/registry/internal/src/main/java/org/eclipse/kapua/service/device/registry/connection/option/internal/DeviceConnectionOptionServiceImpl.java
```diff
@@ -68,11 +70,9 @@ public DeviceConnectionOption update(DeviceConnectionOption deviceConnectionOpti
         AuthorizationService authorizationService = locator.getService(AuthorizationService.class);
         PermissionFactory permissionFactory = locator.getFactory(PermissionFactory.class);
         authorizationService.checkPermission(permissionFactory.newPermission(DeviceDomains.DEVICE_CONNECTION_DOMAIN, Actions.write, deviceConnectionOptions.getScopeId()));
-        DeviceConnectionOptionListResult list = query(query);
-        for (DeviceConnectionOption deviceConnectionOptionDB : list.getItems()) {
-            if (deviceConnectionOptionDB.getReservedUserId() != null && deviceConnectionOptionDB.getReservedUserId().equals(deviceConnectionOptions.getReservedUserId())) {
-                throw new UserAlreadyReservedException(deviceConnectionOptions.getScopeId(), deviceConnectionOptions.getId(), deviceConnectionOptions.getReservedUserId());
-            }
+        query.setPredicate(new AttributePredicateImpl<>(DeviceConnectionAttributes.RESERVED_USER_ID, deviceConnectionOptions.getReservedUserId()));
```
lorthirk: The logic should be inverted. If you look for an entry with the same Reserved User ID, you would also find the reservation you're trying to change: basically this check will prevent any update. On the other hand, you should look for any reservation for the same device AND a different user name. If such one exists, you can't add a new one for the same user.

service/api/src/main/java/org/eclipse/kapua/ConnectionServiceException.java
```diff
@@ -0,0 +1,25 @@
...
+package org.eclipse.kapua;
+
+public class ConnectionServiceException extends KapuaException{
+
+    /**
+     * 
+     */
+    private static final long serialVersionUID = -3657002518916941124L;
+
+    public ConnectionServiceException(KapuaErrorCode code) {
```
Coduz: This class should have its own class for error codes

service/api/src/main/java/org/eclipse/kapua/KapuaErrorCodes.java
```diff
@@ -95,5 +95,10 @@
     /**
      * Max number of items is reached
      */
-    MAX_NUMBER_OF_ITEMS_REACHED
+    MAX_NUMBER_OF_ITEMS_REACHED,
+
+    /**
+     * User is already reserved by another connection
+     */
+    USER_ALREADY_RESERVED_BY_ANOTHER_CONNECTION
```
Coduz: Move this inside a dedicated `KapuaErrorCodes` (called ConnectionServiceErrorCodes) inside the device module

LeoNerdoG:
I have verified that after last commit everything work fine on GUI side so this can be merged when all tests complete. 
@Coduz @lorthirk can you please check the solution if it is OK and merge?

service/device/registry/internal/src/main/java/org/eclipse/kapua/service/device/registry/connection/option/internal/DeviceConnectionOptionServiceImpl.java
```diff
@@ -62,10 +69,23 @@ public DeviceConnectionOption update(DeviceConnectionOption deviceConnectionOpti
...
         // Check Access
         KapuaLocator locator = KapuaLocator.getInstance();
+        DeviceConnectionService deviceConnectionService = locator.getService(DeviceConnectionService.class);
         AuthorizationService authorizationService = locator.getService(AuthorizationService.class);
         PermissionFactory permissionFactory = locator.getFactory(PermissionFactory.class);
         authorizationService.checkPermission(permissionFactory.newPermission(DeviceDomains.DEVICE_CONNECTION_DOMAIN, Actions.write, deviceConnectionOptions.getScopeId()));
 
+        if (deviceConnectionOptions.getReservedUserId() != null) {
+            AndPredicateImpl deviceAndPredicate = new AndPredicateImpl();
+            deviceAndPredicate.and(new AttributePredicateImpl<>(DeviceConnectionAttributes.SCOPE_ID, deviceConnectionOptions.getScopeId()));
```
lorthirk: What's the point of this predicate? I'd say you don't need this since you already create the query for the correct scopeId in line 81

service/device/registry/internal/src/main/java/org/eclipse/kapua/service/device/registry/connection/option/internal/DeviceConnectionOptionServiceImpl.java
```diff
@@ -62,10 +69,23 @@ public DeviceConnectionOption update(DeviceConnectionOption deviceConnectionOpti
...
+            deviceAndPredicate.and(new AttributePredicateImpl<>(DeviceConnectionAttributes.RESERVED_USER_ID, deviceConnectionOptions.getReservedUserId()));
+            DeviceConnectionQuery deviceConnectionQuery = new DeviceConnectionQueryImpl(deviceConnectionOptions.getScopeId());
+            deviceConnectionQuery.setPredicate(deviceAndPredicate);
+            long cnt = deviceConnectionService.count(deviceConnectionQuery);
```
lorthirk: When looking for other connections that are already assigned to this user, you have to exclude this connection itself, otherwise you will always find at least one result. Please exclude the connection you're modifying from the result set

ct-goranpalibrk:
@Coduz I moved my exceptions to proper module.

Coduz:
Hi @ct-goranpalibrk ,

Now exception is in the right place, but why leave the `ConnectionServiceErrorCodes.java` in the `kapua-service-api` module?? They need to stay together....

_Alberto_

ct-goranpalibrk:
Hi @Coduz I moved that class into right module.

PR #2041, opened 2018-09-21
Fix Swagger home redirection with Jetty

lorthirk:
After merging #1966 it looks like Jetty proposes a wrong redirect from the REST API home page to Swagger UI. This PR removes the "index.jsp" file from the proposed redirect.

**Related Issue**
N/A

Simply replace `index.jsp` with an empty string in the default redirection

**Screenshots**
N/A

N/A


Coduz:
This PR does not affect tests.

PR #2042, opened 2018-09-21
Migrate Docker images definitions to Dockerfiles

### LINKED ISSUE #2016: Adopt usage of Dockerfile in assembly modules
Currently we keep the definition of the Dockerfile for the Docker container in the assembly pom.xml files.

ie: https://github.com/eclipse/kapua/blob/develop/assembly/api/pom.xml#L266

This has some limitations and it would be nice to adopt the native docker file, which support all of our usages and improve readability of the configuration itself

Move all current definition to a dedicated Dockerfile in the module under the folder `{moduleRootDir}/docker`

_None_

**Additional context**
During the refactoring, make also changes to adhere to the Linux system directory conventions

lorthirk:
Docker images definitions are now migrated to `Dockerfile`s. Fabric8 plugin will read such files instead of having configurations in `pom.xml` files.

**Related Issue**
This PR fixes #2016

A new `Dockerfile` has been created for every Assembly project; Fabric8 will read such files instead of using XML configurations to create the images

**Screenshots**
N/A

N/A


lorthirk:
After merging this PR, #2136 needs to be merged as well and released in the same version, otherwise OpenShift deployment will fail.

PR #2043, opened 2018-09-21
Additional simplification for Swagger UI home page rewrite

lorthirk:
An additional simplification after #2041. Should also additionally improve compatibility with Jetty.

**Related Issue**
N/A

Now simple HTTP redirection is used instead of a JSP based one

**Screenshots**
N/A

N/A


Coduz:
This PR does not affect tests.

PR #2082, opened 2018-10-04
Use SystemToken to authenticate Pooled MQTT Clients

### LINKED ISSUE #2031: User kapua-sys auto lockout for no apparent reason
I'm running kapua-1.0.0M7 in docker and the user kapua-sys keeps getting locked out.
I have created another admin account to unlock the kapua-sys user but every time I log out and try to log back in again kapua-sys is locked out. 

The password credential is set up never to expire.
Is there any setting that turns on auto-expiry of an account?

lorthirk:
Currently every Kapua application (Console, REST API) that wants to connect to the broker has to provide a username and a password via system properties (`transport.credential.username` and `transport.credential.password`). This means that every change to this user's password must be reflected in such properties.

With this PR System Tokens are introduced, so that knowing and configuring the connecting user's password in advance is not needed anymore.

Additionally, Shiro Realms have been refactored to better reflect classes hierarchy.

**Related Issue**
This PR fixes #2031

`PooledMqttClientFactory` will automatically generate, when connecting a new client, a new credential of type `SYS_TOKEN` that will only last 15 seconds (configurable via the `system.token.ttl` system property) and will be disabled after one successful login. `KapuaSecurityBrokerFilter` will then be able to match the login attempt against this credential.

Please be aware that the `transport.credential.password` system properties has been removed, and that `transport.credential.username` is now called `systoken.auth.username` and moved from `kapua-transport-mqtt` module to `kapua-security-shiro` module.

**Screenshots**
N/A

N/A

### DIFF
--- a/broker-core/src/main/java/org/eclipse/kapua/broker/core/plugin/KapuaSecurityBrokerFilter.java
+++ b/broker-core/src/main/java/org/eclipse/kapua/broker/core/plugin/KapuaSecurityBrokerFilter.java
@@ -52,6 +52,7 @@ import org.eclipse.kapua.broker.core.setting.BrokerSettingKey;
 import org.eclipse.kapua.commons.security.KapuaSecurityUtils;
 import org.eclipse.kapua.commons.security.KapuaSession;
 import org.eclipse.kapua.commons.setting.system.SystemSetting;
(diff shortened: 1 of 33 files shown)


lorthirk:
Obsolete after merging #3067.

### ** DO NOT DELETE THE BRANCH YET. **

PR #2098, opened 2018-10-08
Fixing failing test TransportAsync

muros-ct:
Wrong timeout duration comparison in test

Tests is failing because time duration that is checked is exactely 100
milli seconds. It should be more, but it is not. If execution is really
fast it could happen that it is exactly 100 milli seconds.

**Related Issue**
This is part of solution to _2088_

Time comparison was changed to also cover equal time duration.

**Screenshots**
Not applicable.

Solves failing travis, which was failing at this test on multiple occasions. Error was:
```
   [INFO] Running org.eclipse.kapua.client.gateway.spi.util.TransportAsyncTest
   PT0.1S
   [ERROR] Tests run: 3, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.239 s <<< FAILURE! - in 
   org.eclipse.kapua.client.gateway.spi.util.TransportAsyncTest
```

### DIFF
--- a/client/gateway/spi/src/test/java/org/eclipse/kapua/client/gateway/spi/util/TransportAsyncTest.java
+++ b/client/gateway/spi/src/test/java/org/eclipse/kapua/client/gateway/spi/util/TransportAsyncTest.java
@@ -42,7 +42,7 @@ public class TransportAsyncTest {
             Duration duration = Duration.between(start, Instant.now());
             System.out.println(duration);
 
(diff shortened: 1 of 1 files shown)


muros-ct:
@riccardomodanese This solves issues with failing travis on your branches feature-externalConsumer*
I saw it in most of travis builds.
Maybe you could use this change on your branch.

PR #2152, opened 2018-10-29
Managing offline Device errors fix

### LINKED ISSUE #2117: Managing Device's Errors When Device Is Offline
If device gets disconnected for any reason and user does not notice that, device still looks conected in the device main list. Problems occur when user wants to e.g. install/uninstall a package, execute a command etc...because Kapua starts reporting errors (which is basically OK) but the information in these errors are not user friendly. 

**Testflow:**
2. Connect a live Kura device to Kapua (e.g. device0)
3. Simulate sudden device disconnect: In Kura GUI disconnect the device from Kapua but do not refresh the Kapua Device page
4. In Devices select device0 and go to Packages - try to Install a package or uninstall a exeisting package - observe the error (Screenshot No. 1)
5. Go to Bundles - try to start/stop bundles (Screenshot No. 2)
6. Try to change a parameter in Configuration (Screenshot No. 3)
7. Try to execute a valid command in Command tab (e.g. "ls -l" - Screenshot No. 4)
8. Check device's Assets tab
**In every case Kapua returns some kind of error, which is OK, but they are not very user friendly.**

**Expected behavior**
There was an attempt to refresh device list every now and then but in the end was not implemented. I suggest we change error messages to something more suer friendly if user has a chance to encounter them.

**Screenshot No. 1:**
screen shot 2018-10-15 at 12 26 59

**Screenshot No. 2:**
screen shot 2018-10-15 at 12 23 36

**Screenshot No. 3:**
screen shot 2018-10-15 at 12 23 46

**Screenshot No. 4:**
screen shot 2018-10-15 at 12 21 52

**Version of Kapua**
1.1.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Others

**Main component affected**
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 68.0.3440.106 (Official Build) (64-bit))
- Safari (Version 11.1.2 (13604.3.5))
- Mozilla Firefox (61.0.2 (64-bit))

ct-ajovanovic:
Brief description of the PR.
Managing offline Device errors fix

**Related Issue**
This PR fixes/closes #2117 

Created new `DeviceMenagementException` which is used in `DeviceCallExecutor `class instead of the existing exception from `kapua-device-commons` package.  Handled this exception with the `KapuaExceptionHandler`. Added a  new Enum with needed error codes, as well as the necessary error messages.
In `FileServlet ` class added a new catch block for handling `DeviceMenagementException ` thrown when executing Command on offline device. Also, replaced the `ServletException` which was shown to the user in the default catch block of the same method.

**Screenshots**
None

None


ct-ajovanovic:
Hi @lorthirk @muros-ct 
Can you please take a look? What do you think about this solution?
Thanks
_Aleksandra_

service/api/src/main/java/org/eclipse/kapua/KapuaDeviceMenagementException.java
```diff
@@ -0,0 +1,22 @@
...
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Eurotech - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.kapua;
+
+public class KapuaDeviceMenagementException extends KapuaException{
```
lorthirk: Why are you creating a new exception here? `DeviceManagementException` didn't work for you?

ct-ajovanovic: It didn't look like a good idea to handle `DeviceManagementException`  with the `KapuaExceptionHandler`, as this is an exception from kapua-device-commons package, so I created that new exception. Should I change this?
It seems like there is no unified way of handling errors in Kapua, some are shown directly to the user from service side and internal packages and for others the KapuaExceptionHandler is used. 
So, what approach should be used here?

lorthirk: Well... `DeviceManagementException` is a `KapuaException` anyway, so from a strict inheritance perspective your new exception doesn't add any different semantics. On the other hand, if we don't have a dependency from `kapua-device-commons` already, it may be a good idea to not create one. Pinging @Coduz for his feedback!

console/core/src/main/java/org/eclipse/kapua/app/console/core/servlet/FileServlet.java
```diff
@@ -220,9 +223,17 @@ private void doPostCommand(KapuaFormFields kapuaFormFields, HttpServletResponse
...
             return;
+        } catch (KapuaDeviceMenagementException edme) {
+            logger.error("Device menagement exception", edme);
+            resp.sendError(404, edme.getMessage());
+            return;
         } catch (Exception e) {
             logger.error("Error sending command", e);
-            throw new ServletException(e);
+            try {
+                throw new KapuaException(KapuaErrorCodes.INTERNAL_ERROR);
```
Coduz: This has some reason to be? 
It seems a rather complicate way to print a stacktrace!

ct-ajovanovic: @Coduz The problem was that this ServletException was actually showed to the user. So we had something like this:
screenshot

My idea was to replace that ServletException with a more generic KapuaException(KapuaErrorCodes.INTERNAL_ERROR) which had to be surrounded with a try/catch block. 
What do you think? Should I change this?

Coduz: By throwing it at line 233 and catching it in 234, and not re-throwing it in the `catch` block causes the exception to be eaten. 

If you wanted to log the exception only, and not returning it to the user, just use

```
new KapuaException(KapuaErrorCodes.INTERNAL_ERROR).printStackTrace();
```
but the exception is already logged at line 231.

If your intention was to throw a KapuaException, the code doesn't do that, and completes the execution of the method with no exception at all in case of an `Exception` thrown by the main method.

LeoNerdoG:
Hey all,
I have checked this PR (also with the latest changes) and there is only one problem - in Configuration tab. 
Kapua returns a servlet Exception to end user, which is not ok. 
@lorthirk @Coduz do you think it should be ok if this is changed so Kapua reports a error like in Command tab (Screenshot No. 2)?

**Screenshot No. 1 (issue that persists):**
screenshot 2018-11-09 at 13 04 38

**Screenshot No. 2 (Suggestion for a error):**
screenshot 2018-11-09 at 13 10 51

lorthirk:
@LeoNerdoG yes, we should avoid exposing stack traces in the UI.

service/api/src/main/java/org/eclipse/kapua/KapuaDeviceMenagementException.java
Coduz: I'm fine with a more specific exception for the `DeviceManagementException`s.

I would just drop the `Kapua` and enforce that a specific set of error codes (as Enums)  is used.

```
public class DeviceMenagementException extends KapuaException{
[...]

public DeviceMenagementException(DeviceMenagementErrorCodes code) {
        super(code);
}

```

What you guys think?

LeoNerdoG:
Hi all, I have retested this and everything is OK regarding these offline messages. After we discuss Alberto's question we can merge this.

ct-ajovanovic:
Hello everyone,
I made the requested changes - renamed the new exception to `DeviceManagementException` and created a new Enum `DeviceMenagementErrorCodes` with the needed error codes.
Can you please check if everything is ok now?

LeoNerdoG:
Hey, I have retested this and it is still OK from GUI standpoint. Can one of the commiters check the code please?

PR #2168, opened 2018-11-07
Fixing Jenkins Rest API tests of Kapua

muros-ct:
Minor changes for making Jenkins test run successfully for Rest API.

**Related Issue**
This PR in not related to any issue.

Embedded H2 database is run locally for rest api tests and exposes TCP connection on local host.
For that it had to be configured in a way to specify localhost as a connection endpoint and not host itself, which is default.

**Screenshots**
Not applicable.

This is just a code part that needs changes, build process on Jenkins has to be changed to. Changes provide similar approach to tests as is used on Travis CI.


muros-ct:
I have successfully run local jenkins CI with all the tests split into junit and separate cucumber tags tests. Works fine, and provides coverage reports.

PR #2171, opened 2018-11-09
Fix user role add form

### LINKED ISSUE #1987: Adding a Role to a User from a parent account causes error
**Describe the bug**
Adding a Role to a User of a child account causes an error.

**To Reproduce**
1. Create a child account (setting limits accordingly)
2. Create a user from the parent account into the newly created account.
3. Create a role from the parent account into the newly created account.
4. Try add a role to the specified user.

First the list of actual roles in the `grant role` is coming from the parent scope and not from the scope of the selected account.

This leads to the second error, which is caused by granting a parent-scoped role to a child-scoped user.

**Expected behavior**
Available role list should be populated with the Roles in the selected scope.

**Screenshots**
_None_

**Version of Kapua**
1.0.0

**Type of deployment**
All
**Main component affected**
Console - All browsers

**Additional context**
_None_

Coduz:
Fixed the retrieval of the Role list into the user role add dialog.

**Related Issue**
This PR fixes #1987 

Changed from `currentSession.getAccountId()` to `currentSession.getSelectedAccountId`.

**Screenshots**
_None_

Fixed also the access role view since it resulted broken when testing this fix

### DIFF
--- a/console/module/authorization/src/main/java/org/eclipse/kapua/app/console/module/authorization/client/tabs/role/AccessRoleAddDialog.java
+++ b/console/module/authorization/src/main/java/org/eclipse/kapua/app/console/module/authorization/client/tabs/role/AccessRoleAddDialog.java
@@ -20,7 +20,6 @@ import com.extjs.gxt.ui.client.widget.form.ComboBox.TriggerAction;
 import com.google.gwt.core.client.GWT;
 import com.google.gwt.user.client.Element;
 import com.google.gwt.user.client.rpc.AsyncCallback;
-import java.util.List;
 import org.eclipse.kapua.app.console.module.api.client.GwtKapuaException;
 import org.eclipse.kapua.app.console.module.api.client.messages.ConsoleMessages;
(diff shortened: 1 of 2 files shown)


LeoNerdoG:
Hi @Coduz,
I tried to build this branch and check the solution but it does not build - there are checkstyle errors. Can you please check? 

screenshot 2018-11-09 at 10 53 09

screenshot 2018-11-09 at 10 53 01

Coduz:
Fixed

LeoNerdoG:
Hi @Coduz,
I have tested this branch and found one issue - if user adds at least two pages of roles to a user and then deletes all of the roles on page 2 (so effectivelly only one page of roles exists), Kapua shows "Page 1 of 1" but there are no results (See scrneeshot bellow). Can you please check? 

**Screenshot No. 1 (What Kapua shows):**
screen shot 2018-11-09 at 12 51 02

**Screenshot No. 2 (What should be shown):**
screen shot 2018-11-09 at 12 51 28

PR #2191, opened 2018-11-19
Added JobStepDefinition filtering on add dialog

Coduz:
This PR adds the capability to exclude some `JobStepDefinition`s using a regex.

**Related Issue**
_None_

Each `JobStepDefinition`.`getName()` will be matched against the regex in the Console setting file. If matched will be not included in the list.

**Screenshots**
_None_

_None_


Coduz:
These changes do not affect tests and [build step](https://travis-ci.org/eclipse/kapua/jobs/456898358) has passed.

PR #2203, opened 2018-11-23
Fix typo in the docker guide

jbtrystram:
There was a typo in the guide, pointing to a non-existing directory.
I also added docker-compose a requirements, as it is often a separate package.

### DIFF
--- a/docs/developer-guide/en/running.md
+++ b/docs/developer-guide/en/running.md
@@ -12,13 +12,13 @@ setup and not on a full blown production setup.
 
 Before running Kapua on Docker, you need to 
 
-1. Install docker
+1. Install `docker` and `docker-compose`
 1. Run `mvn -f assembly -Pdocker` once to build containers
 
 
 Now, you can start Kapua by using Docker Compose. To do so, run
   
-    kapua/deployments/docker/docker-deploy.sh
+    kapua/deployment/docker/docker-deploy.sh
     
 After Kapua has been started, you can navigate your browser to http://localhost:8080 and log in using the following credentials:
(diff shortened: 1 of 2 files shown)


LeoNerdoG:
Hej @jbtrystram,
Why did you change the first line from "docker" to "'docker' and 'docker-compose'?
I do not understand, because as fas as I understandt the procedure, user has to first install Docker program/application as a whole and not docker and docker-compose? 

Or am I missing somethng else? 

Also if you are changing typos in a file, make sure to read the file entirely and correct all the mistakes ("deployments" keyword should be changed to "deployment" in whole file, not just for the Docker setup). 

BR,
Leo

jbtrystram:
> Why did you change the first line from "docker" to "'docker' and 'docker-compose'?
> I do not understand, because as fas as I understandt the procedure, user has to first install Docker program/application as a whole and not docker and docker-compose?

because docker-compose does not come necessarily when istalling docker. It is the case for Mac & Windows distributions of docker. However, on linux, when installing docker from the docker website, docker-compose is not installed and requires additional installation. (See https://docs.docker.com/compose/install/)

I'll fix the remaining of the typos right away as well

LeoNerdoG:
Hej @jbtrystram,
thanks for the explanation =) 

Let us know when you commit new changes so we merge this as fast as possible. There should be no problems since only documentation will be changed. 

Leo

jbtrystram:
@LeoNerdoG other typos fixed, sqashed and pushed.

Coduz:
Hi @jbtrystram ,

Thank you for your contribution!

Issue #2196, opened 2018-11-20
Kapua Accepts Retry Interval And Cron Schedule


LeoNerdoG:
If user enters Retry interval and cron schdule in a job schedule, Kapua does not report anything, so this effectivelly means that Kapua has two intervals for one job. I think this is not OK, since tooltips say: **"Enter a valid retry interval for the job in seconds OR enter a valid Cron expression..."**

**Testflow:**
1. Login as kapua-sys or any other user that has proper permissions
2. Go to Jobs, create a job (e.g. job1)
3. Select job1 and go to its Schedules, create a schedule with a unique name. Also enter start date and time both of retry interval and Cron schedule. 
**Kapua does not report any error  although both, retry interval and Cron schedule are entered. **

**Expected behavior**
Kapua should report an error if both, retry interval and Cron schedule are defined for a job. 

**Screenshot:**
![screenshot 2018-11-20 at 12 43 21](https://user-images.githubusercontent.com/15121999/48771543-f9e24900-ecc1-11e8-91ce-6db184b89442.png)

**Version of Kapua**
1.1.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Openshift (in its variants) 
[ ] Others

**Main component affected**
[x] Console (in case of console please report info on which browser you encountered the problem)
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 69.0.3497.100 (Official Build) (64-bit))
- Safari (Version 12.0 (13604.3.5))
- Mozilla Firefox (63.0 (64-bit))
- Opera (Version 55.0.2994.44)

Coduz:
I think that they should be mutually exclusive

LeoNerdoG:
I have verified that this issue has been resolved.

PR #2214, opened 2018-11-28
error message when user select both cron expression and retry interval

ct-goranpalibrk:
Brief description of the PR.
Added check if retry interval and cron expression are filled. If they are, new exception is thrown.

**Related Issue**
This PR fixes issue #2196 

/

**Screenshots**
/

/


muros-ct:
@Coduz can you comment on this please. Is it ok to create new Exception with hierarchy?

Issue #851, opened 2017-10-11
Kapua UI Blinking Uncontrollably if Account Name Too Long


LeoNerdoG:
When trying to change accounts the Kapua UI starts to blink uncontrollably (Google Chrome and Opera), if account list contains names that are longer than 200 characters (consequently non of the accounts can be selected). 

The two screenshots provided show, how this looks like (change between the two is rapid, several times per second). 

<img width="1440" alt="screen shot 2017-10-11 at 13 37 58" src="https://user-images.githubusercontent.com/15121999/31438894-68bd29de-ae8a-11e7-9cd9-bb1cb7c31a97.png">

<img width="1434" alt="screen shot 2017-10-11 at 13 38 16" src="https://user-images.githubusercontent.com/15121999/31438899-6ce24d1e-ae8a-11e7-8b85-711913a73001.png">

In Mozilla Firefox there is a different problem, as seen on screenshot bellow (names of accounts disappear): 
<img width="1440" alt="screen shot 2017-10-11 at 14 00 50" src="https://user-images.githubusercontent.com/15121999/31439545-ceb76e0a-ae8c-11e7-9da6-f26a3b8edb7e.png">

In Safari there are no problems with long name accounts. 

Browsers used: 
- Google Chrome (Version 61.0.3163.100 (Official Build) (64-bit))
- Safari (Version 11.0 (12604.1.38.1.7))
- Mozilla Firefox (56.0.1 (64-bit))
- Opera (Version 48.0.2685.39)

Maybe the name of an account should be limited to 100-120 characters, as this is enough for any relevant name possible?

lorthirk:
@LeoNerdoG do we have a limit on account name characters?

LeoNerdoG:
@lorthirk I think this issue has been resolved with PR #1382 but I will recheck it to be sure.

LeoNerdoG:
I have verified that this issue is automatically fixed by PR mentioned above. Closing.

lorthirk:
Unfortunately I had to revert #1382 because with that implementation the account name would be partially hidden most of the time if it's more than a few characters long, which is a very common scenario. Can you please test again this issue and, if needed, reopen it after #1521 is merged?

LeoNerdoG:
I am reopening as PR #1382 has been reverted.

Coduz:
Is this still occurring?

LeoNerdoG:
Hi @Coduz this is still an issue, yes.

[...]

PR #2218, opened 2018-11-29
set max length to 50 for account name

ct-goranpalibrk:

Brief description of the PR.
Max length for account name reduced to 50.

**Related Issue**
This PR fixes issue #851 

/

**Screenshots**
/
/


LeoNerdoG:
I have verified that this PR successfuly resolves issue #851. I have proposed that we shorten the max length of child account to 50, as that seems to be a good limit as for a descriptive name and for the GUI limitations, so nothing blinks, goes out of bounds etc.

PR #2234, opened 2018-12-06
Test dockerization

muros-ct:
Added support for integration tests with docker for Kapua deployment

**Related Issue**
None

Integration and/or functional tests can now be run either with embedded servers or with servers running in dockers. With servers I mean ES, DB, console, broker and API.
 
**Screenshots**
Not applicable.

Tests with embedded servers are still successfully run on Travis and Jenkins.


assembly/broker/docker/Dockerfile
```diff
@@ -39,7 +39,7 @@ ENV ACTIVEMQ_OPTS "-Dcommons.db.schema.update=true \
                    -Dcommons.eventbus.url=\${SERVICE_BROKER_ADDR} \
                    -Dcertificate.jwt.private.key=file:///var/opt/activemq/key.pk8 \
                    -Dcertificate.jwt.certificate=file:///var/opt/activemq/cert.pem \
-                   -Dbroker.ip=broker"
+                   -Dbroker.ip=localhost"
```
lorthirk: Hey @muros-ct,

Have you also tried a Docker Compose deployment with this setting? I suspect it may lead to some issues...

muros-ct: I will try another docker build and try it again. I think that it didn't work with broker. Because it sets device broker to broker instead of actual broker address.
What exactly does broker mean in this case? It should be ip or dns name. Right?

lorthirk: It actually gets resolved by Docker Compose to the ip address of the broker, since they're in the same Compose template

muros-ct: I tried with broker as broker.ip.
Problem is that broker name is matched to broker ip only inside docker. But if I connect external device to that broker then device tries to communicate with broker by this name. But because device itself is not inside docker it can't connect to broker as it doesn't exist outside, where device is.
Now to approach this issue?

muros-ct: @LeoNerdoG can you please check my branch for this PR.
Does dockers still run? Especially, can a device connect to Kapua and can you issue commands on that device?

muros-ct:
If everyone is OK, I would merge this changes.
It effects only tests and changes are needed for further refactoring of qa-steps.

PR #2258, opened 2018-12-12
Updated CreatedBy/ModifiedBy grid properties

### LINKED ISSUE #2207: ID Instead Of Username In Child Accounts -> Users -> Created By
**Created by** column in Child Accounts -> Users has ID of the user that has created this one and not its username as this is the case in all other Created By columns. 

**Testflow:**
1. Login as kapua-sys
2. Go to Child Accounts and create an account (e.g. child0)
3. Create a user within this account (go to Users -> Add)
4. Observe "Created By" column 

**Expected behavior**
There should be user-frendly username and not its ID.

**Screenshot:**
screenshot 2018-11-26 at 08 32 58

**Version of Kapua**
1.1.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Others

**Main component affected**
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 69.0.3497.100 (Official Build) (64-bit))
- Safari (Version 12.0 (13604.3.5))
- Mozilla Firefox (63.0 (64-bit))

ct-ajovanovic:
Brief description of the PR.
Updated CreatedBy/ModifiedBy grid properties

**Related Issue**
This PR fixes/closes #2207 

- Set _N/A_ as the value displayed on the grid when the true value is hidden from the logged in user.
- Replaced the _Created By_ property in Users->Roles grid with _Granted By_ property and added the necessary implementation in the `GwtAccessRoleServiceImpl` class. 
- Added new `getUsersForAccount()` method to `GwtUserServiceImpl `class that returns results displayed on the Child Accounts -> Users grid. This method uses two accountIds:  `accountId` of the currently selected account for listing Users and another `currentSession.getSelectedAccountId()`  used for setting the _Created By_ property for each individual User. This method replaced the previously used `query` method of the same class that used only `accountId` for both queries and as a result hid some of the values from the currently logged user.

**Screenshots**
_None_

_None_

### DIFF
--- a/console/module/authorization/src/main/java/org/eclipse/kapua/app/console/module/authorization/client/group/GroupGrid.java
+++ b/console/module/authorization/src/main/java/org/eclipse/kapua/app/console/module/authorization/client/group/GroupGrid.java
@@ -14,9 +14,14 @@ package org.eclipse.kapua.app.console.module.authorization.client.group;
 import com.extjs.gxt.ui.client.data.PagingLoadConfig;
 import com.extjs.gxt.ui.client.data.PagingLoadResult;
 import com.extjs.gxt.ui.client.data.RpcProxy;
+import com.extjs.gxt.ui.client.store.ListStore;
 import com.extjs.gxt.ui.client.widget.grid.ColumnConfig;
+import com.extjs.gxt.ui.client.widget.grid.ColumnData;
+import com.extjs.gxt.ui.client.widget.grid.Grid;
+import com.extjs.gxt.ui.client.widget.grid.GridCellRenderer;
 import com.google.gwt.core.client.GWT;
 import com.google.gwt.user.client.rpc.AsyncCallback;
+import org.eclipse.kapua.app.console.module.api.client.messages.ConsoleMessages;
 import org.eclipse.kapua.app.console.module.api.client.ui.grid.EntityGrid;
 import org.eclipse.kapua.app.console.module.api.client.ui.view.AbstractEntityView;
 import org.eclipse.kapua.app.console.module.api.shared.model.query.GwtQuery;
@@ -35,6 +40,8 @@ public class GroupGrid extends EntityGrid<GwtGroup> {
 
     private static final GwtGroupServiceAsync GWT_GROUP_SERVICE = GWT.create(GwtGroupService.class);
     private static final ConsoleGroupMessages MSGS = GWT.create(ConsoleGroupMessages.class);
+    private static final ConsoleMessages CMSGS = GWT.create(ConsoleMessages.class);
+
     private GwtGroupQuery query;
     private GroupToolbarGrid toolbar;
 
@@ -86,6 +93,13 @@ public class GroupGrid extends EntityGrid<GwtGroup> {
         columnConfigs.add(columnConfig);
 
         columnConfig = new ColumnConfig("userName", MSGS.gridGroupColumnHeaderCreatedBy(), 200);
+        columnConfig.setRenderer(new GridCellRenderer<GwtGroup>() {
+
+            @Override
+            public Object render(GwtGroup gwtGroup, String property, ColumnData config, int rowIndex, int colIndex, ListStore<GwtGroup> store, Grid<GwtGroup> grid) {
+                return gwtGroup.getUserName() != null ? gwtGroup.getUserName() : CMSGS.notAvailable();
+            }
+        });
         columnConfig.setSortable(false);
         columnConfigs.add(columnConfig);
 
--- a/console/module/authorization/src/main/java/org/eclipse/kapua/app/console/module/authorization/client/group/GroupGrid.java
+++ b/console/module/authorization/src/main/java/org/eclipse/kapua/app/console/module/authorization/client/group/GroupGrid.java
@@ -14,14 +14,11 @@ package org.eclipse.kapua.app.console.module.authorization.client.group;
 import com.extjs.gxt.ui.client.data.PagingLoadConfig;
 import com.extjs.gxt.ui.client.data.PagingLoadResult;
 import com.extjs.gxt.ui.client.data.RpcProxy;
-import com.extjs.gxt.ui.client.store.ListStore;
 import com.extjs.gxt.ui.client.widget.grid.ColumnConfig;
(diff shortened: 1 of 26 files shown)


LeoNerdoG:
I have verified that this PR successfully resolves issue #2207. 
@Coduz can you please check if the solution is OK? I have added additional comment in the issue (basically some other issues that I noticed while testing) so if you think that this solution is OK, please approve and merge it.

console/module/authorization/src/main/java/org/eclipse/kapua/app/console/module/authorization/shared/model/GwtAccessRole.java
```diff
@@ -41,4 +41,11 @@ public void setRoleName(String roleName) {
         set("roleName", roleName);
     }
 
+    public String getGrantedByName() {
```
Coduz: Why not use the existing `createdBy` from `GwtUpdatableEntityModel`? There is any reason?  
The `createdBy` field in the `AccessRole` and `GwtAccessRole` represents who has created the "link" between a `User` and a `Role`, so it can be seen as "grantedBy"

console/module/job/src/main/java/org/eclipse/kapua/app/console/module/job/client/JobGrid.java
```diff
@@ -89,6 +95,14 @@ protected void selectionChangedEvent(GwtJob selectedItem) {
         columnConfigs.add(columnConfig);
 
         columnConfig = new ColumnConfig("userName", MSGS.gridJobColumnHeaderCreatedBy(), 200);
+        columnConfig.setRenderer(new GridCellRenderer<GwtJob>() {
```
Coduz: Uhmmm this column can be namend "CreatedBy" since also the MSGS refers to a `gridJobColumnHeaderCreatedBy` messages. 

I think that `username` column name is a mistake :) 

After rename, the `CreatedByNameCellRenderer` could be used. Please also remove `getUsername()` and `setUsername()` from GwtJob  and use `createtBy` instead.

console/module/authorization/src/main/java/org/eclipse/kapua/app/console/module/authorization/shared/model/GwtRolePermission.java
```diff
@@ -11,10 +11,10 @@
  *******************************************************************************/
 package org.eclipse.kapua.app.console.module.authorization.shared.model;
 
-import org.eclipse.kapua.app.console.module.api.shared.model.GwtEntityModel;
+import org.eclipse.kapua.app.console.module.api.shared.model.GwtUpdatableEntityModel;
 import org.eclipse.kapua.app.console.module.authorization.shared.model.GwtPermission.GwtAction;
 
-public class GwtRolePermission extends GwtEntityModel {
+public class GwtRolePermission extends GwtUpdatableEntityModel {
```
Coduz: RolePermission is not updatable.
Please revert this.

console/module/authorization/src/main/java/org/eclipse/kapua/app/console/module/authorization/client/group/GroupGrid.java
```diff
@@ -86,6 +93,13 @@ protected void load(Object loadConfig,
         columnConfigs.add(columnConfig);
 
         columnConfig = new ColumnConfig("userName", MSGS.gridGroupColumnHeaderCreatedBy(), 200);
+        columnConfig.setRenderer(new GridCellRenderer<GwtGroup>() {
```
Coduz: Uhmmm this column can be namend "CreatedBy" since also the MSGS refers to a `gridGroupColumnHeaderCreatedBy` messages. 

I think that `username` column name is a mistake :) 

After rename, the `CreatedByNameCellRenderer` could be used. Please also remove `getUsername()` and `setUsername()` from GwtGroup  and use `createtBy` instead.

console/module/authorization/src/main/java/org/eclipse/kapua/app/console/module/authorization/shared/model/GwtRolePermission.java
ct-ajovanovic: @Coduz The reason for this change was that when using the new `CreatedByNameCellRenderer<GwtRolePermission>()` in the `RolePermissionGrid` I got this error:

`Bound mismatch: The type GwtRolePermission is not a valid substitute for the bounded parameter <M extends GwtUpdatableEntityModel> of the type CreatedByNameCellRenderer<M>`

Without that change in GwtRolePermission class, I guess we could create a `new GridCellRenderer` in the `RolePermissionGrid` like this:
```
        columnConfig.setRenderer(new GridCellRenderer<GwtRolePermission>() {

			@Override
			public Object render(GwtRolePermission model, String property, ColumnData config, int rowIndex,
					int colIndex, ListStore<GwtRolePermission> store, Grid<GwtRolePermission> grid) {
				return model.getCreatedBy() != null ? model.getCreatedByName() : COMMONS_MSGS.notAvailable();
			}
		});
```
Would that be a better solution?

Thanks,
_Aleksandra_

Coduz: Hi @ct-ajovanovic ,

I see. The problem is that `CreatedByNameCellRenderer` is defined as:

```
public class CreatedByNameCellRenderer<M extends GwtUpdatableEntityModel> implements GridCellRenderer<M> {
```

instead of 

```
public class CreatedByNameCellRenderer<M extends GwtEntityModel> implements GridCellRenderer<M> {
```

`CreatedByNameCellRenderer` only needs the entity `GwtEntityModel` to access `getCreatedBy` and `getCreatedByName`. 

`ModifiedByNameCellRenderer` instead requires a `GwtUpdatableEntityModel` (which extends from `GwtEntityModel`)  `getModifiedBy` and `getModifiedByName`. 

Changing that should solve the problem :) 

Regards,

_Alberto_

LeoNerdoG:
@Coduz @ct-ajovanovic 
I have checked this branch again and can confirm that everything is OK on the GUI side.

Issue #2186, opened 2018-11-15
Wrong error message when creating an already existing user


lorthirk:
**Describe the bug**
A wrong error message is shown when creating a new user whose name is already used by another user already in the system (even in a different scope)

**To Reproduce**
1. Create a user, i.e. `name-1`
2. Create another user, in any account, and name it `name-1`
3. Check the error message

**Expected behavior**
The error message should report that such username already exists

**Screenshots**
![unknown](https://user-images.githubusercontent.com/8140139/48568359-78746a80-e8ff-11e8-871e-7df720c12bca.png)

**Version of Kapua**
1.0.0

**Type of deployment**
[x] Local Vagrant deployment
[x] Docker
[x] Openshift (in its variants) 
[x] Others

**Main component affected**
[x] Console
[ ] REST API
[ ] Message Broker
[ ] - Others

**Additional context**
N/A

Coduz:
Don't close this! PR #2261 has not fixed the bug properly.

LeoNerdoG:
I have verified that this issue has been resolved.

PR #2261, opened 2018-12-13
Change error message when creating user which already exists

ct-anaalbic:

Brief description of the PR.
Change of error message if the user who is already exist in another account is created.

**Related Issue**
This PR fixes/issue #2186 

Added doPrivileged in create method in GwtUserServiceImpl. Also, text field for username in UserAddDialog is set in error state if entity exist in another account.

**Screenshots**
/
/


Coduz:
Hi all,

unfortunately the solution adopted does not fix the problem completely. 

Since it is performed from the `GwtUserServiceImpl` will only work on the Web Console. Currently REST API are still exposed to this bug. 

Another problem is using the `doPrivileged()` to create the whole user. Doing so, will enable an attacker to provide any user an being able to create it, since all security checks are skipped (because of the `doPrivileged()`.

The proper solution requires the revert of the changes introduced with this PR and the addition of the `doPrivileged()` to https://github.com/eclipse/kapua/blob/42cde20e6f3ae8cc9e4168a7f95f5ed164f9c8ed/service/user/internal/src/main/java/org/eclipse/kapua/service/user/internal/UserServiceImpl.java#L106

Hope that helps!

Regards, 

_Alberto_

PR #2268, opened 2018-12-18
Add Job Restart button

lorthirk:
This PR adds a new button in the Jobs view in the console to restart the selected job for all the targets assigned to the job. The difference between Start and Restart is that, with the former, every target will resume processing the job starting from the last successful step, while with the latter all the target will restart processing from the first step.

**Related Issue**
No related issue

Just added a new button and a new `restart()` method in `GwtJobEngineServiceImpl`

**Screenshots**
schermata 2018-12-18 alle 11 26 57

No side notes


LeoNerdoG:
Hi @lorthirk,
I have tested this in GUI and it looks OK. I would just like to know why a part of the text in the confirmation window is bold? 

I have also a suggesstion: do you think a "restart" button would be a good idea also in Target view - so user could restart a job just on certain device? I think this would be a good addition to the job functionality.

lorthirk:
About the bold part: well, to give it more emphasis :) But if you think it's not coherent with the other dialogs we can surely remove it.

Me and @Coduz discussed about adding it to the Targets tab as well, but we thought it wasn't necessary after all. Let's hear from him as well :)

Coduz:
Hi @LeoNerdoG ,

this was actually a request coming from a user. 
The idea is that when a job is run and completed, a user may wants to run again the same job easily. 

Currently a user must create another job with the same configuration. 

For now I would keep thing as they are and maybe add it if requested.

Regards, 

_Alberto_

LeoNerdoG:
Hi @Coduz,
ok, thanks for the explanation 😸

Issue #2221, opened 2018-11-30
"Not Available" Client ID When Device Deleted Before Adding It To a Job


LeoNerdoG:
If user deletes a device just before it is added as a target to a job, Kapua adds the device anyway and the _name_ of the device is "No available" (look at the screenshot).

**Testflow:**
1. Login as kapua-sys or any other user that has proper permissions
2. Go to Devices, add a device (e.g. device0)
3. Go to Jobs create a job (e.g. job0) and go to its Target tab, click Add
4. Now open kapua in new browser (e.g. Chrome or Opera), login as kapua-sys and go to Devices. 
5. Select device0 and delete it. 
6. Now go back to the first browser and try to add the deleted device. 

**Expected behavior**
Kapua should close the Add Targets window and report that "device is not available/missing" or soemthing similar.

**Screenshots**
![screenshot 2018-11-30 at 09 40 02](https://user-images.githubusercontent.com/15121999/49288448-0c6a3880-f4a1-11e8-935e-399ae8ab5d44.png)

**Version of Kapua**
1.1.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Openshift (in its variants) 
[ ] Others

**Main component affected**
[x] Console (in case of console please report info on which browser you encountered the problem)
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 69.0.3497.100 (Official Build) (64-bit))
- Safari (Version 12.0 (13604.3.5))
- Mozilla Firefox (63.0 (64-bit))
- Opera (Version 55.0.2994.44)

LeoNerdoG:
I have checked status of this issue in latest develop but it seems that it has not been fixed - @lorthirk, @muros-ct  can you check on your side please?
I have successfully reproduced this issue on two different machines (2 Firefox browsers, 2 Opera browsers) on latest develop.

LeoNerdoG:
@lorthirk @muros-ct 
I have rebuild the develop branch once again and now I get the following error if I just go to Jobs -> Targets:
![screenshot 2019-01-07 at 09 49 05](https://user-images.githubusercontent.com/15121999/50758187-c3589180-1261-11e9-98d9-4891329ba34f.png)

Can you check on your side if you can encounter this issue?

lorthirk:
It seems to work for me...

Chrome 71.0.3578.98 on MacOS 10.14.2

[...]

PR #2270, opened 2018-12-18
Added error message when trying to add deleted device to job

ct-anaalbic:

Brief description of the PR.
Error message has been added when trying to add deleted device to job.

**Related Issue**
This PR fixes/issue #2221 

Added error message instead of "Not available" text in the clientId column when the device that was deleted is added to the job.

**Screenshots**
/

/

### DIFF
--- a/console/module/job/src/main/java/org/eclipse/kapua/app/console/module/job/server/GwtJobTargetServiceImpl.java
+++ b/console/module/job/src/main/java/org/eclipse/kapua/app/console/module/job/server/GwtJobTargetServiceImpl.java
@@ -11,9 +11,9 @@
  *******************************************************************************/
 package org.eclipse.kapua.app.console.module.job.server;
 
-import com.extjs.gxt.ui.client.data.BasePagingLoadResult;
-import com.extjs.gxt.ui.client.data.PagingLoadConfig;
-import com.extjs.gxt.ui.client.data.PagingLoadResult;
+import java.util.ArrayList;
+import java.util.List;
+import org.eclipse.kapua.KapuaEntityNotFoundException;
 import org.eclipse.kapua.KapuaException;
 import org.eclipse.kapua.app.console.module.api.client.GwtKapuaException;
 import org.eclipse.kapua.app.console.module.api.server.KapuaRemoteServiceServlet;
@@ -37,9 +37,9 @@ import org.eclipse.kapua.service.job.targets.JobTargetFactory;
 import org.eclipse.kapua.service.job.targets.JobTargetListResult;
 import org.eclipse.kapua.service.job.targets.JobTargetQuery;
 import org.eclipse.kapua.service.job.targets.JobTargetService;
-
-import java.util.ArrayList;
-import java.util.List;
+import com.extjs.gxt.ui.client.data.BasePagingLoadResult;
+import com.extjs.gxt.ui.client.data.PagingLoadConfig;
+import com.extjs.gxt.ui.client.data.PagingLoadResult;
 
 public class GwtJobTargetServiceImpl extends KapuaRemoteServiceServlet implements GwtJobTargetService {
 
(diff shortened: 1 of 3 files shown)


LeoNerdoG:
I have verified that this PR successfully resolves issue #2221 from GUI side. Can one of the commiters check the code?

service/api/src/main/java/org/eclipse/kapua/KapuaEntityNotFoundException.java
```diff
@@ -52,6 +52,12 @@ public KapuaEntityNotFoundException(String entityType, KapuaId entityId) {
         this.entityId = entityId;
     }
 
+    public KapuaEntityNotFoundException(String entityType) {
```
Coduz: Please remove this and check comment on [GwtJobTargetServiceImpl.java](https://github.com/eclipse/kapua/pull/2270/files#diff-bf02bba069510b97ad33fa57f2702882)

console/module/job/src/main/java/org/eclipse/kapua/app/console/module/job/server/GwtJobTargetServiceImpl.java
```diff
@@ -112,18 +114,23 @@
...
+                         jobTargetCreator.setJobTargetId(GwtKapuaCommonsModelConverter.convertKapuaId(gwtJobTargetCreator.getJobTargetId()));
+
+                         // Create the Job Target
+                         JobTarget jobTarget = JOB_TARGET_SERVICE.create(jobTargetCreator);
+
+                         // convert to GwtJobTarget and return
+                         gwtJobTargetList.add(KapuaGwtJobModelConverter.convertJobTarget(jobTarget));
+                         //
+                    } else {
+                         throw new KapuaEntityNotFoundException(Device.TYPE);
```
Coduz: Please add the JobTargetId as a parameter, so the user has feedback on which device is not found.

console/module/job/src/main/java/org/eclipse/kapua/app/console/module/job/server/GwtJobTargetServiceImpl.java
```diff
@@ -112,18 +114,23 @@
...
-                //
-                // Create the Job Target
-                JobTarget jobTarget = JOB_TARGET_SERVICE.create(jobTargetCreator);
-
-                // convert to GwtJobTarget and return
-                gwtJobTargetList.add(KapuaGwtJobModelConverter.convertJobTarget(jobTarget));
+                } 
+                    Device device = DEVICE_REGISTRY_SERVICE.find(KapuaEid.parseCompactId(gwtJobTargetCreator.getScopeId()), KapuaEid.parseCompactId(gwtJobTargetCreator.getJobTargetId()));
+                    if(device != null) {
+                         KapuaId creatorScopeId = KapuaEid.parseCompactId(gwtJobTargetCreator.getScopeId());
```
Coduz: Fix indentation. 

If you are using Eclipse IDE we have formatting and cleanup profiles under `dev-tools/src/main/eclipse/codeStyle/` 

`

LeoNerdoG:
Hi @Coduz @ct-anaalbic,
I have retested this and I have to say that the ID that is in error message I do not like 😄 
I think the previous version was much better (without the ID part in error message) as this does not carry any specific information for end user. 

**Screenshot:**
screenshot 2018-12-19 at 15 32 10

Coduz:
HI @LeoNerdoG ,

we can trick the thing by adding to `GwtJobTargetCreator` a new field `clientId`. By populating that value, we can retrieve it on the `GwtJobTargetServiceImpl` and use `getClientId()` instead of `getJobTargetId()` when throwing the exception.

Would it work? 

Regards, 

_Alberto_

LeoNerdoG:
@Coduz If it works or not, we will see 😄 
I think this is a good solution if it would work, because everything is better than a "raw" client ID from database. 
Ana is already working on it. 

Regards,
Leo

LeoNerdoG:
@Coduz @ct-anaalbic 
I have checked the latest changes and I can confirm that this is OK from GUI side. 
I am also attaching a screenshot. 

screenshot 2018-12-21 at 09 51 20

Issue #1406, opened 2018-02-14
True/False Fields Can Be Implicitly Selected


LeoNerdoG:
All the True/False fields listed bellow in Kapua can be implicitly selected by clicking not directly on the field (but rather on space between text and the box, etc.).

Devices: 
 - All the true/false boxes in Configuration for every connected device (ClockService, CloudService...)
 - Child Accounts (all the true/false boxes in Account Settings)

Screenshot (red circle shows where user has actually clicked and the check box was selected):
![screen shot 2018-08-30 at 09 57 19](https://user-images.githubusercontent.com/15121999/44837915-2f374780-ac3b-11e8-870f-703898e87ef0.png)

Browsers used:
- Google Chrome (Version 64.0.3282.140 (Official Build) (64-bit)
- Safari (Version 11.0.3 (13604.3.5))
- Mozilla Firefox (58.0.2  (64-bit))
- Opera (Version 51.0.2830.26)

LeoNerdoG:
I have verified that this issue has been resolved.

Issue #1405, opened 2018-02-14
Check Boxes Can Be Implicitly Selected


LeoNerdoG:
All the check boxes listed bellow in Kapua can be implicitly selected by clicking not directly on the box (but rather on space between text and the box, etc.).

Devices: 
 - Operation Options (Device Packages -> Install)
 - Uninstall Package (Device Packages -> Uninstall)
Conncetions:
 - Allow User Change (Connections -> Edit Connection)
Users: 
 - Forwardable (User -> Permissions -> Grant Permissions)
Roles: 
 - Forwardable (User -> Permissions -> Grant Permissions)

Screenshot (red circle shows where user has actually clicked and the check box was set):
![screen shot 2018-08-30 at 09 53 51](https://user-images.githubusercontent.com/15121999/44837849-00b96c80-ac3b-11e8-8dc2-ca644647aebc.png)

Browsers used:
- Google Chrome (Version 64.0.3282.140 (Official Build) (64-bit)
- Safari (Version 11.0.3 (13604.3.5))
- Mozilla Firefox (58.0.2  (64-bit))
- Opera (Version 51.0.2830.26)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #2280, opened 2019-01-08
Fix for implicitly selected inputs

### LINKED ISSUE #1402: Dropdown Menus Can Be Implicitly Selected
All the dropdown menus listed bellow in Kapua can be implicitly selected by clicking not directly on the box (but rather on space between text and the box, etc.).

Devices: 
 - Device Status, Access Group (Add/Edit Device)
 - clock.provider (Device -> Configuration)
 - device.display-name (Device -> Configuration)
 - birth.cert.policy (Device -> Configuration)
 - payload.encoding (Device -> Configuration)
 - real.limit.time.unit (Device -> Configuration)
 - Iwt.qos (Device -> Configuration)
 - in-flight.persistance (Device -> Configuration)
 - protocol-version (Device -> Configuration)
 - ssl.hostname.verification (Device -> Configuration)
Connections:
 - User Coupling Mode,Reserved User (Edit Connection)
Data: 
 - Start Date,Start Time, End Date, End Time (Custom Date Range)
Jobs: 
 - Step Definition (Add Step)
 - Both Time fields (Add Schedule)
Users:
 - Status (New User dialogue box, Credentials dialogue box)
 - Credentials Type (New credentials dialogue box)
 - Role (Assign Role dialogue box)
 - Domain, Access Group, Action (Grant Permission dialogue box)
Roles: 
 - Domain, Access Group, Action (Grant Permission dialogue box)
Child Accounts:
 - Status (New User dialogue box, Credentials dialogue box)
 - dataIndexBy (MessageStoreService in Account Settings)
 - deviceConnectionUserCouplingDefaultMode (DeviceConnectionService in Account Settings)

Screenshot (red circle shows where user has clicked, orange circle shows that dropdown menu has expanded, although user did not clicked directly on it):
screen shot 2018-08-30 at 09 50 42

Browsers used:
- Google Chrome (Version 64.0.3282.140 (Official Build) (64-bit)
- Safari (Version 11.0.3 (13604.3.5))
- Mozilla Firefox (58.0.2  (64-bit))
- Opera (Version 51.0.2830.26)

ct-ajovanovic:

Brief description of the PR.
Fix for implicitly selected inputs.

**Related Issue**
This PR fixes/closes #1402, #1405, #1406

Disabled `pointer-events` on problematic labels(_.x-form-cb-label_, ._x-form-item-label_) using CSS. 
The radio input fields and their labels are wrapped with _.x-form-check-wrap_ div on which the pointer-events were enabled as a whole. After disabling the pointer events on the mentioned wrap div, those events were then enabled on the radio inputs themselves. 
Also the height of the _dirtyPlugin_, the ComponentPlugin used on the AccountConfigPanel was changed from 10 to 5 px. There was some overlapping of the unneeded empty space added by those additional 5px and the radio fields causing problems with field selection.

**Screenshots**
_None_

_None_

### DIFF
--- a/console/module/account/src/main/java/org/eclipse/kapua/app/console/module/account/client/AccountConfigPanel.java
+++ b/console/module/account/src/main/java/org/eclipse/kapua/app/console/module/account/client/AccountConfigPanel.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2018 Eurotech and/or its affiliates and others
+ * Copyright (c) 2017, 2019 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
@@ -126,7 +126,7 @@ public class AccountConfigPanel extends LayoutContainer {
                             dirtyIcon.setStyleName("x-grid3-dirty-cell");
                             dirtyIcon.setStyleAttribute("top", "0");
                             dirtyIcon.setStyleAttribute("position", "absolute");
-                            dirtyIcon.setSize(10, 10);
+                            dirtyIcon.setSize(10, 5);
                             dirtyIcon.show();
                         }
                         thePanel.fireEvent(Events.Change);
--- a/console/web/src/main/webapp/css/console.css
+++ b/console/web/src/main/webapp/css/console.css
@@ -271,3 +271,11 @@ button::-moz-focus-inner, button::-moz-focus-outer, button:-moz-focusring, butto
 	display: block; 
 	overflow: hidden;
 }
+
+.x-form-cb-label, .x-form-item-label, .x-form-check-wrap {
+	pointer-events: none;
+}
+
+.x-form-radio {
+	pointer-events: all;
+}


LeoNerdoG:
I have verified that this PR successfully resolves all of the above issues: #1402, #1405 and #1406.

muros-ct:
@lorthirk can you confirm this css changes, I am always afraid of those.

muros-ct:
It solves the issue, but there was a css change. What might change too?

lorthirk:
According to the pointer-event docs I read on [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events), this should be actually safe for our use case. Anyway, @LeoNerdoG, my suggestion is to have a quick test on any form field type (textboxes, textareas, datetimes, checkboxes, radios, fileinputs and any other input component you may think of).

LeoNerdoG:
Hi @lorthirk,
I have already tested all of the above in all of GUI (see my first comment) - that is why I wrote that everything is OK. 
I have tested everything that you mentioned above on 4 different places (front, back, up, down) and I haven't found anything. 
From my standpoint I think this can be merged.

Issue #2285, opened 2019-01-08
Unable To Add Devices To a Job If One Device Is Deleted


LeoNerdoG:
If there is at least one device that is not in the Device list any more (because it has been deleted from another browser) Kapua does not add any devices, instead error message is shown "".

I believe that better behaviour would be that Kapua adds all the devices that are in database and skips the ones that are not? Also we could create new info/error message that would inform user that not all the devices have been added (e.g. "Info: Some of the selected devices could not be added. Please check Device list").

**Testflow:**
1. Login as kapua-sys or any other user that has proper permissions
2. Go to Devices, add several devices (e.g. device0, device1...)
3. Go to Jobs, create a job (e.g. job0) 
4. Select job0 and go to its Targets tab. Select "Selected Only" option and select all the devices on this page
5. In another browser go to Devices tab and delete one of the devices
6. Press Submit in first browser (Jobs -> Target -> Add window)
**Kapua returns an error**

**Expected behavior**
I suggest that all the device that can be added, be added to the target list and others (that are not available) are skipped (see proposal above).

**Screenshot No. 1 (Selecting "Selected Only" option):**
![screenshot 2019-01-08 at 10 33 42](https://user-images.githubusercontent.com/15121999/50829604-4f8fb500-1345-11e9-8f2b-a2172e45dc49.png)

**Screenshot No. 2 (Selecting all the devices):**
![screenshot 2019-01-08 at 10 34 10](https://user-images.githubusercontent.com/15121999/50829602-4d2d5b00-1345-11e9-8cbe-1ce5a1721b8e.png)

**Screenshot No. 3 (Error thrown if at least one device is not in DB):**
![screenshot 2019-01-08 at 11 37 09](https://user-images.githubusercontent.com/15121999/50829597-4bfc2e00-1345-11e9-8df9-96d8fb0b16aa.png)

**Version of Kapua**
1.1.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Openshift (in its variants) 
[ ] Others

**Main component affected**
[x] Console (in case of console please report info on which browser you encountered the problem)
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 71.0.3578.98 (Official Build) (64-bit))
- Safari (Version 12.0 (14606.3.4))
- Mozilla Firefox (64.0 (64-bit))
- Opera (Version 57.0.3098.110)

[...]

PR #2309, opened 2019-01-15
Added available targets to job

ct-anaalbic:

Brief description of the PR.
All available devices are added to jobs.

**Related Issue**
This PR fixes issue #2285.

If some devices are deleted they are skipped and other available devices are added to job target list.

**Screenshots**
/

/


LeoNerdoG:
I have verified that this PR successfully resolves issue #2285.

LeoNerdoG:
@ct-anaalbic 
I have retested this and now I can see that Kapua adds the non-existing device anyway and the result is the same as in issue #2221 which has been fixed by you. 

**Screenshot:**
screenshot 2019-01-16 at 14 20 11

Can you please recheck the PR and make the necessarry fixes?

LeoNerdoG:
Hi @ct-anaalbic,
after last commit it looks like that issue has returned - the end result is same as in the original issue. 
Can you please check again? 

**Screenshot:**

ct-anaalbic:
@LeoNerdoG Yes, I can. I will see now.

LeoNerdoG:
@ana I have figured out what is wrong at last. 
If user is trying to add a list of devices and the first device is not in DB any more, Kapua will not throw proper error. 
But if user is trying to add a list of devices and first device is present in DB (it is not deleted) but there is a device somewhere in this list that is not in DB, kapua will successfully add all the other devices and skip the one, which is not in DB. 

So the problem must be in you "if" statement in the fix, please take a look.

LeoNerdoG:
I have verified that this PR with its last commit successfully resolves issue #2285 and can be merged if the solution is OK.

PR #2334, opened 2019-01-28
Fix for date field value changes not enabling the Submit button

### LINKED ISSUE #2298: Submit Button Not Enabled When Changes Are Made In Users -> Edit
If end user wants to edit certain user's Expiration date and the enter date is not valid, Submit button is not active (as it should be, although the date provided is not valid). 
This happens only if "previous" Expiration date was empty - if previous date was a valid one and user now enters invalid one, Submit button is active.

**Testflow:**
2. Go to Users, create a user (e.g. user0 - with empty Expiration date field!) or user one of the existing ones (they must have empty Expiration Date field)
3. Select user0 and click "Edit". Go to Expiration Date field and enter "1231231231"
4. Try to click Submit button - it is impossible, since it is disabled (but it should not be). 
5. now enter a valid date into Expiration Date field (e.g. 11/11/1111)
6. Click Submit
7. Repeat stepst 3 and 4 - now the Submit button is clickable (this is OK).

**Expected behavior**
Submit button should be enabled if Expiration date has valid date or not. 

**Screenshot No.1 (Expiration Date field is empty):**
screenshot 2019-01-09 at 14 23 20

**Screenshot No. 2 (entering invalid Date into Expiration Date field):**
screenshot 2019-01-09 at 14 23 25

**Screenshot No. 3 (Expiration Date field has valid Date):**
screenshot 2019-01-09 at 14 23 02

**Screenshot No. 4 (Entering invalid Date into Expiration Date field):**
screenshot 2019-01-09 at 14 23 11

**Version of Kapua**
1.1.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Others

**Main component affected**
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 71.0.3578.98 (Official Build) (64-bit))
- Safari (Version 12.0 (14606.3.4))
- Mozilla Firefox (64.0 (64-bit))
- Opera (Version 57.0.3098.110)

ct-ajovanovic:

Brief description of the PR.
Fix for date field value changes not enabling the Submit button

**Related Issue**
This PR fixes/closes #2298 

This PR solves problem regarding Submit button not being enabled if invalid Date value is entered in the KapuaDateField. The cause for this was that the previously created logic for enabling the Submit button depended only on the set value of the field - if the newly set value  of the field is different from its original value (isDirty).  With the DateField if the entered raw value is not a valid Date it will not be set, and this comparison would not work as the value for the field would stay null. 
This PR adds additional check for raw values being entered in the dateFields if the original value is null. Also new constructor and listener methods were added to the KapuaDateField. Affected Add/Edit dialogs were updated as well.

**Screenshots**
_None_

_None_

### DIFF
--- a/console/module/account/src/main/java/org/eclipse/kapua/app/console/module/account/client/toolbar/AccountAddDialog.java
+++ b/console/module/account/src/main/java/org/eclipse/kapua/app/console/module/account/client/toolbar/AccountAddDialog.java
@@ -60,7 +60,7 @@ public class AccountAddDialog extends EntityAddEditDialog {
     protected final KapuaTextField<String> accountNameField = new KapuaTextField<String>();
     protected final KapuaTextField<String> accountPassword = new KapuaTextField<String>();
     protected final KapuaTextField<String> confirmPassword = new KapuaTextField<String>();
-    protected final KapuaDateField expirationDateField = new KapuaDateField();
+    protected final KapuaDateField expirationDateField = new KapuaDateField(this);
 
     // broker cluster
     protected final NumberField optlock = new NumberField();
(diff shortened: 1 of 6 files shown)


LeoNerdoG:
I have verified that this PR successfully resolves issue #2298.

PR #2352, opened 2019-01-31
Proposal for solving delay in reporting a locked user

### LINKED ISSUE #1997: Kapua Reports Locking User After n+1 Attempt
If user sets lockoutPolicy.maxFailures to e.g. 4, Kapua will report that user is locked after the fifth attempt and not after fourth (see screenshots).

**Testflow:**
1. Login as kapua-sys
2. Go to Child Accounts and add a account (e.g. child0)
3. Set lockoutPolicy.maxFailure to 4, InfiniteChildUsers to True and add a user (e.g. user0)
4. Try to login 4 times with wrong password
**After the fourth attempt try to login again with wrong password and observe how Kapua just now reports that user has been locked.**

**Expected behavior**
Kapua should report that user has been locked after fourth attempt ant not after fifth. 

**Screenshot No. 1 (After fourth attempt Kapua only reports invalid username or password and not also that user has been locked):**
screen shot 2018-09-12 at 11 40 39

**Screenshot No. 2 (After fifth attempt Kapua that user has been locked, although this shoul be reported after fourth attemtp):**
screen shot 2018-09-12 at 11 40 01

**Version of Kapua**
1.0.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Others

**Main component affected**
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 68.0.3440.106 (Official Build) (64-bit))
- Safari (Version 11.1.2 (13604.3.5))
- Mozilla Firefox (61.0.2 (64-bit))

ct-ajovanovic:

Brief description of the PR.
Proposal for solving delay in reporting a locked user

**Related Issue**
This PR fixes/closes #1997 

As the lockout check is done before the actual login, the locking of the credentials is reported after additional login attempt and not at the moment when it's actually locked. 
This solution proposes additional check in the `handleTokenLoginException()` and the block for handling the `IncorrectCredentialsException`. If the credential is locked `KapuaAuthenticationException` with `LOCKED_LOGIN_CREDENTIAL` will be thrown. 

**Screenshots**
_None_

_None_

### DIFF
--- a/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
+++ b/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
@@ -39,6 +39,10 @@ import org.eclipse.kapua.locator.KapuaProvider;
 import org.eclipse.kapua.service.authentication.AuthenticationService;
 import org.eclipse.kapua.service.authentication.LoginCredentials;
 import org.eclipse.kapua.service.authentication.SessionCredentials;
+import org.eclipse.kapua.service.authentication.credential.Credential;
+import org.eclipse.kapua.service.authentication.credential.CredentialListResult;
+import org.eclipse.kapua.service.authentication.credential.CredentialService;
+import org.eclipse.kapua.service.authentication.credential.CredentialType;
 import org.eclipse.kapua.service.authentication.shiro.setting.KapuaAuthenticationSetting;
 import org.eclipse.kapua.service.authentication.shiro.setting.KapuaAuthenticationSettingKeys;
 import org.eclipse.kapua.service.authentication.token.AccessToken;
@@ -52,6 +56,8 @@ import org.eclipse.kapua.service.certificate.CertificateQuery;
 import org.eclipse.kapua.service.certificate.CertificateService;
 import org.eclipse.kapua.service.certificate.CertificateStatus;
 import org.eclipse.kapua.service.certificate.util.CertificateUtils;
+import org.eclipse.kapua.service.user.User;
+import org.eclipse.kapua.service.user.UserService;
 import org.jose4j.jws.AlgorithmIdentifiers;
 import org.jose4j.jws.JsonWebSignature;
 import org.jose4j.jwt.JwtClaims;
@@ -75,6 +81,8 @@ import java.util.UUID;
 public class AuthenticationServiceShiroImpl implements AuthenticationService {
 
     private static final Logger LOG = LoggerFactory.getLogger(AuthenticationServiceShiroImpl.class);
+    private UserService userService = KapuaLocator.getInstance().getService(UserService.class);
+    private CredentialService credentialService = KapuaLocator.getInstance().getService(CredentialService.class);
 
     static {
         // Make the SecurityManager instance available to the entire application:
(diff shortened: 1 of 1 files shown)


ct-ajovanovic:
Hi @lorthirk @Coduz @muros-ct 
Can you take a look at this PR? It affects the authentication process, and a deeper review would be appreciated.
Thanks
_Aleksandra_

LeoNerdoG:
I have verified that this PR resovles issue #1997 from console point of view. 
@Coduz @lorthirk can you please check if the solution is OK as we have been talking about? 

regards,
leo

lorthirk:
Conceptually I think that the implementation is correct. @Coduz, any thoughts to share?

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
```diff
@@ -477,4 +489,34 @@ private String generateJwt(KapuaEid scopeId, KapuaEid userId, Date now, long ttl
...
 
+    private Boolean checkIfCredentialHasJustBeenLocked(AuthenticationToken authenticationToken) throws KapuaException {
+        String principal = (String) authenticationToken.getPrincipal();
+        User user = KapuaSecurityUtils.doPrivileged(() -> userService.findByName(principal));
+        Credential credential = null;
+        if (user != null) {
+            credential = KapuaSecurityUtils.doPrivileged(() -> {
+                CredentialListResult credentialList = credentialService.findByUserId(user.getScopeId(), user.getId());
+
+                if (credentialList != null && !credentialList.isEmpty()) {
```
Coduz: `CredentialService.findByUserId()` will never return `null` as result . So this line can be removed.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
```diff
@@ -477,4 +489,34 @@ private String generateJwt(KapuaEid scopeId, KapuaEid userId, Date now, long ttl
...
+                    Credential credentialMatched = null;
+                    for (Credential c : credentialList.getItems()) {
+                        if (CredentialType.PASSWORD.equals(c.getCredentialType())) {
+                            credentialMatched = c;
+                            break;
+                        }
+                    }
+                    return credentialMatched;
+                } else {
+                    return null;
```
Coduz: Returning `null` here will produce a `NullPointerException` on the invocation on line 362.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
```diff
@@ -477,4 +489,34 @@ private String generateJwt(KapuaEid scopeId, KapuaEid userId, Date now, long ttl
...
+                }
+            });
+
+        }
+        Date now = new Date();
+        if (credential != null && credential.getLockoutReset() != null && now.before(credential.getLockoutReset())) {
```
Coduz: Just `return` the result of the `if`

```
return credential != null && credential.getLockoutReset() != null && new Date().before(credential.getLockoutReset())
```

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
ct-ajovanovic: The `null` value in question is assigned to the `credential` variable in case that the `credentialList` is empty. Than that credential is checked and boolean value returned. 
I can't see how that would produce a `NullPointerException`. Am I missing something?

ct-ajovanovic:
Hi @Coduz @LeoNerdoG 
I added a new commit with requested changes. Can you please check if everything is ok now?
Thanks,
Aleksandra

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
Coduz: Oh sorry, I missed the fact it was the return inside a lambda! So this is ok :)

Issue #2015, opened 2018-09-14
Generic Messages When No Items Are Added In Item Grid


LeoNerdoG:
This is more a suggestion tha an issue - I would like some feedback on this. 
When item lists are empty, there is a generic message "No result found". 
I suggest we change this to service specific error messages ("No connections/devices/jobs/... found") because there are also special messages for lack of Data ("No topics", "No Devices", "No Metrics"). 

I also suggest to change the message in the bottom right corner of the item list window (look at the screenshot).
Bellow there is a complete list of messsages I would change:

_**Connections:**_
 - "No result found" in connections item list into "No connections found"
 - "No result to show" in bottom right corner of item list into "No connections to show"
 - "Displaying 1-xxx of xxx results" in Paging toolbar into "Displaying 1-xxx of xxx connections"
 - "No item selected" in Description tab into "None of the connections selected"

_**Devices - general:**_
 - "No result found" in devices item list into "No devices found"
 - "No result to show" in bottom right corner of item list into "No devices to show"
 - "Displaying 1-xxx of xxx results" in Paging toolbar into "Displaying 1-xxx of xxx devices"
 - "No item selected" in Description tab into "None of the devices selected"

_**Devices - Tags:**_
 - "No result found" in device's Tag tab item list into "No tags applied to be shown"
 - "No result to show" in bottom right corner of Device's Tag item list into "No tags applied"
 - "Displaying 1-xxx of xxx results" in Tag paging toolbar into "Displaying 1-xxx of xxx tags"

_**Devices - Events:**_
 - "No Events Received" in device's Event tab list into "No events received"
 - "No result to show" in bottom right corner of Device's Tag item list into "No events to show"
 - "Displaying 1-xxx of xxx results" in Events paging toolbar into "Displaying 1-xxx of xxx events"

_**Batch Jobs - general:**_
 - "No result found" in Jobs item list into "No jobs found"
 - "No result to show" in bottom right corner of item list into "No jobs to show"
 - "Displaying 1-xxx of xxx results" in Jobs paging toolbar into "Displaying 1-xxx of xxx jobs"
 - "No item selected" in Description tab into "None of the jobs selected"

_**Batch Jobs - Targets:**_
 - "No result found" in Job's Target tab list into "No targets added"
 - "No result to show" in bottom right corner of Job's Target item list into "No targets to show"
 - "Displaying 1-xxx of xxx results" in Target paging toolbar into "Displaying 1-xxx of xxx targets"

[...]

PR #2394, opened 2019-02-13
Changed generic messages on grids and paging toolbars

ct-anaalbic:

Brief description of the PR.
Generic messages on grids and paging toolbars are changed.

**Related Issue**
This PR fixes issue #2015

Changed generic messages to be more specific on every grid and every paging toolbar. 

**Screenshots**
/

/


LeoNerdoG:
I have verified that this PR successfully resolves issue #2015 from console perspective.

ct-anaalbic:
Hi @Coduz,

I have looked your requested changes. Did you mean that I have to put messages only in one place, for example in `ConsoleMessages.properties`, and than when I implement messages argument will
be only entity name? I tryed to do this, but I'm not sure is it the best solution, because that look like this.

In `ConsoleMessages.properties` i made this changes:

`pagingToolbarNoResult = No {0}s to show`

But in implementation I have to accept parameter "user" and to return message, like 
`return MSGS.pagingToolbarShowingPost("user")`. This way of sending parameter isn't look ok, so I also tryed to put new message in existing one, like `MSGS.pagingToolbarShowingPost(MSGS.user())`, but I think this isn't good idea. Maybe I didn't understand you fine.

Regards,

_Ana_

LeoNerdoG:
I have verified that this PR is still OK. @Coduz can you plese check if the solution is OK?

PR #2396, opened 2019-02-13
Fix for permissions problems in Roles tabs

ct-ajovanovic:

Brief description of the PR.
Fix for permissions problems in Roles tabs

**Related Issue**
This PR fixes/closes part of issue 1801

Updated delete method in `RolePermissionServiceImpl` class, to check for _ACCESS_INFO_DOMAIN delete_ instead of _ROLE_DOMAIN delete_ permission. 
Set permissions needed for enabling the Add button in Roles->Permissions to reflect those already set on the service side (access_info:read, access_info:write, domain:read i role:write).
Updated permissions needed for enabling the Granted Users tab. 

**Screenshots**
_None_

This PR solves another part of issue 1801. The rest of the issue will be solved in the upcoming PRs.

### DIFF
--- a/service/security/shiro/src/main/java/org/eclipse/kapua/service/authorization/role/shiro/RolePermissionServiceImpl.java
+++ b/service/security/shiro/src/main/java/org/eclipse/kapua/service/authorization/role/shiro/RolePermissionServiceImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2017 Eurotech and/or its affiliates and others
+ * Copyright (c) 2011, 2019 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
@@ -128,7 +128,7 @@ public class RolePermissionServiceImpl extends AbstractKapuaService implements R
         KapuaLocator locator = KapuaLocator.getInstance();
         AuthorizationService authorizationService = locator.getService(AuthorizationService.class);
         PermissionFactory permissionFactory = locator.getFactory(PermissionFactory.class);
-        authorizationService.checkPermission(permissionFactory.newPermission(AuthorizationDomains.ROLE_DOMAIN, Actions.delete, scopeId));
+        authorizationService.checkPermission(permissionFactory.newPermission(AuthorizationDomains.ACCESS_INFO_DOMAIN, Actions.delete, scopeId));
 
         entityManagerSession.onTransactedAction(em -> {
             if (RolePermissionDAO.find(em, scopeId, rolePermissionId) == null) {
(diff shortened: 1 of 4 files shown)


LeoNerdoG:
I have veerified that this PR resolves part of issue 1801 as described in PR description.

muros-ct:
I am getting really worried regarding this kind of permission checks.
Permissions should be checked less explicitly.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authorization/role/shiro/RolePermissionServiceImpl.java
```diff
@@ -128,7 +128,7 @@ public void delete(KapuaId scopeId, KapuaId rolePermissionId) throws KapuaExcept
         KapuaLocator locator = KapuaLocator.getInstance();
         AuthorizationService authorizationService = locator.getService(AuthorizationService.class);
         PermissionFactory permissionFactory = locator.getFactory(PermissionFactory.class);
-        authorizationService.checkPermission(permissionFactory.newPermission(AuthorizationDomains.ROLE_DOMAIN, Actions.delete, scopeId));
+        authorizationService.checkPermission(permissionFactory.newPermission(AuthorizationDomains.ACCESS_INFO_DOMAIN, Actions.delete, scopeId));
```
Coduz: Why change this permission check? 

Since we this service manages the Permission assigned to the Role. So `role` was the correct permission to check. While `access_info` is the domain for the services/resources that are managing the assignment of the roles and permission to the user

ct-ajovanovic: @Coduz Before this change the user had to have `ROLE_DOMAIN` delete permission for deleting a permission (not a role, but permission) on Roles tab. It just didn't seem right. So, after serious thought I changed it to `ACCESS_INFO_DOMAIN` delete permission. Was I wrong?

Coduz: Actually that was right, since those are the permission associated to the Role. 

The fact that those changes affetcs user's permission is a side effect. But the actual service to manage permission and roles associated to a user are the `AccessPermissionService` and the `AccessRoleService` that both require `access_info`.

As I said before

> [...] `access_info` is the domain for the services/resources that are managing the assignment of the roles and permission to the user. [...]

Do you agree?

ct-ajovanovic: @Coduz Then this check should be changed to the previous state? The `role:delete` instead of the `access_info:delete` ? Than the affected permission delete button's enabled state should be changed as well, to need the `role:delete` permission.  Should the button need only `role:delete`? Or `access_info:delete` as well?
I wanted to confirm this with you, before making changes :)

Coduz: The only permission required to delete a Permission from a Role is `role:delete`.

So you can proceed to revert the change on the permission required in `RolePermissionServiceImpl` and change the permission checked when setting the enablement status of the Delete Button. You may also want to check that all permission on the Role view refers to the `RoleSessionPermission` except the enablement of the "Granted User" Tab, which requires `access_info:read`

Issue #2477, opened 2019-03-04
Edit Windows Too Small For Long Device/User Names


LeoNerdoG:
If device or user name is too long, they are not displayed properly when editing them (see screenshot bellow).

**Testflow:**
1. Login as kapua-sys or any other user that has proper permissions
2. Go to Device/Users and create an item with verey long name (dashes must be included)
3. Select the newly created item and click Edit

**Expected behavior**
Either windows should be enlarged or names should be shortened in order to see all the info in edit windows.

**Screenshot:**
![screenshot 2019-03-04 at 10 15 43](https://user-images.githubusercontent.com/15121999/53722999-0c465400-3e67-11e9-9ef5-be498aa1cea6.png)

**Version of Kapua**
1.1.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Openshift (in its variants) 
[ ] Others

**Main component affected**
[x] Console (in case of console please report info on which browser you encountered the problem)
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 72.0.3626.119 (Official Build) (64-bit))
- Safari (Version 12.0.3 (14606.4.5))
- Mozilla Firefox (65.0.1 (64-bit))
- Opera (Version:58.0.3135.79)

ct-anaalbic:
Hi @lorthirk.
I'am workint on this issue. With device dialogs i didn't have any problem, but in user edit dialog when I set auto height shadow not disappeared. It looks like this:
![image](https://user-images.githubusercontent.com/45655642/53941093-83722700-40b7-11e9-92cd-519fe765e7cf.png)
I tryed to remove this shadow, but unsuccessful. My idea is to set some limit to user name, like in accounts. Is this good solution? 

Thank's.
Regards.

ct-anaalbic:
Hi @lorthirk. I found other solution for this problem. Like I commented, this shadow on screenshot above can not be removed. My second solution is on screenshot bellow.

![image](https://user-images.githubusercontent.com/45655642/55713564-39030380-59f1-11e9-9119-d5616627d179.png)

Are you agree with this?

Regards,
_Ana_

lorthirk:
LGTM! @LeoNerdoG, do you think this is coherent enough with the rest of the UI?

LeoNerdoG:
Hey @lorthirk,
[...]

Issue #2483, opened 2019-03-07
Max Length Of Child Account Name 


LeoNerdoG:
In issues #851 #1191 and #1192 we have (or still are) discussing about long account/user names. 
For now the we have "fixed" issue 851 with PR #2218 by setting max legth of a child account to 50 characters. 
This is problematic if user creates a child account via REST API, because then user would see only first 50 characters out of 255 characters of child account's name in GUI so this should be solved differently. 

From extensive GUI testing I suggest that we limit usernames and account names to 50 or so characters, otherwise we will have problems with displaying them. None of this is problem when working with REST but because we also have GUI here, we should take this into consideration. 

**Testflow:**
1. Login as kapua-sys or any other user that has proper permissions
2. Go to Child Accounts, create an account that has 255 characters long name
3. Select the newly craeate account and click Edit

**Expected behavior**
We should consider different options and approaches for solving this issue. 

**Screenshot No. 1 (First problem with long account names):**
<img width="1434" alt="132123123123" src="https://user-images.githubusercontent.com/15121999/53946386-f4b7d700-40c3-11e9-8829-77d06f6815cf.png">

**Screenshot No. 2 (Second problem with long account names):**
<img width="1440" alt="12313423424234324" src="https://user-images.githubusercontent.com/15121999/53946394-f97c8b00-40c3-11e9-853b-b3ec83cf2b1e.png">

**Screenshot No. 3 (Third problem with long account names - sample screenshot from Users):**
![screenshot 2019-03-07 at 10 39 45](https://user-images.githubusercontent.com/15121999/53947067-5593df00-40c5-11e9-8ab2-bdf218a1ec21.png)

**Version of Kapua**
1.1.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Openshift (in its variants) 
[ ] Others

**Main component affected**
[x] Console (in case of console please report info on which browser you encountered the problem)
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 72.0.3626.119 (Official Build) (64-bit))
- Safari (Version 12.0.3 (14606.4.5))
- Mozilla Firefox (65.0.1 (64-bit))
- Opera (Version:58.0.3135.79)

lorthirk:
I feel that impose a limit on a backend functionality just to adapt to the UI is not right. It should be a UI responsibility to adapt and properly show the information correctly.

[...]

PR #2487, opened 2019-03-08
Changed height of dialogs and length of user name

ct-anaalbic:

Brief description of the PR.
Changed height of device dialogs and set user name length.

**Related Issue**
This PR fixes issue #2477 

User name is limited to 50 characters, and device dialogs are optimized to have auto height.

**Screenshots**
/

/

### DIFF
--- a/console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/DeviceAddDialog.java
+++ b/console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/DeviceAddDialog.java
@@ -25,6 +25,7 @@ import com.extjs.gxt.ui.client.widget.form.SimpleComboBox;
 import com.extjs.gxt.ui.client.widget.layout.FormData;
 import com.extjs.gxt.ui.client.widget.layout.FormLayout;
 import com.google.gwt.core.client.GWT;
+import com.google.gwt.user.client.Element;
 import com.google.gwt.user.client.rpc.AsyncCallback;
 import org.eclipse.kapua.app.console.module.api.client.GwtKapuaErrorCode;
 import org.eclipse.kapua.app.console.module.api.client.GwtKapuaException;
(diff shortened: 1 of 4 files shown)


LeoNerdoG:
This PR is a temporary solution for issue #2477. A new issue has been opened (#2483) which describes the problem in more detail. 
For now this can be merged but we should consider another fix for this in the future.

console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/DeviceAddDialog.java
```diff
@@ -107,7 +108,14 @@
...
 
-        DialogUtils.resizeDialog(this, 550, 450);
+        DialogUtils.resizeDialog(this, 550, 500);
+    }
+
+    @Override
+    protected void onRender(Element parent, int pos) {
+        super.onRender(parent, pos);
+        bodyPanel.setAutoHeight(true);
+        setAutoHeight(true);
```
Coduz: Just a question.

Would make sense to move 

```
bodyPanel.setAutoHeight(true);
setAutoHeight(true);
```

to some of the parent Classes? 
To cover all Dialogs and not only the DeviceAdd one
Or does it causes other problems?

ct-anaalbic: Hi @Coduz. I can try to do this, but all dialogs will have auto height. Is that ok?  
_Regards._

Coduz: Hi @ct-anaalbic ,

well I don't actually know what are the consequences of that! 
Could you please try and see if that makes everything else ok, or it causes problems? 

If the second, we can just keep things as they are now.

Regards,

_Alberto_

ct-anaalbic: I tried to put this two line of codes into parent class, but that is not ok, because all dialogs have auto height, and problem with shadow is back, also. I think the best solution is to put this code into proper dialog, like above.

_Regards._

LeoNerdoG:
I suggest we close this PR since it does not solve issue, but only postpones its solution (limiting item names to 50 characters). We should have a more serious talk about this and how these types of bugs should be solved (in REST API also in mind).

muros-ct:
I agree with @LeoNerdoG. Some things, just can't be fixed in GWT without complete rework of a component, as dialog popup in this case.

PR #2491, opened 2019-03-11
The liquibase client is moved to the kapua-commons package.

andrejn-ct:
## Liquibase client move
The liquibase client package was moved from the service module to the kapua-commons module. The liquibase client is more of a helper class so the new location makes more sense.

### Special note
There is no functional change. The existing code was just moved into a new folder.


andrejn-ct:
Still some open issues.

Issue #2489, opened 2019-03-11
Should Filter Fields Be Refreshed After User Input? 


LeoNerdoG:
I would like input on this issue since this is more a proposal than actual issue (everything works fine).

If user (for example) searches for some specific items in Kapua console (e.g. tags, roles, groups...) and inbetween decides to add an item without reseting filter fields, changes in main item grid are not visible due to entries in Fitler fields (see testflow bellow for detailed explanation).
So for now user has to first reset filter fields otherwise it will look like nothing new has been added. 

**Testflow:**
1. Login as kapua-sys or any other user that has proper permissions
2. Go to Tags/Roles/Access Groups/... and add several items (e.g. item1, item2, item3)
3. in Filter section enter some paramteres so Kapua finds some items that have been previously added
4. Without pressing Reset button, try to add several new items and observe the list
**Because filter is still act4ive it looks like that nothing has been added.**

**Expected behavior**
I think that reseting filter fields could be a good idea when adding new items to the grid - what do you others think? 
I think this does not have a negative impact on the console behaciour but adds a feature that can save end user some time and frustration 😄 

**Screenshot No. 1:**
![Screenshot 2019-03-11 at 09 29 10](https://user-images.githubusercontent.com/15121999/54109940-59cf3d80-43e0-11e9-8584-86933b9a7c71.png)

**Screenshot No. 2:**
![Screenshot 2019-03-11 at 09 29 15](https://user-images.githubusercontent.com/15121999/54109935-58057a00-43e0-11e9-9475-f622cd0eeced.png)

**Version of Kapua**
1.1.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Openshift (in its variants) 
[ ] Others

**Main component affected**
[x] Console (in case of console please report info on which browser you encountered the problem)
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 72.0.3626.119 (Official Build) (64-bit))
- Safari (Version 12.0.3 (14606.4.5))
- Mozilla Firefox (65.0.1 (64-bit))
- Opera (Version:58.0.3135.79)

LeoNerdoG:
I am closing this due to discussion in PR #2493.

PR #2493, opened 2019-03-12
reset filter fields after user input

ct-goranpalibrk:

Brief description of the PR.
Filter fields reset when 
**Related Issue**
This PR fixes issue #2489 

Just called reset filter fields method in EntityCRUDToolbar after rendering KapuaDialog.

**Screenshots**


LeoNerdoG:
I have verified that this PR successfully resolves issue #2489.

LeoNerdoG:
@Coduz Edit dialog should not be covered anyway since is user is editing something, nothing new is added and therefore the total number of items is not affected.
With the first point I also disagree, but we can change this so filter fields are not affected if user clicks "cancel".

Coduz:
Hi @LeoNerdoG ,

when editing a dialog and the result set is filtered, the edit could affect one of the filtered attributes, right? So then it should be removed from the result set, shouldn't it? 

Regards, 

_Alberto_

LeoNerdoG:
@Coduz you are right. 
Should we close this then without merge?

Coduz:
If after an edit or add of entity the grid is refreshed with the current filters, I think that is ok, and we can close this

muros-ct:
So as we agreed on today's skype call, we are closing this PR, and leave the things as they were.

Issue #1991, opened 2018-09-11
Problem With Checkboxes


LeoNerdoG:
If user ticks all the checkboxes in Jobs -> Targets or in dataview, and after that presses Space key, only the last checkbox gets deselected/selected instead of all of them. If user uses mouse to check/uncheck the field, everything works fine.

**Testflow:**
1. Login as kapua-sys or any other user that has proper permissions
2. Go to Devices, create at least 5 devices with unique names
3. Go to Jobs -> Targets and select "Selected devices only". 
4. Click on checkbox that selects all the targets. After that press Space key. 
**Observe how Kapua deselects only one of the targets.**
5. Connect a live Kura device to Kapua with installed Example publisher
6. Go to Data view, select Example Publisher under Topics and tick the checkbox for selecting all the metrics. After that press Space key. 
**Again observe how Kapua deselects only one of the targets.**

**Expected behavior**
Behaviour should be the same if user uses mouse to click on "select all" check box or the Space key. 

**Screenshot No. 1 (Selecting all the options):**
![screen shot 2018-09-11 at 12 46 40](https://user-images.githubusercontent.com/15121999/45355455-c21f9c80-b5c0-11e8-8634-400cd2d04751.png)

**Screenshot No. 2 (When using Space key to deselect all the items, only last one gets deselected):**
![screen shot 2018-09-11 at 12 46 44](https://user-images.githubusercontent.com/15121999/45355460-c3e96000-b5c0-11e8-9abf-fbb4d13379b9.png)

**Version of Kapua**
1.0.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Openshift (in its variants) 
[ ] Others

**Main component affected**
[x] Console (in case of console please report info on which browser you encountered the problem)
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 68.0.3440.106 (Official Build) (64-bit))
- Safari (Version 11.1.2 (13604.3.5))
- Mozilla Firefox (61.0.2 (64-bit))
- Opera (Version 55.0.2994.44)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #2523, opened 2019-03-22
Changed functionality of check boxes in data and job targets

ct-anaalbic:

Brief description of the PR.
Changed functionality of check boxes in data tab and job targets tab.

**Related Issue**
This PR fixes issue #1991. 

In Data tab -> Metrics grid, and in Jobs -> Targets add dialog, functionality of check boxes is changed. Event _onKeyUp_ was added on this check boxes for _Enter_ keyboard key. Problem in original issue was _Space_ key, but this key has his own function which is defined in GWT class, so in this PR event is added for Enter keyboard key.

**Screenshots**
/

/


LeoNerdoG:
This PR does not solve issue #1991 directly but as Ana has described above -  "Space" key has its own function in GWT class so we have concentrated more on the Enter key. 
For now, the "Enter" key de/selects all items in the grid and "Space" key de/selects only the last one, which is actually better, because we have two different keys with two different functions (rather than having two different keys that do the same thing).

Coduz:
So is this ok or not?

LeoNerdoG:
This is OK.

PR #2524, opened 2019-03-22
Device job deploy v2 improvements

Coduz:
This PR introduces a bunch of improvements for the Job Engine to work properly with async package management service.

**Related Issue**
_None_

Now the NotificationMessage listener checks also for job to be restarted for a target that has completed the Package install

**Screenshots**
_None_

Javadoc and small code improvements applied.


stefanomorson:
After reviewing the work in this PR a possible improvement emerged. It is described here below.

When a long running operation is executed, there is a limited possibility that some unsolicited messages sent by the device in order to notify the progress are delivered before the operation execution returns.
As a consequence, if the operation was invoked by a device job, the notification handler is not able to address the correct job in order to update the status and trigger execution of the next job steps.
This is currently solved in the job business logic by passing a unique id as a parameter for
the operation execution. While this works it exposes the client code to more complexity
that is related to implementation details.

Instead of passing the id in, the job should wait until the execution of the operation returns passing back the id of the operation registry to the job. Possible notifications received from the device while the operation execution was in progress, are already logged and persisted in the operation registry notification table so they can be used by the job to recover.
After the operation execution returns, the job should use the id of the operation registry to retrieve the notifications already collected from the operation registry notification table and use them to update the job: update the progress and trigger execution of the next job steps (e.g. enqueuing a job execution).

PR #2527, opened 2019-03-25
Logout deleted user

### LINKED ISSUE #2454: "NULL" Error If User Deleted
If user is accidentally or deliberatelly deleted from database, kapua returns "null" error if this user wants to perform actions (look at the screenshot). I suggest we change the "null" nessage to something more user-friendly (if it is possible).

**Testflow:**
2. Go to Users, create a user (e.g. user123) 
3. Select user123 and add User:Read permisson to it
4. From a different browser login as user123
5. From first browser delete user123
6. In second browser try to click User tab or anything else - observe the "null" error message.

**Expected behavior**
I suggest (if it is possible) to handle this exception a little more gracefully (e.g. "This user could not be found, please refresh browser window" or something similar).

**Screenshot:**
screenshot 2019-02-25 at 14 59 00

**Version of Kapua**
1.1.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Others

**Main component affected**
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 

muros-ct:
Changed PR by Ana - PR 2458

**Related Issue**
This PR fixes #2454, but differently.

This solution is a bit different than Ana's. It takes the functionality of logging user out when it is deleted and can not perform any more UI operations.

**Screenshots**
Not applicable.

### DIFF
--- a/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/token/shiro/AccessTokenServiceImpl.java
+++ b/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/token/shiro/AccessTokenServiceImpl.java
@@ -226,9 +226,13 @@ public class AccessTokenServiceImpl extends AbstractKapuaService implements Acce
         // Do find
         entityManagerSession.onTransactedResult(em -> {
             AccessToken accessToken = AccessTokenDAO.find(em, scopeId, accessTokenId);
-            accessToken.setInvalidatedOn(new Date());
+            if (accessToken != null) {
+                accessToken.setInvalidatedOn(new Date());
 
-            return AccessTokenDAO.update(em, accessToken);
+                return AccessTokenDAO.update(em, accessToken);
+            } else {
+                throw new KapuaEntityNotFoundException(AccessToken.TYPE, scopeId);
+            }
         });
     }
 
--- a/console/core/src/main/java/org/eclipse/kapua/app/console/core/client/ChangePasswordDialog.java
+++ b/console/core/src/main/java/org/eclipse/kapua/app/console/core/client/ChangePasswordDialog.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2018 Eurotech and/or its affiliates and others
+ * Copyright (c) 2017, 2019 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
@@ -11,6 +11,7 @@
  *******************************************************************************/
 package org.eclipse.kapua.app.console.core.client;
 
+import org.eclipse.kapua.app.console.core.client.util.Logout;
 import org.eclipse.kapua.app.console.module.api.client.GwtKapuaErrorCode;
 import org.eclipse.kapua.app.console.module.api.client.GwtKapuaException;
 import org.eclipse.kapua.app.console.module.api.client.resources.icons.KapuaIcon;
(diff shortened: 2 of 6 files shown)


muros-ct:
@LeoNerdoG please evaluate this solution. It is different, but I think this is the correct behaviour of UI when user is logged out.

LeoNerdoG:
Hi Uros,
I have checked this PR and found only one issue, otherwise everything is working as expected.
But problem is if user has been deleted from DB and in different browser this user wants to change its password - in this case Kapua does not report "you are not logged in" but "Error while changin password: NULL"

**Screenshot:**
Screenshot 2019-03-26 at 09 48 27

Can you please check this?

lorthirk:
Can we please have a more descriptive name on this PR?

muros-ct:
Sure, I have to change name. It started with fixes to Ana's PR.

muros-ct:
@LeoNerdoG check it out, it works also for deleted user changing password.

LeoNerdoG:
Ok just one last thing - it is not critical or whatsoever but it is an interesting error. 
If user is deleted, then Kapua logs you out. And if you then want to login with any other user (that actually exists), Kapua returns error message "Invalid username or password" - can you please check this?

All later attempts are "valid"and Kapua successfully logs you in.

Screenshot 2019-03-27 at 08 47 35

LeoNerdoG:
Last commit fixed the upper issue so this can now be merged.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/token/shiro/AccessTokenServiceImpl.java
```diff
@@ -226,9 +226,13 @@ public void invalidate(KapuaId scopeId, KapuaId accessTokenId) throws KapuaExcep
...
         entityManagerSession.onTransactedResult(em -> {
             AccessToken accessToken = AccessTokenDAO.find(em, scopeId, accessTokenId);
-            accessToken.setInvalidatedOn(new Date());
+            if (accessToken != null) {
+                accessToken.setInvalidatedOn(new Date());
 
-            return AccessTokenDAO.update(em, accessToken);
+                return AccessTokenDAO.update(em, accessToken);
+            } else {
+                return null;
```
Coduz: Hi @muros-ct ,

why not throw KapuaNotFoundException?

Regards,

_Alberto_

muros-ct: Instead of returning null you mean. Like:

`throw new KapuaEntityNotFoundException(AccessToken.TYPE, scopeId);`

The problem might be that AccessTokenServiceImpl.invalidate() is called from logout, and if that one receives exception it doesn't complete the logout and next time user tries to login it is denied, as it is effectively still logged in. After first failed attempt it can than login successfully, as it is logged out by unsuccessful login. Complicated I know :)

Coduz: I would throw the `KapuaEntityNotFoundException` and catch it in the caller:

AuthenticationServiceShiroImpl.logout();
AuthenticationServiceShiroImpl.refreshAccessToken();

and manage in there the exception.\

Would it make more sense? 

Regards, 

_Alberto_

muros-ct: Sure, I will give it a try. I think it should work.

muros-ct: @Coduz I commited changes.

LeoNerdoG:
I have rechecked this and it is still OK.

muros-ct:
I had to force push twice because there i accidentally removed Ana's sign-off while squashing commits.

Issue #2509, opened 2019-03-18
Broken Columns In Jobs And Child Accounts


LeoNerdoG:
If user clicks on "Definition Name" in Jobs or on "Created By" column in Child Accounts -> Users, bottom views get stuc in "Loading..." screen.

**Testflow:**
1. Login as kapua-sys or any other user that has proper permissions
2. Go to Jobs and create a job (e.g. job0)
3. Select job0 and go to its Steps; Click on "Definition Name" column
4. Go to Child Accounts, create an account (e.g. account0) 
5. Select account0 and go to Users tab, click on "Created By" column

**Expected behavior**
Bottom view should not get disabled if user click on "Definition Name" column in Jobs or "Created By" column in Child Accounts -> Users

**Screenshot No. 1:**
![Screenshot 2019-03-18 at 13 48 30](https://user-images.githubusercontent.com/15121999/54533095-ae972900-4989-11e9-9c65-6395dc4e093a.png)

**Screenshot No. 2:**
![Screenshot 2019-03-18 at 14 25 11](https://user-images.githubusercontent.com/15121999/54533096-b060ec80-4989-11e9-9961-3de3c1cae161.png)

**Version of Kapua**
1.1.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Openshift (in its variants) 
[ ] Others

**Main component affected**
[x] Console (in case of console please report info on which browser you encountered the problem)
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 72.0.3626.119 (Official Build) (64-bit))
- Safari (Version 12.0.3 (14606.4.5))
- Mozilla Firefox (65.0.1 (64-bit))
- Opera (Version:58.0.3135.79)

LeoNerdoG:
I have verified that this issue has been resolved - the columns are working again.

LeoNerdoG:
I am reopnening this since issue #2519 has been reverted.

LeoNerdoG:
I have verified that this issue has been resolved.

PR #2531, opened 2019-03-27
Enabled sorting of columns in grids

ct-anaalbic:

Brief description of the PR.
Changed sort option in columns on child user grid and job steps grids.

**Related Issue**
This PR fixes issue #2509.

Sort option in columns _"Created By"_ in child users, and _"Definition Name"_ in job steps is changed. Also, the proper grid view is changed to not be disabled if user click on _"Definition Name"_ column in Jobs -> Steps or _"Created By"_ column in Child Accounts -> Users.

**Screenshots**
/

/


LeoNerdoG:
I have verified that from UX standpoint everything is OK now. Can one of commiters check the code and merge if it is OK?

Coduz:
Code right now fixes the behaviour of the column.

Keep in mind that the sorting on `*Id` fields that are translated to user-friendly names will never be accurate, since the sorting of the `Id` is not the same at the sorting of the user-friendly name of that id.

Issue #1620, opened 2018-04-20
"Login" Button Looks Disabled if User Logs Out


LeoNerdoG:
If a user logs out, the **Login** button looks disabled, although it operates normaly. If user for example changes anything (username or password), button is shown as active. 

Testflow: 
1. Login as kapua-sys
2. After you login, logout (upper right corner -> logout)
**After you logout, Kapua should bring you to login screen. Look at the "Login" button - it looks disabled as if user did not entere username or password, which is not the case here.**

Screenshot:
![screen shot 2018-04-20 at 10 03 51](https://user-images.githubusercontent.com/15121999/39037885-26462c98-4482-11e8-807f-40f4e5eac181.png)

Browsers used: 
- Google Chrome (Version 65.0.3325.181 (Official Build) (64-bit)
- Safari (Version 11.1 (13604.3.5))
- Mozilla Firefox (59.0.2  (64-bit))
- Opera (Version 52.0.2871.64)

Coduz:
Yeah I confirm that this is due to the fact that GWT does not "sense" the autocompletion from the browser. A simple TAB hit enables the buttons.

LeoNerdoG:
Hi @ct-anaalbic,
I have found another problem with you fix; If user first enters unexisting user (e.g. 123/123), clicks "Login" and after that enters only username (without password) and click "Login", Kapua does not mark "Password" field as invalid, although it should (it is empty). 

Can you please take a look? 😺 

![Screenshot 2019-04-02 at 08 38 12](https://user-images.githubusercontent.com/15121999/55383378-55ea9300-5528-11e9-9a05-7ae26609e1ab.png)

ct-anaalbic:
Hi @LeoNerdoG ,

I will check this now. :)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #2533, opened 2019-03-28
enabled login button when user logs out

ct-anaalbic:

Brief description of the PR.
Enabled login button when user logs out.

**Related Issue**
This PR fixes issue #1620.

In this issue, problem is disabled Login button when user logs out, and his credentials are saved in browser which means that fields on login dialog are auto completed. This problem is caused by GWT, and every event, like _TAB_ click on keyboard, activates this button. Because of that, in this PR logic on login page is changed. 
First, Login button is enabled in every case. After that, some validation are added. If username and password fields on dialog are empty, proper error message is shown in the bottom right corner, and fields are marked invalid. Also, `onFocus` event is removed from reset button, because fields have to be in error state always when they are not filled. 
On this way, issue is resolved and button field is not disabled after log out, and, also, functionality is not changed or disturbed.

**Screenshots**
/

/

### DIFF
--- a/console/core/src/main/java/org/eclipse/kapua/app/console/core/client/LoginDialog.java
+++ b/console/core/src/main/java/org/eclipse/kapua/app/console/core/client/LoginDialog.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2018 Eurotech and/or its affiliates and others
+ * Copyright (c) 2017, 2019 Eurotech and/or its affiliates and others
  *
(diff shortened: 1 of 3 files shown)


LeoNerdoG:
I have verified that this PR successfully resolves issue #1620. 
@Coduz @lorthirk can you please check if the solution is OK? 
From GUI standpoint I think it is OK - at least I have not found anything.

muros-ct:
@lorthirk What do you think about this solution? It is a solution that assumes that login button is always enabled.

PR #2534, opened 2019-03-29
Fix for Created by/Modified by properties not shown correctly

### LINKED ISSUE #2465: "N/A" Instead "kapua-sys"In Endpoints And In Settings
If kapua-sys user creates an Endpoint in a child account, under "Created by" there is "N/A" but it should be "kapua-sys" since kapua-sys is absolute admin user and should be able to see everything. 
In bottom Description part there is "kapua-sys" so problem is only in "Created by" column. 
The second problem is that there is "N/A" in Settings which is again not ok (look at second testflow). 

The third problem is that there is "N/A" at Expiration date - it should be "Never". 

**Testflow:**
1. Login as kapua-sys
2. Go to Child Accounts, create an account (e.g. account0)
3. Switch to account0 and go to Endpoints and create an endpoint
**Observe the "Created by" column.**

**Expected behavior**
There should not be "N/A" in Created By column, but "kapua-sys".

**Screenshot:**
screenshot 2019-02-27 at 08 17 59

**Testflow for Settings:**
1. Login as kapua-sys
2. Go to Child Accounts, create an account (e.g. account0)
3. Switch to account0 and go to its Settings
**Observe the "Created by"  and "Modified By" parameters.**

**Expected behavior**
There should not be "N/A" in Created By column, but "kapua-sys".

**Screenshot:**
screenshot 2019-02-27 at 14 08 47

**Version of Kapua**
1.1.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Others

**Main component affected**
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 

ct-ajovanovic:

Brief description of the PR.
Fix for Created by/Modified by properties not shown correctly.

**Related Issue**
This PR fixes/closes #2465 

Added new parameter `currentSession.getAccountId()` to:
- `GwtAccountService(Impl)` class's `getAccountInfo()` method
- `GwtEndpointService(Impl)` class's `query()`  and `getEndpointDescription()` methods
That additional parameter is used when querying Users for the needed Created By/Modified By properties.

Overridden the `renderValueCell()` method in `AccountDetailsTabDescription` regarding the display of _Expiration Date_ property.

**Screenshots**
_None_

_None_

### DIFF
--- a/console/module/account/src/main/java/org/eclipse/kapua/app/console/module/account/server/GwtAccountServiceImpl.java
+++ b/console/module/account/src/main/java/org/eclipse/kapua/app/console/module/account/server/GwtAccountServiceImpl.java
@@ -203,27 +203,28 @@ public class GwtAccountServiceImpl extends KapuaRemoteServiceServlet implements 
     }
 
     @Override
-    public ListLoadResult<GwtGroupedNVPair> getAccountInfo(String scopeIdString, String accountIdString)
+    public ListLoadResult<GwtGroupedNVPair> getAccountInfo(String scopeIdString, String accountIdString, String selectedAccountIdString)
             throws GwtKapuaException {
         final KapuaId scopeId = KapuaEid.parseCompactId(scopeIdString);
-        KapuaId accountId = KapuaEid.parseCompactId(accountIdString);
+        final KapuaId accountId = KapuaEid.parseCompactId(accountIdString);
+        final KapuaId selectedAccountId = KapuaEid.parseCompactId(selectedAccountIdString);
 
         List<GwtGroupedNVPair> accountPropertiesPairs = new ArrayList<GwtGroupedNVPair>();
         try {
-            final Account account = ACCOUNT_SERVICE.find(scopeId, accountId);
+            final Account account = ACCOUNT_SERVICE.find(scopeId, selectedAccountId);
 
             User userCreatedBy = KapuaSecurityUtils.doPrivileged(new Callable<User>() {
 
                 @Override
                 public User call() throws Exception {
-                    return USER_SERVICE.find(scopeId, account.getCreatedBy());
+                    return USER_SERVICE.find(accountId, account.getCreatedBy());
                 }
             });
             User userModifiedBy = KapuaSecurityUtils.doPrivileged(new Callable<User>() {
 
                 @Override
                 public User call() throws Exception {
-                    return USER_SERVICE.find(scopeId, account.getModifiedBy());
+                    return USER_SERVICE.find(accountId, account.getModifiedBy());
                 }
             });
 
--- a/console/module/account/src/main/java/org/eclipse/kapua/app/console/module/account/client/AccountDescriptionTab.java
+++ b/console/module/account/src/main/java/org/eclipse/kapua/app/console/module/account/client/AccountDescriptionTab.java
@@ -45,7 +45,7 @@ public class AccountDescriptionTab extends EntityDescriptionTabItem<GwtAccount> 
 
             @Override
             protected void load(Object loadConfig, AsyncCallback<ListLoadResult<GwtGroupedNVPair>> callback) {
-                GWT_ACCOUNT_SERVICE.getAccountInfo(currentSession.getSelectedAccountId(), selectedEntity.getId(), callback);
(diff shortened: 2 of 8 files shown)


ct-ajovanovic:
Hi @Coduz 
Can you please take a look at this PR? Is this solution ok?
Thanks
_Aleksandra_

LeoNerdoG:
I have verified that this PR successfully resolves issue #2465. I have only one questin regarding visibility of parameter "Modified By" (general in kapua).
If parent user (e.g. admin-user0) creates several child accounts and some users in them (e.g. account0, account1, user0, user1) and then logs in as user0 and user1 (wihin account0 and account1 respectively) and change some parameters (e.g. descriptions in Tags or creates Schedules in jobs,...), admin-user0 will see only "N/A" in "Modified By"  parameters instead of actual users that made chamges to this field - is this OK or not? 
So if user0 changes something in account0, admin-user0 will see "N/A" in "Modified By" columns of these items instead user0.

I would expect that admin user (top user of all the child accounts and users) should be able to see all the parameteres of all accounts and all users - or am I wrong? 

i hope this question is not too confuzing - otherwise please leave a comment! 😄

Coduz:
>I would expect that admin user (top user of all the child accounts and users) should be able to see all the parameteres of all accounts and all users - or am I wrong?

I agree with this. "admin-user0" should be able to see all "Created By" and "Modified By" of all of its sub-accounts

ct-ajovanovic:
Hi @Coduz 
I can't seem to find a solution for the problem @LeoNerdoG mentioned above. Can you please help?
The problem is in the `GWT_ACCOUNT_SERVICE.getAccountInfo()` method and it's call to the  `USER_SERVICE.find()` method when setting the created by/modified by properties.
I can see two ways of setting the `accountId` parameter in the `USER_SERVICE.find(accountId, userId)` method:
1) currentSession.getSelectedAccountId() <- previous solution
2) currentSession.getAccountId() <- this PR

It the first case, as mentioned in #2465 we have an issue where _parent user_/ _kapua-sys_ sees N/A value for Created By/ Modified By properties of the child account he himself created/modified.
The second case is my idea for solving the above mentioned problem, and covers most cases. Except the one @LeoNerdoG mentioned. If the user from a child account updates any of its properties, the name of that user will not be visible to the _parent user_/_kapua-sys_. 
What do you think? How can we solve this?
Thanks
_Aleksandra_

Coduz:
Hi all,

the current solution and the proposed one have the same problem, that is to not being able to find the right user when the `modifiedBy` and `createdBy` are not related to the `scopeId` of the entity nor the current logged user's `scopeId`.

The correct solution would be to find the `modifiedBy` and `createdBy` only by their `id` without given the `scopeId`. By doing so, the `.checkAccess` method will determine if the `scopeId` of the found user matches the permissions.

Regards, 

_Alberto_

console/module/account/src/main/java/org/eclipse/kapua/app/console/module/account/server/GwtAccountServiceImpl.java
```diff
@@ -203,27 +203,28 @@ public GwtAccount find(String accountIdString)
...
         try {
-            final Account account = ACCOUNT_SERVICE.find(scopeId, accountId);
+            final Account account = ACCOUNT_SERVICE.find(scopeId, selectedAccountId);
 
             User userCreatedBy = KapuaSecurityUtils.doPrivileged(new Callable<User>() {
 
                 @Override
                 public User call() throws Exception {
-                    return USER_SERVICE.find(scopeId, account.getCreatedBy());
+                    return USER_SERVICE.find(accountId, account.getCreatedBy());
```
Coduz: This still does not work if the currently logger user's account is a parent of the currently selected account, does it?

muros-ct: Leonardo and I are working on this. We have another solution, I will create PR tomorrow.

ct-ajovanovic: @Coduz Yes, this solution still doesn't cover the case Leonardo mentioned - if the user from the child account makes a change the Created by/ Modified user information is not visible to the parent account.

muros-ct:
This PR is being closed, as it was replaced by PR #2554, which resolved an issue #2465 on a bit different way. See PR #2554.

PR #2539, opened 2019-04-02
Changed password field on login dialog to be in error state

ct-anaalbic:

Brief description of the PR.
Changed password field on login dialog to be invalid if is empty.

**Related Issue**
This PR is continuation of PR 2533.

After invalid username or password input on login dialog, if user enters something only in username field and password field is empty, password field is changed to bi in error state.

**Screenshots**
/

/


LeoNerdoG:
When re-checking issue #1620 I found another case that has not been covered in PR #2533. 
This PR covers also this case so when this is merged all of the cases should be covered. 
can one of the commiters check the code and merge please?

Issue #1964, opened 2018-08-31
Password And Username Fields In Error State If Reset Button Pressed


LeoNerdoG:
If user focuses (clicks on) Password field in the Kapua login screen and then presses Reset, Password field gets in error state. After the button is released, the error state is gone. 

Also if user clicks on Username field and does not enter anything (and instead presses Reset button), Username field is in Error state, but it should not be. 

**Testflow:**
1) When prompted for username and password at Kapua login page, click on Password field
2) When focus is on Password field, click on Reset Button.
**Observe how Password field state when Reset button is pressed and when is released.**

**Expected behavior**
If Password field is empty and Reset button is pressed, Password field should not get into Reset state. 

**Screenshot No. 1:**
![screen shot 2018-08-31 at 12 49 44](https://user-images.githubusercontent.com/15121999/44908615-5e71b580-ad1c-11e8-93d9-043ca36f0aa8.png)

**Version of Kapua**
1.0.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Openshift (in its variants) 
[ ] Others

**Main component affected**
[x] Console (in case of console please report info on which browser you encountered the problem)
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 68.0.3440.106 (Official Build) (64-bit))
- Safari (Version 11.1.2 (13604.3.5))
- Mozilla Firefox (61.0.2 (64-bit))
- Opera (Version 55.0.2994.44)

ct-anaalbic:
Hi @lorthirk. 
I resolved this problem and my solution works on some browsers, like Opera and Chrome, but in some other browsers it not works. 
First solution is to remove `focus` from username field after reset fields. On this way, fields are not in error state in every browsers. 
Second solution is to disable reset button if username and password fields are not filled. 
Can you tell me which of these two solutions are better, and are any of them ok? 

Thank you.
_Regards_,
Ana

lorthirk:
The latter sounds most promising. There's no sense in having the Reset button enabled if the fields are empty. If this covers all the cases, then go for this one. Thanks!

LeoNerdoG:
I ahhve verified that this issue has been resolved.

PR #2540, opened 2019-04-04
Disabled reset button if fields are empty

ct-anaalbic:

Brief description of the PR.
Disabled reset button on login dialog if fields are not filled.

**Related Issue**
This PR fixes issue #1964.

If username and password fields are empty on login dialog, reset button is disabled. Only if user enters some characters in both of these fields, reset button will be enabled.

**Screenshots**
/

/


LeoNerdoG:
I have verified that this PR successfully resolves issue #1964. @lorthirk we have implemented the solution we have been talking about (disabled Reset button if Username and Password fields are not filled ) and it looks OK to me on GUI side, so from my perspective this can be merged.

PR #2549, opened 2019-04-09
fix service events entity scope id (see issue #2547)

### LINKED ISSUE #2547: Service Events: missing entity scope id.
**Describe the bug**
The Service Event scope id is the scope id of the user that triggered the event. In some case (i.e. when the admin user set/modify some entity outside his scope) this value doesn't reflect the entity scope id preventing the possibility to retrieve the original entity that triggered the event.
So adding a new field to track the entity scope id it's a good solution to allow event consumers to retrieve the entity that originated the event.
The filed name can be something like entityScopeId.

**To Reproduce**
see description

**Expected behavior**
see description

**Screenshots**
none

**Version of Kapua**
all

**Type of deployment**
all

**Main component affected**
all components interract with Service Events

**Additional context**
none

riccardomodanese:

proposed solution for #2547
@stefanomorson @Coduz @lorthirk can you take a look to the solution proposed and give me a feedback.
Can be acceptable or do you suggest another way (or an integration) to solve the issue?

**Related Issue**
fix #2547 

just get the scope id using the same logic implemented to get the entity id.

**Screenshots**
none

none

### DIFF
--- a/commons/src/main/java/org/eclipse/kapua/commons/event/RaiseServiceEventInterceptor.java
+++ b/commons/src/main/java/org/eclipse/kapua/commons/event/RaiseServiceEventInterceptor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017 Eurotech and/or its affiliates and others
+ * Copyright (c) 2017, 2019 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
@@ -14,6 +14,8 @@ package org.eclipse.kapua.commons.event;
 import org.aopalliance.intercept.MethodInterceptor;
 import org.aopalliance.intercept.MethodInvocation;
 import org.eclipse.kapua.commons.core.InterceptorBind;
+import org.eclipse.kapua.commons.metric.MetricServiceFactory;
+import org.eclipse.kapua.commons.metric.MetricsService;
 import org.eclipse.kapua.commons.model.id.KapuaEid;
 import org.eclipse.kapua.commons.security.KapuaSecurityUtils;
 import org.eclipse.kapua.commons.security.KapuaSession;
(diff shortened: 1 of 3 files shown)


riccardomodanese:
Can we add some metrics to track the "unclear" condition (such as more than one KapuaEntity found or more than 2 KapuaId found?)

Coduz:
@riccardomodanese what would be te purpose?

Coduz:
No I meant regarding this

> Can we add some metrics to track the "unclear" condition (such as more than one KapuaEntity found or more than 2 KapuaId found?)

commons/src/main/java/org/eclipse/kapua/commons/event/RaiseServiceEventInterceptor.java
```diff
@@ -158,14 +148,80 @@ private void fillEvent(MethodInvocation invocation, ServiceEvent serviceEvent) {
         }
     }
 
+    private void useEntityToFillEvent(ServiceEvent serviceEvent, List<KapuaEntity> entities) {
+        if (entities.size()>1) {
+            LOG.warn("Found more than one KapuaEntity in the parameters! Assuming to use the first one!");
```
riccardomodanese: @Coduz I mean add hee a metric to keep track of such dangerous behaviour

commons/src/main/java/org/eclipse/kapua/commons/event/RaiseServiceEventInterceptor.java
```diff
@@ -158,14 +148,80 @@ private void fillEvent(MethodInvocation invocation, ServiceEvent serviceEvent) {
...
+        KapuaEntity entity = entities.get(0);
+        serviceEvent.setEntityType(entity.getClass().getName());
+        serviceEvent.setEntityId(entity.getId());
+        serviceEvent.setEntityScopeId(entity.getScopeId());
+        LOG.info("Entity '{}' with id '{}' and scopeId '{}' found!", entity.getClass().getName(), entity.getId(), entity.getScopeId());
+    }
+
+    private void useKapuaIdsToFillEvent(ServiceEvent serviceEvent, List<KapuaId> ids, Class<?>[] implementedClass) {
+        if (ids.size()>2) {
+            LOG.warn("Found more than two KapuaId in the parameters! Assuming to use the first two!");
```
riccardomodanese: @Coduz I mean add hee a metric to keep track of such dangerous behaviour

riccardomodanese:
@Coduz  I commented the code

Coduz:
The case in which that could happen could be only a bug on the code, not from a special contition when running the application, right?

riccardomodanese:
No it's a normal condition if for example, the event is triggered by a code block inside a doPrivileged call.
I think could be a very common case.

PR #2554, opened 2019-04-10
Display of modified and created by user fields

muros-ct:
This solves issue with display of created by and modified by fields on multiple screens.
Issue is shown when using sub account and not kapua-sys account.

**Related Issue**
This is one way that would be possible to solve issue #2465.
Aleksandra tried with her PR #2534.

This approach is a bit more aggressive on user service, where it uses
KapuaId.ANY in query() methods to find all records.
Of course if permission check is executed before, user has to have permission
ANY to execute such a query.
In case of doPrivileged() this also shouldn't be a problem as it is
privileged after all.

Fixed created by / modified by on all grids and details.

Big change was done on DAO query itself. It is now possible to do query on all items by using KapuaId.ANY. **Is this a possible issue**? I think not, but would be glad to discuss.

**Screenshots**
Not applicable.

Review of DAO changes.

### DIFF
--- a/commons/src/main/java/org/eclipse/kapua/commons/service/internal/ServiceDAO.java
+++ b/commons/src/main/java/org/eclipse/kapua/commons/service/internal/ServiceDAO.java
@@ -357,7 +357,7 @@ public class ServiceDAO {
         //
         // WHERE
         QueryPredicate kapuaPredicates = kapuaQuery.getPredicate();
-        if (kapuaQuery.getScopeId() != null) {
+        if ((kapuaQuery.getScopeId() != null) && (!kapuaQuery.getScopeId().equals(KapuaId.ANY))) {
 
             AndPredicate scopedAndPredicate = kapuaQuery.andPredicate(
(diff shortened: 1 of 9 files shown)


muros-ct:
Hi @Coduz, can you give a comment on this solution or propose the one we discussed on friday skype call, about replacing this one with null check omission on accountId.

Coduz:
Hi @muros-ct ,

I saw the changes and using the `.query` to find the user without filtering by `scopeId` could work to make things work. 

ServiceDAO was already supporting the `null` as a `scopeId` value to not filter by `scopeId`. Maybe some services still check for `query.getScopeId() != null` but those can be removed. 

If we are going with this solution instead, please find all usages of `Factory.newQuery(null)` and replace with your approach and re-introduce the check on `query.getScopeID != null`. I would like to have only one of the two approach to be allowed.

Which one you prefer? 

Regards, 

_Alberto_

muros-ct:
@Coduz I applied approach that is already used in Kapua. It brings less changes and it works as well.

ct-bbunic:
I have verified that this PR successfully resolves issue #2465.

Issue #2340, opened 2019-01-29
"Write-only" Boolean Asset Value Visible In Kapua 


LeoNerdoG:
If user connects a Kura device to Kapua that has Boolean write-only assets, values of these assets are visible to user, altohugh they should not be, as they are write-only. 

**Testflow:**
1. Login as kapua-sys or any other user that has proper permissions
2. Connect a live Kura device (e.g. device0) to kapua that has Boolen Write-only asset and Integer Write-only asset
3. In Kapua go to Devices, select device0 and go to its Assets. 
4. Observe Write-only assets: integer and all other write-only asset values are hidden while boolean write-only assets are visible.

**Expected behavior**
I think that also boolean write-only asset values should be hidden  as all others. 

**Screenshots**
![screenshot 2019-01-25 at 12 53 37](https://user-images.githubusercontent.com/15121999/51901661-724b4180-23b8-11e9-8f8a-3b200b995be7.png)

**Version of Kapua**
1.1.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Openshift (in its variants) 
[ ] Others

**Main component affected**
[x] Console (in case of console please report info on which browser you encountered the problem)
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 71.0.3578.98 (Official Build) (64-bit))
- Safari (Version 12.0 (14606.3.4))
- Mozilla Firefox (64.0 (64-bit))
- Opera (Version 57.0.3098.110)

lorthirk:
Well... This is no issue at all, IMHO!

You need to have that channel shown in the Console, otherwise you would have no way to write a value in it. Probably the correct workflow interaction would be an empty, writable field that gets cleared as soon as the channel is written, and never read back. But again, the GUI element needs to be there.

LeoNerdoG:
@lorthirk yes, as you said:

> We need to show write only channels to let the user interact with them.

This was done before and was not the case of this issue. With PR #2557 we have fixed only that user does not see the specific value (true or false) as other values from other data types are hidden. 
Asset names themselves are shown but not the values so the PR is ok from my standpoint 😄

lorthirk:
Ok, shame on me for not reading carefully 😄It's fine for me as well!

LeoNerdoG:
I have verified that this issue has been resolved.

PR #2557, opened 2019-04-11
Changed visibility of write boolean assets

ct-anaalbic:

Brief description of the PR.
Visibility of write Boolean assets are changed.

**Related Issue**
This PR fixes issue #2340.

Value of assets is not visible if assets are Boolean and if they only have `write` permission. It looks like on screenshot bellow.

**Screenshots**
image

/

### DIFF
--- a/console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/assets/DeviceAssetsPanel.java
+++ b/console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/assets/DeviceAssetsPanel.java
@@ -369,6 +369,12 @@ public class DeviceAssetsPanel extends LayoutContainer {
             radioFalse.setValue(true);
             radioGroup.setOriginalValue(radioFalse);
         }
+        if(channel.getModeEnum().equals(GwtDeviceAssetChannelMode.WRITE)) {
+            radioTrue.setValue(null);
+            radioGroup.setOriginalValue(radioTrue);
+            radioFalse.setValue(null);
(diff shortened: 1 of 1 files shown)


LeoNerdoG:
I have verified that this PR successfully resolve issue #2340 - can one of the commiters check the code and merge when possible?

console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/assets/DeviceAssetsPanel.java
```diff
@@ -369,6 +369,12 @@ public void handleEvent(BaseEvent be) {
             radioFalse.setValue(true);
             radioGroup.setOriginalValue(radioFalse);
         }
+        if(channel.getModeEnum().equals(GwtDeviceAssetChannelMode.WRITE)) {
+            radioTrue.setValue(null);
```
muros-ct: What means setting value to null?

ct-anaalbic: Hi @muros-ct.

It means if channel is only with write permission, than value in radio button is null, and user can not see value, but he can chose false or true option.

muros-ct: Ok, I understand now. It is disabling radio button functionality. None of buttons is selected, neither true or false. Hack.

PR #2564, opened 2019-04-16
added message for refreshing application when displayname is changed

ct-goranpalibrk:

Brief description of the PR.
Added new message which Informs current user that if his display name is changed, he should refresh application if he wants to update username in top right corner.

**Related Issue**
This PR fixes issue  #1600

Due to inability to trigger refreshing display name when user change display name, @LeoNerdoG  and I have discussed about solution with this message.

**Screenshots**
kapua


ct-goranpalibrk:
Hi @lorthirk do you agree with this solution?

lorthirk:
Honestly I'd go for the refresh only as a last resort. Don't we have a way to reach for the upper menu through an event?

ct-goranpalibrk:
Hi @lorthirk , in PR #1711 it is said that we don't use events. That is the reason why I have tried this solution.

Coduz:
I agree with Claudio. 

There should be a DialogListener that refresh the grid when the AddDialog, the EditDialog or the DeleteDialog are closed, isn't that the case?

ct-goranpalibrk:
@Coduz yes, you are right, but that listener only refreshes grids, not the upper right corner which is different component.

PR #2565, opened 2019-04-16
Added shift of step index when deleting a jobStep

### LINKED ISSUE #2425: "Step Index" Number Not Reset After Steps Are Deleted From Job
If user deletes on of the steps in Jobs, the step index does not get reset, so there can be only one step indexed with "3" which is not ok, if there is only one step. 

**Testflow:**
2. Go to Jobs, create a job (e.g. job0) 
3. Select job0 and go to its Steps, create 3 sample steps
4. Delete steps with index 1 and 2

**Expected behavior**
If step is deleted from a job, the step indexes should be reset (look at the screenshot).

**Screenshot:**
screenshot 2019-02-19 at 14 49 07

**Version of Kapua**
1.1.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Others

**Main component affected**
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 

Coduz:
Implemented the shift of the other JobSteps when deleting a JobStep of a Job

**Related Issue**
This PR fixes #2425 

When deleting a JobStep, now we check if there are other JobStep defined after the one that is being deleted and we are decreasing the JobStep.stepIndex property of 1 

**Screenshots**
_None_

_None_

### DIFF
--- a/service/job/internal/src/main/java/org/eclipse/kapua/service/job/step/internal/JobStepImpl.java
+++ b/service/job/internal/src/main/java/org/eclipse/kapua/service/job/step/internal/JobStepImpl.java
@@ -56,8 +56,8 @@ public class JobStepImpl extends AbstractKapuaNamedEntity implements JobStep {
     private KapuaEid jobStepDefinitionId;
 
     @Basic
-    @Column(name = "step_index", nullable = false, updatable = false)
-    private int stepIndex;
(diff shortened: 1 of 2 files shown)


LeoNerdoG:
I have verified that this PR successfully resolves issue #2425. @coduz you can merge this 😄

Issue #2558, opened 2019-04-12
"Admin" Role In Kapua-sys Account Should Not Be Deletable


LeoNerdoG:
This issue replaces #2392 - the original issue was that if admin role in kapua-sys is deleted, there is no way kapua-sys user can see "Endpoints" in the main menu on the left, even if has ALL:ALL:ALL:YES permission. 
We have decided to close that issue and open this one - Admin role should not be deletable at all. 

**Testflow:**
1. Login as kapua-sys
2. Go to Roles, select Admin role, try to delete it

**Expected behavior**
Admin role in kapua-sys account should not be deletable.

**Screenshots**
/

**Version of Kapua**
1.1.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Openshift (in its variants) 
[ ] Others

**Main component affected**
[x] Console (in case of console please report info on which browser you encountered the problem)
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 72.0.3626.119 (Official Build) (64-bit))
- Safari (Version 12.0.3 (14606.4.5))
- Mozilla Firefox (65.0.1 (64-bit))
- Opera (Version:58.0.3135.79)

LeoNerdoG:
I have verified that this issue has been resolved.

PR #2568, opened 2019-04-19
Admin role can not be deleted

ct-anaalbic:

Brief description of the PR.
Admin role can not be deleted.

**Related Issue**
This PR fixes issue #2558.

User in _kapua-sys_ account can not delete admin role. Instead of that, if user try to do this, he get exception that this action is not allowed.

**Screenshots**
/

/


ct-bbunic:
I have verified that this PR successfully resolves issue #2558.

LeoNerdoG:
@muros-ct @ct-anaalbic,
I have found a problem with this solution - user can still select "admin" role and delete its Permissions - and the end result is the same as if user would delete admin role. 
Can we fix this so neither admin role nor its permission can be deleted? 
I suggest we create another PR and just reference this one and the issue so we keep track.

Coduz:
Hi @LeoNerdoG 

can you please open another issue about that? 

Regards, 

_Alberto_

PR #2569, opened 2019-04-19
Fix for Error/Info boxes stacking

### LINKED ISSUE #1739: Error/Info Boxes Stacking One Upon Another
If user does something, that produces an error (e.g. creating a user with non-unique name) and immediatelly after that does something "legal" (in this case creating  user with unique name) info messages get stacked up one uppon error messages  (see examples bellow) and vice versa. 

Bellow there is a list where and how this issue manifests.
**_Welcome:/_**
**_Connections:/_**
**_Devices:_** 
 - Adding a **device** with non-unique name and after that adding a device with unique name
 - Adding a **device** with non-unique name and after that editing/deleting a device
 - Adding a **tag** with non unique name and after that adding a tag with unique name
 - Deleting a **tag** and after that adding a tag

**_Batch Jobs:_**
 - Adding a **job** with a non-unique name and after that adding a job with a unique name
 - Changing/Editing **job** name to non-unique one and after that changing it to unique one
 - Startin/stopping a **job** an after that deleting it
 - Adding a **step** with non-unique name and after that deleting a step
 - Adding a **step** with non-unique name and after that changing its name to unique one
 - Changing a **step** name to a non-unique one and after that chanign it to unique one
 - Adding a **Schedule** with non-unique name and after that deleting a schedule

**_Tags:_**
- Adding a **tag** witn a non-unique name and after that adding a tag with unique one.
- Changing/Editing a **tag** name to a non-unique one and after that changing it to a unique one
- Changing/Editing a **tag** name from non-unique one to unique and immediatelly after that deleting it

**_Users:_**
 - Adding a **user** with non-unique name and after that adding user with unique name
 - Adding already added **role** to user and after that adding a role, which is not added yet. 
 - Removing a **role** after trying to add a role, which is already added
 - Adding another **API KEY** after trying to add another password to user
 - Deleting a **PASSWORD/API KEY** after trying to add another password to a user
 - Adding a **permission** that is already added, after successfully adding a permission

**_Roles:_**
  - Adding a **role** with unique name after trying to add a role with non-unique name
  - Changing/Editing a **role** name to unique one after trying to change it to non-unique one
  - Deleting a **role** after trying to change its name to a non-unique one
  - Deleting a **role** after trying to add new role with non-unique name

**_Access Groups:_**
  - Adding a **group** with unique name after trying to add a group with non-unique name
  - Changing/Editing a **group** name to unique one after trying to change it to non-unique one
  - Deleting a **group** after trying to change its name to a non-unique one
  - Deleting a **group** after trying to add new group with non-unique name

**_Child Accounts:_**
  - Adding an **account** with unique name after trying to add an account with non-unique name
  - Changing/Editing an **account** name to unique one after trying to change it to non-unique one
  - Deleting an **account** after trying to change its name to a non-unique one
  - Deleting an **account** after trying to add new account with non-unique name
  - Adding a **user** with a non-unique name 
  - Deleting a **user** after trying to add a user with non-unique name
 
**_Endpoints:_**
 - Adding an **endpoint** with unique name/parameters after trying to add an endpoint with non-unique parameters
 - Deleting an **endporint** after trying to add an endpoint with non-unique parameters

**_Settings:/_**
**_About:/_**

This also happens if combination of these commands are used, for example: if user wants to add a group with non unique name and after that goes to Roles and adds a role with unique name.

Testflow: 
1. Logina as kapua-sys or any other user with proper permissions
2. Follow the instructions above for producing the errors described.

Sample Screenshot:
error_boxes_stacking

Browsers used: 
- Google Chrome (Version 65.0.3325.181 (Official Build) (64-bit)
- Mozilla Firefox (59.0.2  (64-bit))
- Opera (Version 52.0.2871.64)

ct-ajovanovic:

Brief description of the PR.
Fix for Error/Info boxes stacking.

**Related Issue**
This PR fixes/closes #1739

`ConsoleInfo` class is now a singleton class. Created `getInstance()` method for accessing the mentioned singleton class. Updated/overridden the affected methods from the extended `Info` class for correct display of Info/Error messages. After these changes only one last message is shown and there is no more stacking.

**Screenshots**
_None_

Set exitStatus to true for the onSuccess() method after AccountAddDialog submit.

### DIFF
--- a/console/module/account/src/main/java/org/eclipse/kapua/app/console/module/account/client/toolbar/AccountAddDialog.java
+++ b/console/module/account/src/main/java/org/eclipse/kapua/app/console/module/account/client/toolbar/AccountAddDialog.java
@@ -354,7 +354,8 @@ public class AccountAddDialog extends EntityAddEditDialog {
 
                     @Override
                     public void onSuccess(GwtAccount account) {
-                        ConsoleInfo.display(MSGS.info(), MSGS.accountCreatedConfirmation());
+                        exitStatus = true;
+                        exitMessage = MSGS.accountCreatedConfirmation();
                         hide();
                     }
                 });
--- a/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/util/ConsoleInfo.java
+++ b/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/util/ConsoleInfo.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017 Eurotech and/or its affiliates and others
+ * Copyright (c) 2017, 2019 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
@@ -11,13 +11,36 @@
  *******************************************************************************/
 package org.eclipse.kapua.app.console.module.api.client.util;
 
+import com.extjs.gxt.ui.client.util.Format;
 import com.extjs.gxt.ui.client.util.Params;
+import com.extjs.gxt.ui.client.util.Point;
 import com.extjs.gxt.ui.client.widget.Info;
 import com.extjs.gxt.ui.client.widget.InfoConfig;
+import com.google.gwt.user.client.Timer;
+import com.google.gwt.user.client.ui.RootPanel;
 
 public class ConsoleInfo extends Info {
     private static final int CHAR_STEP = 20;
     private static final int STEP_HEIGHT = 18;
+
+    //Private static instance of ConsoleInfo singleton class
+    private static final ConsoleInfo SINGLE_INSTANCE = new ConsoleInfo();
+    private static Timer timer;
+
+    /**
+     * Private ConsoleInfo class constructor
+     */
+    private ConsoleInfo() {
+
+    }
+
+    /**
+     * Method for returning ConsoleInfo singleton instance.
+     * @return static instance of ConsoleInfo class
+     */
+    public static ConsoleInfo getInstance() {
+        return SINGLE_INSTANCE;
+    }
 
     public static void display(String title, String text) {
         display(title, text, 5000);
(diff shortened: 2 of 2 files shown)


andrejn-ct:
I have verified that this PR successfully resolves issue #1739

muros-ct:
@lorthirk can you please confirm this change.
It is change that affects complete UI info/error popup information.
If this change is used it will always display only last message received, no stacking of messages.
Is this desired functionality?

lorthirk:
Honestly I'm not sure we want to hide messages to the user just because of overlapping windows. @ct-ajovanovic Can you elaborate a bit more on why this would be a solution? Thanks!

ct-ajovanovic:
Hi @lorthirk 
Besides the visual overlapping of the message boxes, there is also the information overlap. For example in the next screenshot:
img123

^ The user sees two opposite messages - the error one is still visible even after a correct action. There is no need for the old no longer valid message to be visible is it?
Also if the actions causing errors are quick enough, we get a whole lot of messages stacked one upon another. Wouldn't displaying only one last error message be more readable for the user?
What do you think?
Thanks
_Aleksandra_

lorthirk:
While I can see why for this specific case could make sense to only show the error message, I'm afraid that we can't just take the last message. What if an error is shown before  another "OK" message?

Moreover, having an operation with simultaneous "OK" and "Error" is certainly a bad UX: so we should get to the root cause of the problem here, and avoid such situations altogether rather than taking such approach.

@Coduz and @LeoNerdoG, your inputs?

muros-ct:
Just as a reminder. Message stacking when user is logged out as session is expired, we get multiple stacked messages saying You were logged out.

LeoNerdoG:
Hi @lorthirk,
I have checked this and other than double "You are logged out message" everything is OK. 

I think this is a good solution, because user has all the control in his hands - there are no situtations that a third person would interwene and some info/error messages would pop-up. 

So if user does something that produces an error (e.g. create a user with non-unique name) Kapua reports this and if user corrects the error, new message is dispayed, that covers the old one, as old one is obsolete now. 
This is also the reason I have reported this issue in the first place - there is no need to stack multiple messages with same info and also there is no need to see "error" and "info" messages one uppon another, if user corrects his mistakes. 

If you think we should change this, we can work on this, but from my standpoint it looks good 😄

ct-ajovanovic:
Hi @LeoNerdoG 
I just did a rebase and some code clean up. Can you please check if everything is still ok on your side?
Thanks 
_Aleksandra_

LeoNerdoG:
I have verified that after the latest changes this PR successfully resolves issue #1739. Can one of the commiters check the code and merge when all tests complete?

PR #2580, opened 2019-05-07
Fixed KapuaSession data management in REST API

### LINKED ISSUE #2579: REST API 'authentication/refresh' sometimes returns 401 unathorized
**Describe the bug**
When using the REST API resource:

`{endpoint}/authentication/refresh` 

sometimes it returns an error of HTTP 401 unauthorized. Retrying the same request can sometimes end up with the same error and sometime being executed correctly.

**To Reproduce**
1. Make a successful request to `authentication/user`
2. Refresh the token repeatedly (with the new AccessToken given at each refresh) until one of the request is not successful and the response is "401 Unauthorized"
3. Use the same request until it answer 200

**Expected behavior**
All request should success consistently

**Screenshots**
_None_

**Version of Kapua**
Previous 1.1.0

**Type of deployment**
[x] Local Vagrant deployment
[x] Docker
[x] Others

**Main component affected**
[x] REST API
[ ] Message Broker
[ ] - Others

**Additional context**
_None_

Coduz:
REST API resources under `authentication/*` (except `authentication/logout`) are not marked as `noSessionCreation` in the `shiro.ini` file. 
This was causing some of the session information to be left on the Thread. When retrieving the access token to match the refreshed access token, some thread had information stored from another thread and this was causing an InternalError, which then was reported as `401 Unauthorized` to the requester.
 
**Related Issue**
This PR fixes #2579 

On all REST API login resources added a clean up phase

**Screenshots**
_None_

_None_

### DIFF
--- a/rest-api/core/src/main/java/org/eclipse/kapua/app/api/core/auth/KapuaTokenAuthenticationFilter.java
+++ b/rest-api/core/src/main/java/org/eclipse/kapua/app/api/core/auth/KapuaTokenAuthenticationFilter.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2016 Eurotech and/or its affiliates and others
+ * Copyright (c) 2016, 2019 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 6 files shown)


Coduz:
After a small talk with Claudio I realized that the previous solution was not covering all scenarios.

The current solution is to use a `javax.servlet.Filter` to clean up manually all request after the processing.

PR #2592, opened 2019-05-13
Removed wire service from configuration panel

ct-anaalbic:

Brief description of the PR.
Removed wire service from configuration.

**Related Issue**
This PR fixes issue #1800.

Wire service is removed from configuration panel in devices. This service is removed by service name, and in `if` statement some cases are tested. Also, with `||` operator this statement don't work, but with `&&` works.

**Screenshots**
/

/

### DIFF
--- a/console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/configuration/DeviceConfigPanel.java
+++ b/console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/configuration/DeviceConfigPanel.java
@@ -433,6 +433,15 @@ public class DeviceConfigPanel extends LayoutContainer {
         field.setAllowBlank(true);
         field.setFieldLabel(param.getName());
         field.addPlugin(dirtyPlugin);
+        if (param.getId().equals("WireGraph")) {
+            field.setMaxLength(1000000);
+        }
+        if (param.getId().equals("brokerXml")) {
+            field.setMaxLength(1000000);
+        }
+        if (param.getId().equals("users")) {
+            field.setMaxLength(1000000);
+        }
 
         if (param.isRequired()) {
             field.setAllowBlank(false);
@@ -454,7 +463,7 @@ public class DeviceConfigPanel extends LayoutContainer {
             field.setMaxLength(1);
             field.setValidator(validator);
         }
-        if (validator instanceof StringValidator) {
+        if (validator instanceof StringValidator && !param.getId().equals("WireGraph") && !param.getId().equals("brokerXml") && !param.getId().equals("users")) {
             field.setValidator(validator);
         }
         field.setLabelStyle("word-break:break-all");


LeoNerdoG:
I have checked this PR and it successfully resolves issue #1800. Basiaclly we have removed the WireGraphService from Service list in Devices -> Configuration so we would like to know if this is OK.
@Coduz can you please check the code? Ana had to use AND operators instead OR, because otherwise this did not work - is this ok?

Coduz:
Hi all, 

The statement should work like that
`if(!(ocd.getName().equals("WireGraphService") || ocd.getName().equals("WireService") || ocd.getName().equals("GraphService")))`

BTW, if the problem is that the limit on the field is wrong, why removing it and not just add an exception on the code that creates the forms? 

Regards, 

_Alberto_

ct-anaalbic:
Hi @Coduz,

I will change this `if` statement, so AND operator will be replaced with OR operator.
I had conversation with @lorthirk  about this issue, and he told me that, in his opinion, the best solution is to remove wire graph service from list of services in device configuration. The only way I found is to remove service by name. I think that end result is the same if wire service is removed from list, or if exception is added. Am I right? 

Regards, 
_Ana_

Coduz:
Hi @ct-anaalbic ,

sorry I didn't know that you talk with @lorthirk via other communication channels. 

I think that removing those configuration components from the list is removing the only way to manage the wires from Kapua, even if it is pretty odd. 

I think that would made much more sense to have a exception to have a greater limit when the current component is the one of the wires around here [DeviceConfigPanel:418](https://github.com/eclipse/kapua/blob/develop/console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/configuration/DeviceConfigPanel.java#L418)

What do you think?

Regards, 

_Alberto_

ct-anaalbic:
If I understand, you think that better solution is to change limit of this field in wire service. I asked @lorthirk  which limit is good in this case, but he told me that he see no use for the field, because we never want to mess with the graph in the first place. Also, he told me that we should not have graph in the list of configurations. I can set limit to be higher, but I don't know which limit is ok for wire graph field. My suggest for Claudio was limit of 10.000.000 characters, but he said that 10.000.000 is equal to no limit. Can you tell me what is good solution and which limitation we can use for wire graph field? 

Regards,
_Ana_

Coduz:
I think that we can think about the size of the actual file. 

10.000.000 char would translate to a 10MB file (more or less) which can be heavy to transfer over MQTT. 

1MB could be a good limit I think, and if we need more, we will increase. 

Does it sound good? 

Regards, 

_Alberto_

ct-anaalbic:
Yes, I will do it now. 

Regards, 
_Ana_

LeoNerdoG:
I have rechecked this PR and now the limit is properly set to 1M characters. i think this is enaugh for several hunderd componentes in Kura wires (it depends on the complexity of the component), which should be more then enaugh for even most compelx graphs. 
If somebody reaches this limit anyway, we can fix that then. 

Nevertheless I think this can be merged as user now cannot enter more than 1M characters. can one of the commiters check the code and merge if everything is OK?

Coduz:
It makes sense to change the limit for all of the String fields? Even if the problem right now is only on few fields?

ct-anaalbic:
Hi @Coduz 

We think that this limit is good solution for other string fields, and also, as Leonardo said, if this limit increase we can change it. But if you have some other idea, or if you think that the better solution is to only wire graph field has this limit of 1000000, I will make the changes.

Regards, 
_Ana_

LeoNerdoG:
I have verified that after the latest changes everything looks fine, the limit is now 1M and cannot be breached. 
Can one of the commiters check the code and mereg if everything is OK?

console/module/device/src/main/java/org/eclipse/kapua/app/console/module/device/client/device/configuration/DeviceConfigPanel.java
```diff
@@ -433,6 +433,15 @@ public void handleEvent(BaseEvent be) {
         field.setAllowBlank(true);
         field.setFieldLabel(param.getName());
         field.addPlugin(dirtyPlugin);
+        if (param.getId().equals("WireGraph")) {
```
muros-ct: DeviceConfigPanel is supposed to dynamically read parameters of any kura deployment.
In this code we explicitly reference parameters by their name and act accordingly. This means that kapua will have to be recompiled when new package is deployed with a field that is not in this if sequence.

Coduz: Yes, the idea is to have a set of well known fields that have a greater limit of characters. This is to fix the fact that Kura seems not providing those max values.

Does it make sense?

muros-ct: You suggest that this could be a viable solution to our problem. So that one of these fields is WIreGraph constant?

Coduz: Yes. A bit dirty as a solution but it works for now, isn't it?

PR #2600, opened 2019-05-14
Implemented Device Connect Job Trigger

### LINKED ISSUE #2599: Implement trigger to fire a job when the device connects
Implement a "process on connect" feature.

For a job, being able to create a trigger that fires a job when a device connects.

_None_

**Additional context**
This is useful when an account has devices that are not always connected and have only small windows of broker connection

Coduz:
This PR implements the feature of the Device Connect Job Trigger.

**Related Issue**
This PR closes #2599 

Implemented a new Trigger Definition Service (similar to the JobStepDefinitionService).

Ported the two existing trigger types (cron and interval) and added a new type for the "Device Connect".

Extended the `DeviceMessageListener` to check also for jobs waiting to be executed on the connecting device. 

Added also the code to convert old trigger into the refactored ones.

**Screenshots**
_None_


LeoNerdoG:
Hi @Coduz,
as agreed, I am posting issues that I found with this PR on GUI, here. 

- [ ] »Starts On » and »Ends On« fields are not alligned with other fields:

Screenshot 2019-05-15 at 08 15 37

- [ ]  If user selects »Cron Job« from schedule definition, the »Add Schedule« window is too small for »Trigger Properties«:

Screenshot 2019-05-15 at 08 47 42

- [ ]  I get »Persistence Error« if Cron Expression is in form »12« (it should be »Unable to validate CRON expression....«):

Screenshot 2019-05-15 at 09 04 11

- [ ]  »Retry Interval [s]« field in error state without error message if user just clicks on the fileld, then clicks outside the box, changes »Schedule definition« from interval job to Cron job and then back to Interval job:

Screenshot 2019-05-15 at 09 13 56

- [ ]  If Start date/time is in past, Kapua does not return error that the start date is in the past as it does for end date/time (example of how it should be):

Screenshot 2019-05-15 at 08 24 18

- [ ] If user enteres invalid number into »Retry interval« field, Kapua does not return in bottom right corner as when CRON schedule is wrong:

Screenshot 2019-05-15 at 14 26 56

- [ ] Error message when manually setting date in the past is wrong. In case of 15.5.2019 it says »The date in this field must be after 15/5/2019« wiich is wrong, as it can be set to 15.5.2019, so the error message should be changed to »The date in this field must be after 15/5/2019«. This is in Start and End date:

Screenshot 2019-05-15 at 14 31 20

Screenshot 2019-05-15 at 14 31 28

I haven't finished testing yet, so I think there will be several more, but the majoriti is here.

Coduz:
On the last one, the start and end date of the "process on connect" defines range of time of which that "trigger" can start a job on a connecting devices. It does not start any job unless one of the target connects.

Coduz:
Hi @LeoNerdoG ,

I should have fixed all the issues you pointed out.

Some of them where already fixed. The first one I would keep it like, and not complain if the start date is in the past when the selected day is the current date.

Regards, 

_Alberto_

PR #2602, opened 2019-05-15
Fix for String Channel values not displayed correctly

### LINKED ISSUE #2597: String Channel Values Not Parsed Correctly
If user has a device with some String assests (with sample valuse such as "string" or "123"), these strings are not properly parsed when displaying in Kapua - so instead "string" and "132" we get &amp;quot;string&amp;quot; and &amp;quot;123&amp;quot;

**Testflow:**
2. Connect a live Kura device to Kapua (e.g. device0) with some sample drivers (e.g. driver0) and assets (e.g. asset0), that have String channels with values "string"
3. In Kapua, go to Devices -> device0 -> Assets and observe the values of the assets

**Expected behavior:**
Double quotes should be parsed as double quotes and not as `&quot'`.

**Screenshot:**
Screenshot 2019-05-14 at 09 28 09

**Version of Kapua**
1.1.0-SNAPSHOT

**Main component affected**
[ ] REST API
[ ] Message Broker
[ ] - Others

Browsers used: 
- Google Chrome (Version 74.0.3729.131 (Official Build) (64-bit))
- Safari (Version 12.1)
- Mozilla Firefox (66.0.3 (64-bit))
- Opera (Version:60.0.3255.70)

ct-ajovanovic:

Brief description of the PR.
Fix for String Channel values not displayed correctly.

**Related Issue**
This PR fixes/closes #2597 

Solved problems when displaying special characters in the Assets -> Channel fields. 
The string value is unescaped using the `KapuaSafeHtmlUtils.htmlUnescape(String)` method. 
Also the `htmlUnescape()` method was updated:
- the if check for `&amp;` string and its replacement with `&` was moved to first place (more efficient as all other cases look for the `&` and not `&amp;` at the beginning of the string)
- added new _if case to_ check for `amp;` which is sometimes unnecessarily added and should be removed (like `amp;amp;amp;amp;amp;` after the first ` &amp;` in `&amp;amp;amp;amp;amp;amp;quot;string&amp;` )

**Screenshots**
_None_

_None_

### DIFF
--- a/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/util/KapuaSafeHtmlUtils.java
+++ b/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/util/KapuaSafeHtmlUtils.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017 Eurotech and/or its affiliates and others
+ * Copyright (c) 2017, 2019 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
@@ -30,6 +30,12 @@ public class KapuaSafeHtmlUtils {
             return null;
         }
 
+        if (safeHtml.indexOf("&amp;") != -1) {
+            safeHtml = safeHtml.replace("&amp;", "&");
+        }
+        if (safeHtml.indexOf("amp;") != -1) {
+            safeHtml = safeHtml.replace("amp;","");
+        }
         if (safeHtml.indexOf("&lt;") != -1) {
             safeHtml = safeHtml.replace("&lt;", "<");
         }
@@ -41,9 +47,6 @@ public class KapuaSafeHtmlUtils {
         }
         if (safeHtml.indexOf("&#39;") != -1) {
             safeHtml = safeHtml.replace("&#39;", "'");
-        }
-        if (safeHtml.indexOf("&amp;") != -1) {
-            safeHtml = safeHtml.replace("&amp;", "&");
         }
         return safeHtml;
     }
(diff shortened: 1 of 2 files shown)


LeoNerdoG:
I have verified that this PR successfully resolves issue #2597. The only problem is when sending special symbols from Kapua to Kura (the other way around is OK) - in this case the string does not get parsed correctly, but that is issue for Kura project. 
Can one of the commiters check the solution and merge if everything is OK?

console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/util/KapuaSafeHtmlUtils.java
```diff
@@ -30,6 +30,12 @@ public static String htmlUnescape(String safeHtml) {
             return null;
         }
 
+        if (safeHtml.indexOf("&amp;") != -1) {
+            safeHtml = safeHtml.replace("&amp;", "&");
+        }
+        if (safeHtml.indexOf("amp;") != -1) {
```
Coduz: Wouldn't this change every occurrence of the string "amp" in every string?

ct-ajovanovic: @Coduz To answer your question, this wouldn't change every occurrence of the string `amp` but  only `amp;`.
There are cases where we have one or multiple unnecessary instances of `amp;` string added, like here:
`&amp;amp;amp;amp;amp;amp;quot;string&amp;`
the sting `&amp;` gets displayed correctly as `&` but we are still have multiple `amp;` instances that prevent the complete "unescape" for all of the string characters, and we are left with something like this:
`&amp;amp;amp;amp;amp;quot;string&;`
The string is partially "unescaped" but not completely.
With this solution the unnecessary `amp;` is  removed, and the correct value displayed after only one `htmlUnescape` call.
I guess that there is also the case where the user intentionally enters `amp` as a part of the string. But what probability is it for it to be entered in combination with`;` ?
What do you think?

muros-ct: @Coduz Is it ok to merge this?

PR #2609, opened 2019-05-23
Feature stepdefinitionexamples

### LINKED ISSUE #2606: Add an example value for JobStepDefinition.properties
Currently on the JobStepAddDialog, when you select a Job Step Definition is not clear what you need to write in the field and there is no hint on what the user is expected to type in those fields.

Have a way to provide the user with some example data that can be edited to the user's preference and then can be submitted to the platform.

_None_

**Additional context**
_None_

gbarbon:
Adding example value to job step property definition.

**Related Issue**
This PR fixes/closes #2606 

Added column in the job_step_definition_property table in the kapua db.
Added "Fill With Template" button in JobStepAddDialog

**Screenshots**
_none_

Added column also in the job_step_property table in the kapua db for simplicity.
Modified button style in Button class in console/module/api/ui/button to avoid focus after click.

### DIFF
--- a/console/module/api/src/main/resources/org/eclipse/kapua/app/console/module/api/client/messages/ConsoleMessages.properties
+++ b/console/module/api/src/main/resources/org/eclipse/kapua/app/console/module/api/client/messages/ConsoleMessages.properties
@@ -11,7 +11,6 @@
 #     Red Hat Inc
 ###############################################################################
 swelcome=Welcome, {0}.
-
 ok=Ok
 yes=Yes
 no=No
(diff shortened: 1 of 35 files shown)


console/module/api/src/main/resources/org/eclipse/kapua/app/console/module/api/client/messages/ConsoleMessages.properties
```diff
@@ -11,7 +11,6 @@
 #     Red Hat Inc
 ###############################################################################
 swelcome=Welcome, {0}.
-
```
Coduz: Please restore empty lines

gbarbon: empty lines restored

PR #2628, opened 2019-06-24
Job Engine Standalone Service

lorthirk:
This PR refactors the Job Engine outside of the Web Console in a standalone service that can run side to side with the current deployment. It exposes all the `JobEngineService` API in a REST endpoint.

**Related Issue**
No related issues

This PR is composed by two parts:
- An API and a reference implementation to develop standalone services based on Eclipse Vert-x and some of its sub projects, like Vert-x Web
- A standalone Spring Boot application that runs said standalone service implementation to expose the Job Engine via REST API. Such application will hold separate instances for any needed service (e.g. Authentication, Authorization, etc.) in the container.

Since the Job Engine won't run in the Web Console container anymore, a Job Engine HTTP Proxy has been developed, so that any interaction with the Job Engine from the Console will occur via REST.


lorthirk:
I'm closing this because I will probably try a different approach

PR #2646, opened 2019-07-09
Rationalized dependency versions in main pom.xml files

lorthirk:
All the dependencies declared in `<dependencyManagement>` with more than one usages have been moved to the root `pom.xml` file or, where appropriate, in a lower level `pom.xml` (e.g. all the dependencies strictly related to the console are declared in `console/pom.xml`; same goes for rest-api and assembly) and their version declared as Maven properties inside the `pom`s.

@ctron, can we also upgrade `org.eclipse.paho:org.eclipse.paho.client.mqttv3` in `kapua-client-gateway-provider-paho` to 1.2.1 as the other paho dependencies we have?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,6 +29,7 @@
 
         <!-- Dependencies versions -->
         <activemq.version>5.14.5</activemq.version>
+        <aopalliance.version>1.0</aopalliance.version>
         <artemis.version>2.2.0</artemis.version>
         <assertj.version>3.2.0</assertj.version>
         <camel.version>2.16.3</camel.version>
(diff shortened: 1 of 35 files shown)


ctron:
> @ctron, can we also upgrade org.eclipse.paho:org.eclipse.paho.client.mqttv3 in kapua-client-gateway-provider-paho to 1.2.1 as the other paho dependencies we have?

I think so.

pom.xml
```diff
@@ -262,20 +283,27 @@
...
                     <executions>
                         <execution>
                             <id>default-cli</id>
                             <configuration>
                                 <licenseMerges>
                                     <licenseMerge>The Apache Software License, Version 2.0|AL 2.0</licenseMerge>
-                                    <licenseMerge>The Apache Software License, Version 2.0|Apache License 2.0</licenseMerge>
-                                    <licenseMerge>The Apache Software License, Version 2.0|Apache License, Version 2.0</licenseMerge>
-                                    <licenseMerge>The Apache Software License, Version 2.0|Apache Public License 2.0</licenseMerge>
+                                    <licenseMerge>The Apache Software License, Version 2.0|Apache License 2.0
```
Coduz: Please remove newlines added by auto-formatting

PR #2651, opened 2019-07-11
Improved detection of Device Management Timeout

### LINKED ISSUE #2155: Create new, more specific error codes to cover various CALL_ERROR instances

Currently, a wide range of issues when a device management operation fails, a lot of different causes are treated as a generic `DeviceManagementException` with the error code `CALL_ERROR`. We should instead provide different, more verbose error codes to be applied to various errors that may occur in a Device Management operation.

- More error codes to be used when we currently throw a CALL_ERROR
- Avoid exception swallowing

Coduz:
This PR improves the Exception model in the `kapua-transport` and `kapua-device-management` modules. 

**Related Issue**
This PR fixes #2155 

Created two specific exception for each layer of the device management.

- TimeoutException: thrown when timeout occurs.
- SendException: thrown for every other error that occurs when trying to send the request.

These exception have been created under `api` modules and they became the exception thrown by the API methods.

**Screenshots**
_None_

_Nono_


service/device/commons/src/main/java/org/eclipse/kapua/service/device/management/commons/call/DeviceCallExecutor.java
```diff
@@ -76,70 +87,85 @@ public DeviceCallExecutor(RQ requestMessage, Long timeout) {
...
-        }
-        break;
-        default:
-            throw new DeviceMenagementException(DeviceMenagementErrorCodes.REQUEST_BAD_METHOD);
-        }
+            //
+            // Send the request
+            DeviceResponseMessage responseMessage;
+            switch (requestMessage.getChannel().getMethod()) {
+                case CREATE: {
```
lorthirk: ha! 🤣

I guess we can remove these curly braces!

PR #2669, opened 2019-07-26
Remove scopeId null check on queries and counts

lorthirk:
Since `scopeId` can have `null` value, meaning that the query runs on all accounts, we shouldn't assume that `scopeId` must not be null in `query` and `count` methods


service/security/shiro/src/main/java/org/eclipse/kapua/service/authorization/access/shiro/AccessInfoServiceImpl.java
```diff
@@ -176,7 +176,6 @@ public AccessInfo findByUserId(KapuaId scopeId, KapuaId userId) throws KapuaExce
     public AccessInfoListResult query(KapuaQuery<AccessInfo> query)
             throws KapuaException {
         ArgumentValidator.notNull(query, "query");
-        ArgumentValidator.notNull(query.getScopeId(), "query.scopeId");
```
Coduz: Missed count method

PR #2830, opened 2020-01-20
Fix H2 URI in Dockerfile

lorthirk:
Since January 15, 2020, [The Maven Central Repository access requires https](https://links.sonatype.com/central/501-https-required) to download artifacts. This PR fixes the URI for H2 Jar that would fail otherwise.

**Related Issue**
No related issues


LeoNerdoG:
Hi @lorthirk,
I can confirm that build passes successfully on my machine. 
There is only one failing tests so maybe you should re-run it and merge after it completes?

Coduz (approved):

PR #2853, opened 2020-02-10
Fix MetricPredicate and add MetricExistsPredicate

### LINKED ISSUE #2852: MetricPredicate translates to incompatible query in Elasticsearch REST Client
**Describe the bug**
The current implementation of `MetricPredicate` translate to a query that consider the Datastore Schema. Specifically, an example `MetricPredicate` would translate to JSON as

```
{
    "query": {
        "range" : {
            "temperature" : {
                "gte" : 10,
                "lte" : 20,
            }
        }
    }
}
```

While it should translate to this JSON to support our current schema:

```
{
    "query": {
        "range" : {
            "metrics.temperature.int" : {
                "gte" : 10,
                "lte" : 20,
            }
        }
    }
}
```

This means that filtering for a metric type and/or value would always return 0 entries

**To Reproduce**
1. Perform a REST call to `/{scopeId}/dataMessages` filtering on `metricName`, `metricType`, `metricMin` and `metricMax`

**Expected behavior**
A set of results matching the name, type and values would return

**Version of Kapua**
1.0.9
1.1.2

**Type of deployment**
[x] Local Vagrant deployment
[x] Docker
[x] Others

**Main component affected**
[x] REST API
[ ] Message Broker
[x] - Others

**Additional context**
Basically any component that uses di ES REST client is affected

lorthirk:
This PR fixes the current implementation of `MetricPredicate` and adds a new `MetricExistsPredicate` to cope with our Elasticsearch data schema

**Related Issue**
Fixes #2852 

`MetricPredicateImpl` has been reworked to match the "metrics.metricName.typ" notation; also `MetricExistsPredicate` has been created to match the schema notation. The current `ExistsPredicate` hasn't changed since it could be used for other fields

### DIFF
--- a/service/datastore/api/src/main/java/org/eclipse/kapua/service/datastore/model/query/MetricExistsPredicate.java
+++ b/service/datastore/api/src/main/java/org/eclipse/kapua/service/datastore/model/query/MetricExistsPredicate.java
@@ -1,0 +1,38 @@
+/*******************************************************************************
+ * Copyright (c) 2011, 2020 Eurotech and/or its affiliates and others
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Eurotech - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.kapua.service.datastore.model.query;
+
+/**
+ * Query predicate for matching messages with existing metrics
+ *
+ * @since 1.2.0
+ */
+public interface MetricExistsPredicate extends ExistsPredicate {
+
+    /**
+     * Gets the metric type to search.
+     * This is required because metric with the same name can have different types.
+     *
+     * @return The metric type
+     * @since 1.2.0
+     */
+    Class<?> getType();
+
+    /**
+     * Sets the metric type so search.
+     *
+     * @param type The metric type to search.
+     * @since 1.2.0
+     */
+    void setType(Class<?> type);
+
+}
--- a/service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/model/query/MetricExistsPredicateImpl.java
+++ b/service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/model/query/MetricExistsPredicateImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2017 Eurotech and/or its affiliates and others
+ * Copyright (c) 2011, 2020 Eurotech and/or its affiliates and others
  *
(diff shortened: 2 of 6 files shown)


service/datastore/api/src/main/java/org/eclipse/kapua/service/datastore/model/query/MetricExistsPredicate.java
```diff
@@ -0,0 +1,38 @@
...
+ * @since 1.0.0
+ */
+public interface MetricExistsPredicate extends ExistsPredicate {
+
+    /**
+     * Gets the metric type to search.
+     * This is required because metric with the same name can have different types.
+     *
+     * @return The metric type
+     * @since 1.0.0
```
Coduz: `@since 1.2.0`

service/datastore/api/src/main/java/org/eclipse/kapua/service/datastore/model/query/MetricExistsPredicate.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Contributors:
+ *     Eurotech - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.kapua.service.datastore.model.query;
+
+/**
+ * Query predicate for matching messages with existing metrics
+ *
+ * @since 1.0.0
```
Coduz: `@since 1.2.0`

service/datastore/api/src/main/java/org/eclipse/kapua/service/datastore/model/query/MetricExistsPredicate.java
```diff
@@ -0,0 +1,38 @@
...
+     * @return The metric type
+     * @since 1.0.0
+     */
+    Class<?> getType();
+
+    /**
+     * Sets the metric type so search.
+     *
+     * @param type The metric type to search.
+     * @since 1.0.0
```
Coduz: `@since 1.2.0`

service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/model/query/MetricExistsPredicateImpl.java
```diff
@@ -0,0 +1,71 @@
...
+import org.eclipse.kapua.service.datastore.internal.schema.KeyValueEntry;
+import org.eclipse.kapua.service.datastore.internal.schema.SchemaUtil;
+import org.eclipse.kapua.service.datastore.model.query.MetricExistsPredicate;
+
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+/**
+ * Implementation of query predicate for checking if a field exists
+ *
+ * @since 1.0
```
Coduz: `@since 1.2.0`

Coduz (changes requested):

Coduz (approved):

PR #2859, opened 2020-02-20
Allow download and upload of Kura native snapshot

### LINKED ISSUE #2704: Device configurations - upload and apply native Kura snapshots
The *Upload & Apply* feature currently supports only Kapua snapshot format. 
Suppose that a user saved the snapshot from the Kura web console by accessing the device locally and wants to apply it to another device already deployed in the field. There's no easy way to upload the file in Kapua without hacking the content of the file.

*Upload & Apply* should allow the user to specify the format of the snapshot and apply the necessary conversion logic in order to translate it into a Kapua snapshot.

Workaround. At the moment a user can:
* connect the first Kura device to Kapua
* *Download* the snapshot
* connect the second Kura device 
* *Upload and Apply* the snapshot

lorthirk:
Currently, Device Snapshots downloaded from the Console or REST APIs cannot be directly applied to a Kura device, since the XML that is downloaded from the Console has a different format the one that Kura can deserialize. With this PR, bot "Standard" (old format) and "Native" (new, Kura-compatible format) snapshots can be both dowloaded and uploaded from the console. Native snapshots can also be directly applied to a device from the Kura console.

**Related Issue**
Fixes #2704

Translation logic has been extracted from `TranslatorAppConfigurationKuraKapua` and `TranslatorAppConfigurationKapuaKura` to a utility class named `KuraDeviceConfigurationUtils`, in the `kapua-device-call-kura` module. The same logic is also called from the Download and Upload servlets to do necessary conversions.

**Screenshots**
Schermata 2020-02-20 alle 17 08 19
Schermata 2020-02-20 alle 17 08 22

Do we agree on "Native" and "Standard" names?


Coduz (changes requested):
"Standard" term alternative wanted.

💰 1.000.000K reward 💰

LeoNerdoG:
> "Standard" term alternative wanted.
> 
> 💰 1.000.000K reward 💰

I suggest "accepted", "common" or "general". 
Did I get the prize? 😸 😸

kcsj17:
Can this be used to upload, via Kapua, a wiregraph that was downloaded via Kura web interface?

LeoNerdoG:
Hi @kcsj17,
yes, you can either upload just a wire snapshot or snapshot of the whole system. 
Just be sure that you select the "Native" option in the UI as @lorthirk described above. 

I have just tested this and it works fine.

lorthirk:
After discussing with @Coduz and @stefanomorson, the following changes have been proposed:

- The format names will be "Platform" and "Device"
- Only one "Upload and Apply" button that can support and autodetect both formats
- The REST API has to be enhanced with the same support (autodetect for upload, query string parameter to choose download format)

Coduz:
Removing project tracking since it is tracked by the issue

lorthirk:
Still needs some thinking as it involves other parts that were overlooked at the beginning. Adding the `proposal` label

PR #2905, opened 2020-03-20
Fix shiro configuration

riccardomodanese:
Brief description of the PR.
Remove hardcoded Shiro configuration and override the ModularRealmAuthorizer to use a more efficient check for multiple permissions at the same time.

**Related Issue**
none

The Shiro configuration was wrongly instantiated by code from the AuthorizationService. So any change done on shiro.ini had no effect (on broker side at least) since it was overwritten later by the AuthorizationService initialization.
The Shiro ModularRealmAuthorizer uses a for loop while checking for multiple permissions. For performance reason, on Kapua side, we would like to use a different method already exposed by the Authorizer.
The shiro.ini configuration is changed and the changes did are described inside the file.

**Screenshots**
none

none

### DIFF
--- a/broker-core/src/main/java/org/eclipse/kapua/broker/core/security/EnhModularRealmAuthorizer.java
+++ b/broker-core/src/main/java/org/eclipse/kapua/broker/core/security/EnhModularRealmAuthorizer.java
@@ -1,0 +1,70 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Eurotech and/or its affiliates and others
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Eurotech - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.kapua.broker.core.security;
+
+import java.util.Collection;
+import java.util.List;
+
+import org.apache.commons.collections.CollectionUtils;
+import org.apache.shiro.authz.Authorizer;
+import org.apache.shiro.authz.ModularRealmAuthorizer;
+import org.apache.shiro.authz.Permission;
+import org.apache.shiro.realm.Realm;
+import org.apache.shiro.subject.PrincipalCollection;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Custom {@link Authorizer} to reduce the query amount using the isPermitted method with the Permission list or String array.
+ * To use this Authorizer a deeply modified shiro.ini is needed.
+ * Without these changes this Authorizer will not have any realm configured. (see shiro.ini for explanation)
+ * This authorizer takes the first valid configured realm and return the isPermitted evaluation skipping any aggregation strategy if more than one valid aggregator is defined.
+ *
+ */
+public class EnhModularRealmAuthorizer extends ModularRealmAuthorizer {
+
+    protected static final Logger logger = LoggerFactory.getLogger(EnhModularRealmAuthorizer.class);
+
+    public EnhModularRealmAuthorizer() {
+    }
+
+    public EnhModularRealmAuthorizer(Collection<Realm> realms) {
+        super(realms);
+    }
+
+    @Override
+    public boolean[] isPermitted(PrincipalCollection principals, List<Permission> permissions) {
+        assertRealmsConfigured();
+        if (CollectionUtils.isEmpty(permissions)) {
+            //return the first realm result
+            //the multiple realms case with aggregator should be handled or do we still have just one realm?
+            for (Realm realm : getRealms()) {
+                return ((Authorizer) realm).isPermitted(principals, permissions);
+            }
+        }
+        return new boolean[0];
+    }
+
+    @Override
+    public boolean[] isPermitted(PrincipalCollection principals, String... permissions) {
+        assertRealmsConfigured();
+        if (permissions != null && permissions.length>0) {
+            //return the first realm result
+            //the multiple realms case with aggregator should be handled or do we still have just one realm?
+            for (Realm realm : getRealms()) {
+                return ((Authorizer) realm).isPermitted(principals, permissions);
+            }
+        }
+        return new boolean[0];
+    }
+
+}
--- a/broker-core/src/main/java/org/eclipse/kapua/broker/core/security/EnhModularRealmAuthorizer.java
+++ b/broker-core/src/main/java/org/eclipse/kapua/broker/core/security/EnhModularRealmAuthorizer.java
@@ -11,8 +11,10 @@
  *******************************************************************************/
 package org.eclipse.kapua.broker.core.security;
 
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.shiro.authz.Authorizer;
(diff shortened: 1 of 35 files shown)


service/security/shiro/src/main/java/org/eclipse/kapua/service/authorization/shiro/AuthorizationServiceImpl.java
```diff
@@ -29,6 +32,28 @@
...
+        else {
+            List<org.apache.shiro.authz.Permission> permissionsShiro = permissions.stream()
+                    .map(permission -> (org.apache.shiro.authz.Permission) permission)
+                    .collect(Collectors.toList());
+            return SecurityUtils.getSubject().isPermitted(permissionsShiro);
+        }
+    }
+
     @Override
     public boolean isPermitted(Permission permission)
```
Coduz, line 57: This should invoke `isPermitted(Lists.newArrayList(permission)` so we can use the same logic and we don't need to keep methods aligned

qa/common/src/main/java/org/eclipse/kapua/qa/common/utils/InitShiro.java
```diff
@@ -0,0 +1,58 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(InitShiro.class);
+
+    @Given("^Init Security Context$")
+    public void start() throws IOException {
+        try {
+            SecurityManager securityManager = SecurityUtils.getSecurityManager();
+            logger.info("Found Shiro security manager {}", securityManager);
+        }
+        catch (UnavailableSecurityManagerException e) {
+            logger.info("Init shiro security manager...");
```
Coduz, line 41: `UnavailableSecurityManagerException` is eaten. 
Add log at least of the message Exception

qa/common/src/main/java/org/eclipse/kapua/qa/common/utils/InitShiro.java
```diff
@@ -0,0 +1,58 @@
...
+            final URL shiroIniUrl = getClass().getResource("/shiro.ini");
+            Ini shiroIni = new Ini();
```
Coduz, line 43: `Ini` supports `.loadFromPath` and using `"classpath:shiro.ini"` should work, right?

broker-core/src/main/java/org/eclipse/kapua/broker/core/security/EnhModularRealmAuthorizer.java
```diff
@@ -0,0 +1,70 @@
...
+            for (Realm realm : getRealms()) {
+                return ((Authorizer) realm).isPermitted(principals, permissions);
+            }
+        }
+        return new boolean[0];
+    }
+
+    @Override
+    public boolean[] isPermitted(PrincipalCollection principals, String... permissions) {
+        assertRealmsConfigured();
```
Coduz: Make this reuse `boolean[] isPermitted(PrincipalCollection principals, List<Permission> permissions)`

broker-core/src/main/java/org/eclipse/kapua/broker/core/security/EnhModularRealmAuthorizer.java
```diff
@@ -0,0 +1,70 @@
...
+    public EnhModularRealmAuthorizer(Collection<Realm> realms) {
+        super(realms);
+    }
+
+    @Override
+    public boolean[] isPermitted(PrincipalCollection principals, List<Permission> permissions) {
+        assertRealmsConfigured();
+        if (CollectionUtils.isEmpty(permissions)) {
+            //return the first realm result
+            //the multiple realms case with aggregator should be handled or do we still have just one realm?
```
Coduz: Per permission checking right now we have only one `AuthorizingRealm` 😁 Remove this question

service/security/shiro/src/main/java/org/eclipse/kapua/service/authorization/shiro/AuthorizationServiceImpl.java
```diff
@@ -29,6 +32,28 @@
...
+
+        if (session == null) {
+            throw new KapuaUnauthenticatedException();
+        }
+        if (session.isTrustedMode()) {
+            boolean[] returnedPermissions = new boolean[permissions.size()];
+            for (int i=0; i<permissions.size(); i++) {
+                returnedPermissions[i] = true;
+            }
+            return returnedPermissions;
```
Coduz: Replace this with `Arrays.fill(new boolean[permissions.size()], true );`

broker-core/src/main/java/org/eclipse/kapua/broker/core/security/EnhModularRealmAuthorizer.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public EnhModularRealmAuthorizer(Collection<Realm> realms) {
+        super(realms);
+    }
+
+    @Override
+    public boolean[] isPermitted(PrincipalCollection principals, List<Permission> permissions) {
+        assertRealmsConfigured();
+        if (CollectionUtils.isEmpty(permissions)) {
```
Coduz: Maybe this is:
`if (!CollectionUtils.isEmpty(permissions)) {`
This is currently doing something only when  the given `List<Permission>` is empty

Coduz (changes requested):

broker-core/src/main/java/org/eclipse/kapua/broker/core/security/EnhModularRealmAuthorizer.java
riccardomodanese: Manipulating data to reuse code (also by just invoking an object instantiation), in a critical path, is not a good practice in my opinion (the performances will be affected)
Moreover, we are using multiple permissions check (so with the List) in just one single point (the broker security plugin, if I'm right).
So, as I said, adapting the most used path code (one single permission) to follow a different path (multiple permissions), involving object manipulation will result just in a performance degradation.

qa/common/src/main/java/org/eclipse/kapua/qa/common/utils/InitShiro.java
riccardomodanese: It's the scope of this code. There is no needs to rethrow an exception if you already know what to do when happens.
If the Shiro context it's already configured the test code proceeds, otherwhise, in the catch block the context it's initialized.

qa/common/src/main/java/org/eclipse/kapua/qa/common/utils/InitShiro.java
riccardomodanese: May be, I didn't test this option.
I reused the already working code from the broker side to init Shiro context for those tests aren't doing it.

broker-core/src/main/java/org/eclipse/kapua/broker/core/security/EnhModularRealmAuthorizer.java
riccardomodanese: Manipulating data to reuse code (also by just invoking an object instantiation), in a critical path, is not a good practice in my opinion (the performances will be affected)

service/security/shiro/src/main/java/org/eclipse/kapua/service/authorization/shiro/AuthorizationServiceImpl.java
riccardomodanese: Manipulating data to reuse code (also by just invoking an object instantiation), in a critical path, is not a good practice in my opinion (the performances will be affected)

broker-core/src/main/java/org/eclipse/kapua/broker/core/security/EnhModularRealmAuthorizer.java
riccardomodanese: Having one or more realms is not completely on our hand. If someone will customize Kapua adding another realm should be aware (at least with a commented code) of possible restrictions.

Coduz: Yes but this question is useless here. 
Documentation should be the proper way 😁

qa/common/src/main/java/org/eclipse/kapua/qa/common/utils/InitShiro.java
Coduz: I said just to log the `Exception` not re-throw it 😁

service/security/shiro/src/main/java/org/eclipse/kapua/service/authorization/shiro/AuthorizationServiceImpl.java
Coduz: This will result in duplicated code. Next "sonar fixes branch" you know that this will be changed anyway....

broker-core/src/main/java/org/eclipse/kapua/broker/core/security/EnhModularRealmAuthorizer.java
Coduz: This will result in duplicated code. Next "sonar fixes branch" you know that this will be changed anyway....

qa/common/src/main/java/org/eclipse/kapua/qa/common/utils/InitShiro.java
riccardomodanese: There is NO sense. It's not an Exception, it's like a check.
Instead of writing
if (!security manager defined)
It's used the try catch pattern

broker-core/src/main/java/org/eclipse/kapua/broker/core/security/EnhModularRealmAuthorizer.java
riccardomodanese: won't fix option can be used

service/security/shiro/src/main/java/org/eclipse/kapua/service/authorization/shiro/AuthorizationServiceImpl.java
riccardomodanese: won't fix option can be used

PR #2909, opened 2020-03-23
Remove propOrder from XmlType and Migrate all the XmlAccessType from FIELD to PROPERTY

lorthirk:
This PR refactors the JAXB annotations so that:

1. `propOrder` is removed from all entities, since it's not needed anymore
2. All `XmlAccessType.FIELD` are migrated to `XmlAccessType.PROPERTY` so that they all reflect getters and setters

**Related Issue**
No related issues

These changes can potentially break REST APIs according to the single contracts. It would be better to merge this after having proper REST API testing.


lorthirk:
In order to be sure that these changes don't break anything, we need to implement proper marshalling/unmarshalling tests

lorthirk:
PR goes back to draft to wait for tests and to fix conflicts

PR #2924, opened 2020-04-01
Fix event bus instance leak

riccardomodanese:

Brief description of the PR.
Fix a possible leak on Event Bus.
When the Event Bus tries a re-start and, for some reason, the start throws an exception, the new instance will be not properly cleaned.
Added a clean-up 

**Related Issue**
none

none

**Screenshots**
none

none


stefanomorson (approved):

stefanomorson (commented):
Moving the listener to its own class is ok but stopping the retries is probably the most important change here. Agree with these two.
I'm a bit concerned that there could be a race condition on the restart() method due to the fact that while the restart is in execution a new exception coming from the old instance or the new instance may trigger a new restart process. However, if restart is going to be synchronized, we have to make sure that the onException() method originating it must not block the qpid library underneath.

PR #2947, opened 2020-04-23
Restrictable Service Configurations

lorthirk:
This PR introduces support for Restrictable Service Configurations. Two methods are introduced in `AbstractKapuaConfigurableService`, meant to be overridden by the extending services:

- `boolean isAvailableService()`: controls whether the whole service is available or not
- `boolean isAvailableProperty(KapuaTad ad)`: controls whether a single property is available or nor

**Related Issue**
No related issues


lorthirk:
The issue where a user could change some settings on its own account via REST API is now fixed by #2954.

Coduz (approved):

lorthirk:
PR goes back to draft to wait for fixes in tests

PR #2951, opened 2020-04-24
Disable auto scroll in timestamp refresh

### LINKED ISSUE #2948: View Data by topic - Topic-List scrolls to top when a folder is expanded
Ensure that in the topic-list panel there's a list of first level topics long enough to activate scrolling. Move down in the the list and expand the first level folder icon. Ensure the folder has sub topics inside and there are some metrics as well. As soon as the list of the metrics appear on the right, the  topic-list on the left is scrolled up to the top.
As an additional info, the issue was observed in a case where there are hundreds of first level folders and the metric list is displayed with some delay after the folder has been expanded.
 
**Version of Kapua**
Kapua v1.1.5

**Main component affected**
[ ] REST API
[ ] Message Broker
[ ] - Others

### LINKED ISSUE #2949: View Data by topic - Topic-List scrolls to top when last-updated-on are updated with delay
Ensure that in the topic-list panel there's a list of first level topics long enough to activate scrolling and ensure that the update of the last-updated-on values is displayed with some delay after the topic-list. Push refresh button and move down in the the list. As soon as the last-updated-on values appear, the  topic-list on the left is scrolled up to the top.
As an additional info, the issue was observed in a case where there are hundreds of first level folders and the last-updated-on are displayed few seconds after the refresh button was clicked.

**Expected behavior**
The topic-list panel should keep the current position (not scrolled to the top).

**Version of Kapua**
1.1.5

**Main component affected**
[ ] REST API
[ ] Message Broker
[ ] - Others

lorthirk:
This PR fixes the view scrolling to the top when expanding a node.

**Related Issue**
Fixes #2948, Fixes #2949

### DIFF
--- a/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/ui/grid/KapuaTreeGrid.java
+++ b/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/ui/grid/KapuaTreeGrid.java
@@ -1,0 +1,26 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Eurotech and/or its affiliates and others
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Eurotech - initial API and implementation
+ *******************************************************************************/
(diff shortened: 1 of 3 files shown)


Coduz:
Remove description data since those are tracked by issues

lorthirk:
The disabling of view buffering has been scratched from this PR because it needs more investigation. Only the scrolling to the top is fixed here.

Coduz (approved):

PR #2954, opened 2020-04-28
Prevent users to change settings on their account

lorthirk:
This PR fixes an issue where a user could change some settings on its own account via REST API.

**Related Issue**
No related issues

Backend configurable services now checks that the user performing the operation is not in the target account, unless the user is the system admin


lorthirk:
We need a refactoring in order to have proper location of `AccountService` in tests

Coduz (approved):

PR #2959, opened 2020-04-30
Feature cache

riccardomodanese:
This PR introduces a caching feature at the service level. This implementation is compliant with the [JCache specification](https://www.jcp.org/en/jsr/detail?id=107), meaning that each cache that implements the JCache interface can be used with this caching feature. 

The cache is implemented a given set of services, in order to improve performances. A simple JCache implementation based on HashMaps is also provided, in order to test the caching feature with the QAs.

**Related Issue**

_N/A_

The package `org.eclipse.kapua.commons.service.internal.cache` contains the `KapuaCacheManager`, which is responsible for instantiating a cache for each service, and the `EntityCache` class, which defines our basic cache implementation  (using entities Ids as keys) on top of JCache. `NamedEntityCache` extends `EntityCache` by providing an additional cache (using entities names as keys). 

The cache allows avoiding repeated queries to the database, in particular when checking permissions to perform an action on a given method. As an example, the `checkPermission` method of the `AuthorizationService` (responsible for checking if an user is allowed to perform a given operation), performs a chain of calls to the `UserService`, `AccessInfoService`, `AccessPermissionService`, `AccessRoleService`, `RoleService` and `RolePermissionService` services, each of which performs a query to the db. Caching all these services allows considerably reducing the time spent on querying the db. 

Also, note that the message processing by the broker requires methods from the `AccountService` and `DeviceRegistryService`, which both perform transactions to the database. Such transactions are performed for each message received by the broker, thus using a cache on those services allows reducing queries when the same data is required.

The QAs will make use of an in-memory HashMaps-based implementation of JCache in order to mimic the behaviour of a real cache instance. In order to mimic the serialisation, this cache is also provided with a clone method which performs a deep clone of the given entity in `put` and `get` operations.

**Screenshots**

_N/A_

Note also that this cache feature requires a JCache implementation to run, otherwise a 'dummy' cache is used.

Services require to be modified in order to implement the cache. The following services have been cached:
- `AccessInfoService`
- `AccessPermissionService`
- `AccessRoleService`
- `AccountService`
- `DeviceConnectionService`
- `DeviceRegistryService`
- `RoleService`
- `RolePermissionService`
- `UserService`

Each service is able to instantiate is own cache through a dedicated cache factory. However, note that `DeviceConnectionService` is not using a dedicated cache factory, but it relies on the `DeviceRegistryCacheFactory`. Moreover, also the `AbstractKapuaConfigurableService` has been provided with a cache instance, in order to cache the configuration values for each services. A `LocalCache` instance (which must not be confused with a JCache cache instance) has also been added in order to store the config metadata, which do not change after the first time the object is loaded. 

The JCache implementation for the QAs is enabled trough the following property  `commons.cache.provider.classname=org.eclipse.kapua.qa.common.cache.MapCachingProvider`


Coduz:
Commits authored from @gbarbon comes from following PRs:
cache - Internal cache implementation - be31ac1 - #2892
cache - Adding cache to services. - 560ff02 - #2945 
cache - Adding HashMap-based JCache implementation to QAs - 5fa44f2 - #2955

Coduz (approved):

PR #2946, opened 2020-04-23
Added Junit Tests for SemanticVersion 

sonja-ct:
Signed-off-by: Sonja <sonja.matic@comtrade.com>

Added Junit Tests for SemanticVersion

PR #2961, opened 2020-05-05
Fixed Failing Cucumber Unit Tests

LeoNerdoG:
With a fix for a certain REST API bug (preventing users to change their account settings) in PR #2946 and PR #2954 the unit tests for services (tags, users, jobs, scheduling,...) started to fail. 
I have fixed this by introducing a mocked account that is used as basis for the service creation. 
I have already ran PR on my Travis and everything was fine. CodeCov should report 0% increased/decreased coverage, since all the changes made are in the ignored folders. 

**Related Issue**
This PR fixes problems with unit tests in PR #2947 and PR #2954.

**Description of the solution adopted:**
I have added a mocked AccountService and mocked AccountFactory classes to the existing unit tests, so that the check does not fail. This is  achieved with Mockito's "spy" method. The necessary Account methods are injected in the _CucumberWithProeprtiesFor(Service)_ file. 

@lorthirk can you please look at the solution adopted and see if it's ok? After you merge this PR into develop, just rebase your PRs to develop and the tests should be passing normally. 

**Screenshots**
N/A

N/A

### DIFF
--- a/service/job/test/pom.xml
+++ b/service/job/test/pom.xml
@@ -44,5 +44,10 @@
             <artifactId>kapua-scheduler-quartz</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.kapua</groupId>
+            <artifactId>kapua-account-internal</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/service/tag/test/pom.xml
+++ b/service/tag/test/pom.xml
@@ -39,5 +39,10 @@
             <artifactId>cucumber-junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.kapua</groupId>
+            <artifactId>kapua-account-internal</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/service/user/test/pom.xml
+++ b/service/user/test/pom.xml
@@ -39,5 +39,10 @@
             <artifactId>cucumber-junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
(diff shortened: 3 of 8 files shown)


service/tag/test/pom.xml
```diff
@@ -39,5 +39,25 @@
             <artifactId>cucumber-junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.kapua</groupId>
+            <artifactId>kapua-account-internal</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
```
lorthirk: This dependency is added 4 times!

service/user/test/pom.xml
```diff
@@ -39,5 +39,15 @@
             <artifactId>cucumber-junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.kapua</groupId>
+            <artifactId>kapua-account-internal</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
```
lorthirk: This dependency is added twice

service/job/test/pom.xml
```diff
@@ -44,5 +44,15 @@
             <artifactId>kapua-scheduler-quartz</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.kapua</groupId>
+            <artifactId>kapua-account-internal</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
```
lorthirk: This dependency is added twice

lorthirk (changes requested):
Please check for duplicate dependencies in `pom.xml` files 🙂

LeoNerdoG:
@lorthirk thanks for the review - I have deleted the extra dependencies.

PR #2963, opened 2020-05-05
disabling datastore feature parameter

riccardomodanese:
Added a feature to disable completely the datastore feature from the broker.
The idea is to allow Kapua to start and run without any exception in a deploy without any datastore.
This is controlled by an environment variable:

- KAPUA_DATASTORE_DISABLE if true the broker doesn't create any datastore related object (and so doesn't try to connect to it)

**Related Issue**
none

none

**Screenshots**
none

In this PR there's also some updates to the code originally merged with #2947. Specifically:

- `isAvailableService()` has been renamed to `isServiceEnabled()`
- `isAvailableProperty()` has been renamed to `isPropertyEnabled()`
- `isServiceEnabled()` has been moved from `AbstractKapuaConfigurableService` to `AbstractKapuaService`
- When reading configuration for a disabled service, now `KapuaServiceDisabledException` is thrown instead of silently return `null`


riccardomodanese:
@lorthirk should we add a similar feature in the console and rest api also?

lorthirk (commented):
This means that `<!-- BEGIN DATASTORE PERSISTENCE -->` and `<!-- END DATASTORE PERSISTENCE -->` comments have now special semantics. Are we sure we want to do this? I suspect that someone could not understand this by just looking at the file. Maybe we should at least document this somewhere?

Coduz (approved):

PR #2969, opened 2020-05-11
datastore driver - adding reconnect to datastore capabilities

riccardomodanese:
There is no way to start the broker if ElasticSearch is down at startup. The broker is expected to start even if ES is down. The connection with the data-store should be tried periodically, until it has been established

**Related Issue**
none

No exception is thrown if an error occurred when establishing the data-store connection. A connection task is then executed periodically until the connection is established.

**Screenshots**
none

none


riccardomodanese:
Note:
Since the Elastisearch transport client is deprecated since a while and it will be removed soon, the deprecated tag is added to the Kapua transport datastore client objects (the rest client should be used instead).
For this reason the changes of this pr are affecting only the Rest client. The transport client behavior remains at it was before.

lorthirk (approved):

Coduz (approved):

PR #2977, opened 2020-05-15
Added JUnit tests for Commons package

code-incubus:
Signed-off-by: code-incubus nemanja.milutinovic@comtrade.com

Added JUnit tests for classes in a commons package:

- ArgumentValidator.java
- KapuaFileUtils.java
- CommonsValidationRegexTest.java

Related Issue
None

Description of the solution adopted
/
Screenshots
/

Any side note on the changes made
- I put a comment at line 82 in class KapuaFileUtils.java below the "FIXME". We need to adjust the tests after the changes in this class.

### DIFF
--- a/commons/src/main/java/org/eclipse/kapua/commons/util/KapuaFileUtils.java
+++ b/commons/src/main/java/org/eclipse/kapua/commons/util/KapuaFileUtils.java
@@ -78,7 +78,6 @@ public class KapuaFileUtils {
             URL url = getAsURL(filePath);
 
             if (hasHttpScheme(filePath)) {
-                // FIXME:
                 file = File.createTempFile("file-" + System.currentTimeMillis(), ".tmp");
                 FileUtils.copyURLToFile(url, file);
             } else {
--- a/commons/src/test/java/org/eclipse/kapua/commons/util/KapuaFileUtilsTest.java
+++ b/commons/src/test/java/org/eclipse/kapua/commons/util/KapuaFileUtilsTest.java
@@ -1,0 +1,86 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Eurotech and/or its affiliates and others
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Eurotech - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.kapua.commons.util;
+
+import org.eclipse.kapua.commons.setting.KapuaSettingException;
+import org.eclipse.kapua.qa.markers.junit.JUnitTests;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+import java.io.File;
+import java.io.IOException;
+import java.lang.reflect.Constructor;
+import java.net.MalformedURLException;
+import java.net.URL;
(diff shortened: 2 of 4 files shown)


commons/src/test/java/org/eclipse/kapua/commons/util/KapuaFileUtilsTest.java
```diff
@@ -0,0 +1,92 @@
...
+        String[] permittedFormats = new String[] {"https://www.example3345.com/", "http://www.example.com:1080/docs/resource1.html",
+                "https://www.example.com:1080/docs/resource2.html", "file:///c:/EXAMPLE/clock.example"};
+        String [] invalidFormats = new String [] {"www.example.com:1080/docs/resource1.html", "/c:/WINDOWS/clock.example",
+                "http://www.example.com:10800000000000/docs/resource1.html", "htp://www.example.com:1080/docs/resource1.html"};
+
+        URL[] expectedUrl = new URL[] {new URL("https://www.example3345.com/"), new URL("http://www.example.com:1080/docs/resource1.html"),
+                                       new URL("https://www.example.com:1080/docs/resource2.html"), new URL("file:///c:/EXAMPLE/clock.example")};
+        for (int i = 0; i < permittedFormats.length; i++) {
+            try {
+                assertEquals(expectedUrl[i], KapuaFileUtils.getAsURL(permittedFormats[i]));
```
Coduz: assertEquals has an overload method that accepts a message as a first argument. So `fail` invoked in the catch exception is not the correct usage of `Assert` stuff. 
Please change this and also other similar usages.

commons/src/test/java/org/eclipse/kapua/commons/util/KapuaFileUtilsTest.java
```diff
@@ -0,0 +1,92 @@
...
+            } catch (Exception ex) {
+                fail("Inappropriate http scheme.");
+            }
+            try {
+                assertEquals(invalidFormats[i], KapuaFileUtils.getAsURL(invalidFormats[i]));
```
Coduz: If they are not intended to match please use `assertNotEquals`. 
If `KapuaFileUtils.getAsURL(...)` is meant to throw an Exception catch the Exception and do asserts on the type of Exception expected.

Coduz (changes requested):
Hi @code-incubus ,

I've made two comments but those should be changed everywhere the same this is done.

Thanks,

_Alberto_

code-incubus:
Hi @Coduz,
I made all the changes that you mentioned above. If something is not ok please let me know.
I'll fix it up quickly.

LeoNerdoG (approved):
Hey @Coduz,
now this PR looks ok, I think this can be merged.

commons/src/main/java/org/eclipse/kapua/commons/util/KapuaFileUtils.java
```diff
@@ -79,6 +79,7 @@ public static File getAsFile(String filePath) throws KapuaSettingException {
 
             if (hasHttpScheme(filePath)) {
                 // FIXME:
+                // After fixing this method, needs to adjust/fix the tests
```
Coduz: Which is the issue with this method?

code-incubus: @Coduz 
That’s the comment I found in that class. So if that method ever changes or whatever ... it is necessary to adjust the test. That's all.

LeoNerdoG (approved):

Coduz (approved):

PR #2978, opened 2020-05-15
Added Integration Tests For SSO Using Keycloak Provider

zanpizmoht:
Tests are added for SSO login using Keycloak provider.

Signed-off-by: zanpizmoht <zan.pizmoht@comtrade.com>

In this PR are added integration tests for SSO using Keycloak provider.

**Related Issue**
None

There are a few changes in this PR: 
- New testing methods for Docker are added to configure keycloak containers,
- Keycloak configuration is added to `kapua-realm.json` file,
- Added test scenarios and methods that are testing SSO login using Keycloak provider,
- Modified tests for creating user.

/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -925,6 +925,11 @@
             </dependency>
             <dependency>
                 <groupId>org.eclipse.kapua</groupId>
+                <artifactId>kapua-sso-test-steps</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.eclipse.kapua</groupId>
                 <artifactId>kapua-sso-provider</artifactId>
                 <version>${project.version}</version>
             </dependency>
--- a/qa/integration-steps/src/main/java/org/eclipse/kapua/qa/integration/steps/DockerSteps.java
+++ b/qa/integration-steps/src/main/java/org/eclipse/kapua/qa/integration/steps/DockerSteps.java
@@ -31,12 +31,21 @@ import cucumber.api.java.en.Then;
 import cucumber.runtime.java.guice.ScenarioScoped;
 import org.apache.activemq.command.BrokerInfo;
 import org.eclipse.kapua.qa.common.StepData;
+import org.apache.shiro.SecurityUtils;
+import org.apache.shiro.config.Ini;
+import org.apache.shiro.config.IniSecurityManagerFactory;
+import org.apache.shiro.mgt.SecurityManager;
 import org.eclipse.paho.client.mqttv3.MqttException;
 import org.junit.Assert;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.inject.Inject;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -49,17 +58,17 @@ public class DockerSteps {
 
     private static final String NETWORK_PREFIX = "kapua-net";
 
+    private static final StepData CONTAINER_MAP = new StepData();
+
     private DockerClient docker;
 
     private NetworkConfig networkConfig;
 
-    private String networkId;
+    private static String networkId;
 
     private boolean debug;
 
     private List<String> envVar;
-
-    private Map<String, String> containerMap;
 
     public Map<String, Integer> portMap;
 
@@ -75,7 +84,6 @@ public class DockerSteps {
     public DockerSteps(StepData stepData) {
 
         this.stepData = stepData;
-        containerMap = new HashMap<>();
     }
 
     @Given("^Enable debug$")
@@ -195,9 +203,26 @@ public class DockerSteps {
 
         docker.startContainer(containerId);
         docker.connectToNetwork(containerId, networkId);
-        containerMap.put("db", containerId);
+        CONTAINER_MAP.put("db", containerId);
         logger.info("DB container started: {}", containerId);
     }
+
+    @And("^Initialize shiro ini file$")
+    public void initShiro() {
+        // initialize shiro context for broker plugin from shiro ini file
+        final URL shiroIniUrl = getClass().getResource("/shiro.ini");
+        Ini shiroIni = new Ini();
+        try (final InputStream input = shiroIniUrl.openStream()) {
+            shiroIni.load(input);
+        } catch (IOException e) {
+            logger.error("Error in plugin installation.", e);
+            throw new SecurityException(e);
+        }
+
+        SecurityManager securityManager = new IniSecurityManagerFactory(shiroIni).getInstance();
+        SecurityUtils.setSecurityManager(securityManager);
+    }
+
 
     @And("^Start ES container with name \"(.*)\"$")
     public void startESContainer(String name) throws DockerException, InterruptedException {
@@ -208,7 +233,7 @@ public class DockerSteps {
 
         docker.startContainer(containerId);
         docker.connectToNetwork(containerId, networkId);
-        containerMap.put("es", containerId);
+        CONTAINER_MAP.put("es", containerId);
         logger.info("ES container started: {}", containerId);
     }
 
@@ -221,7 +246,7 @@ public class DockerSteps {
 
         docker.startContainer(containerId);
         docker.connectToNetwork(containerId, networkId);
-        containerMap.put(name, containerId);
+        CONTAINER_MAP.put(name, containerId);
         logger.info("EventBroker container started: {}", containerId);
     }
 
@@ -239,14 +264,28 @@ public class DockerSteps {
 
         docker.startContainer(containerId);
         docker.connectToNetwork(containerId, networkId);
-        containerMap.put(bcData.getName(), containerId);
+        CONTAINER_MAP.put(bcData.getName(), containerId);
         logger.info("Message Broker {} container started: {}", bcData.getName(), containerId);
+    }
+
+    @And("^Start Keycloak container with name \"(.*)\"$")
+    public void startKeycloakContainer(String name) throws DockerException, InterruptedException, IOException, URISyntaxException {
+        logger.info("Starting Keycloak container...");
+        ContainerConfig keycloakConfig = getKeycloakContainerConfig();
+        ContainerCreation keycloakContainerCreation = docker.createContainer(keycloakConfig, name);
+        String containerId = keycloakContainerCreation.id();
+
+        docker.copyToContainer(Paths.get(getClass().getResource("/keycloak").toURI()), "keycloak", "/imports");
+        docker.startContainer(containerId);
+        docker.connectToNetwork(containerId, networkId);
+        CONTAINER_MAP.put(name, containerId);
+        logger.info("Keycloak container started: {}", containerId);
     }
 
     @Then("^Stop container with name \"(.*)\"$")
     public void stopContainer(String name) throws DockerException, InterruptedException {
         logger.info("Stopping container {}...", name);
-        String containerId = containerMap.get(name);
+        String containerId = (String) CONTAINER_MAP.get(name);
         docker.stopContainer(containerId, 3);
         logger.info("Container {} stopped.", name);
     }
@@ -254,7 +293,7 @@ public class DockerSteps {
     @Then("^Remove container with name \"(.*)\"$")
     public void removeContainer(String name) throws DockerException, InterruptedException {
         logger.info("Removing container {}...", name);
-        String containerId = containerMap.get(name);
+        String containerId = (String) CONTAINER_MAP.get(name);
         docker.removeContainer(containerId);
         logger.info("Container {} removed.", name);
     }
@@ -403,6 +442,33 @@ public class DockerSteps {
                 .hostConfig(hostConfig)
                 .exposedPorts(String.valueOf(brokerPort))
                 .image("kapua/kapua-events-broker:1.3.0-SNAPSHOT")
+                .build();
+    }
+
+    /**
+     * Creation of docker container configuration for Keycloak provider.
+     *
+     * @return Container configuration for Keycloak instance.
+     */
+    private ContainerConfig getKeycloakContainerConfig() {
+        final int keycloakPort = 8080;
+        final int hostPort = 9090;
+        final Map<String, List<PortBinding>> portBindings = new HashMap<>();
+        addHostPort("0.0.0.0", portBindings, keycloakPort, hostPort);
+        final HostConfig hostConfig = HostConfig.builder().portBindings(portBindings).build();
+        return ContainerConfig.builder()
+                .hostConfig(hostConfig)
+                .exposedPorts(String.valueOf(keycloakPort))
+                .addVolume("imports")
+                .env(
+                        "KEYCLOAK_USER=admin",
+                        "KEYCLOAK_PASSWORD=admin",
+                        "KEYCLOAK_IMPORT=/imports/kapua-realm.json"
+                )
+                .image("jboss/keycloak:8.0.1")
+                .cmd(
+                        "-b 0.0.0.0 -Dkeycloak.import=/imports/kapua-realm.json"
+                )
                 .build();
     }
 
--- a/qa/integration/pom.xml
+++ b/qa/integration/pom.xml
@@ -233,7 +233,14 @@
             <artifactId>kapua-translator-kura-jms</artifactId>
             <scope>provided</scope>
         </dependency>
-
+        <dependency>
+            <groupId>org.eclipse.kapua</groupId>
+            <artifactId>kapua-sso-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.kapua</groupId>
+            <artifactId>kapua-sso-test-steps</artifactId>
+        </dependency>
         <dependency>
             <groupId>org.eclipse.kapua</groupId>
             <artifactId>kapua-transport-api</artifactId>
--- a/qa/integration/src/test/resources/features/sso/SsoLoginUsingGenericProvider.feature
+++ b/qa/integration/src/test/resources/features/sso/SsoLoginUsingGenericProvider.feature
@@ -1,0 +1,18 @@
+###############################################################################
+# Copyright (c) 2020 Eurotech and/or its affiliates and others
+#
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License v1.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v10.html
+#
+# Contributors:
+#     Eurotech - initial API and implementation
+###############################################################################
+@sso
+@integration
+
+Feature: Sso Login using Generic Provider
+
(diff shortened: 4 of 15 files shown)


LeoNerdoG:
Hey @Coduz,
can you please file a CQ for this one? We could not fit it into 1k lines. 
I will also need a bit more time to go through everything, but I will let you know when I finish the review. 😄

sso/test-steps/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+        Red Hat Inc - initial API and implementation
+ -->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>kapua</artifactId>
+        <groupId>org.eclipse.kapua</groupId>
+        <version>1.2.0-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
```
Coduz: Is it `relativePath` wrong? It shouldn't be parent of `kapua-sso`?

qa/integration/pom.xml
```diff
@@ -233,6 +233,15 @@
             <artifactId>kapua-translator-kura-jms</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.kapua</groupId>
+            <artifactId>kapua-sso-api</artifactId>
+        </dependency>
+        <dependency>
```
Coduz, line 240: Please define the version of this artifact on the root `pom.xml` in the `dependencyManagement` section and use in here only `groupId` and `artifactId`

Coduz (changes requested):

qa/integration/pom.xml
zanpizmoht: So the line `<version>1.2.0-SNAPSHOT</version>` should then be in file kapua\qa\pom.xml ?

Coduz: On the root pom, like you did! 👍

sso/test-steps/pom.xml
```diff
@@ -0,0 +1,116 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Copyright (c) 2020 Red Hat Inc and others
```
Coduz: The Copyright must report Eurotech here and also few lines below.

sso/test-steps/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+        Red Hat Inc - initial API and implementation
+ -->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>kapua</artifactId>
+        <groupId>org.eclipse.kapua</groupId>
+        <version>1.2.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
```
Coduz: Just remove this since this is the default value

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
```diff
@@ -111,6 +117,39 @@
     private final AccessPermissionService accessPermissionService = locator.getService(AccessPermissionService.class);
     private final AccessPermissionFactory accessPermissionFactory = locator.getFactory(AccessPermissionFactory.class);
 
+    static {
```
Coduz: We can no longer init shiro via a `static` into the `AuthenticatinServiceShiroImpl`.

Do you need this init otherwise test will not work properly?

Coduz (changes requested):

sso/test-steps/pom.xml
zanpizmoht: Yes it was wrong. Now it is changed.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
zanpizmoht: Yes, this is needed for the tests to work properly. Do you think there is any other way to introduce this configuration?

Coduz: Some time ago we ditched the usage of the `static` init in order to use the initialization via `shiro.ini` file. 
I think a `shiro.ini` file must be created under the `test/resources` directory of the test.

riccardomodanese: Hi @zanpizmoht as @Coduz said static Shiro initialization was removed by #2905 
Since this qa tests are using the embedded environment the security context should be initialized by the broker (or the REST api, if any) during the startup and available for the test classes also.
If your test doesn't need embedded broker or REST api, so the Shiro context is not initialized, please use this approach: [cucumber Shiro init](https://github.com/eclipse/kapua/commit/c39b6856d2b3689784a68210a11428a3d1d81fd5#diff-0959952e566e7430b191a2c674b1352aR18)
The "Init Security Context" step does the Shiro context initilization and, at the end of the test, the step "Reset Security Context for all scenarios" reset the Shiro context.
There is a probably not needed shiro.ini file in the qa path (qa-common) and @lorthirk is working on it.
So please, remove the static initialization and try adding the step I suggested to your test and give me a feedback.

lorthirk: I can confirm that the `shiro.ini` under `kapua-qa/common/src/main/resources` is not needed. In [this branch](https://github.com/lorthirk/kapua/tree/change-removeQaCommonShiroIni) I attempted to remove it, and as you can see from [Travis](https://travis-ci.com/github/lorthirk/kapua/builds/167098920) the build was successful.

zanpizmoht: I have fixed this issue by removing static configuration and added shiro configuration to the Cucumber test scenarios

zanpizmoht:
Hey @Coduz,
I made all the requested changes if you can take a look and approve those changes.

pom.xml
```diff
@@ -924,6 +924,11 @@
                 <artifactId>kapua-sso-api</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
```
LeoNerdoG, line 926: Please change header to: "    Copyright (c) 2016, 2020 Eurotech and/or its affiliates and others"

qa/integration-steps/src/main/java/org/eclipse/kapua/qa/integration/steps/DockerSteps.java
```diff
@@ -31,12 +31,21 @@
 import cucumber.runtime.java.guice.ScenarioScoped;
 import org.apache.activemq.command.BrokerInfo;
 import org.eclipse.kapua.qa.common.StepData;
+import org.apache.shiro.SecurityUtils;
```
LeoNerdoG, line 34: Please change header to: " * Copyright (c) 2019, 2020 Eurotech and/or its affiliates and others"

qa/integration/pom.xml
```diff
@@ -233,7 +233,14 @@
             <artifactId>kapua-translator-kura-jms</artifactId>
             <scope>provided</scope>
         </dependency>
-
+        <dependency>
```
LeoNerdoG, line 236: Please change header to: "    Copyright (c) 2017, 2020 Eurotech and/or its affiliates and others"

sso/test-steps/src/main/java/org/eclipse/kapua/service/sso/steps/SsoServiceSteps.java
```diff
@@ -0,0 +1,211 @@
+/*******************************************************************************
```
LeoNerdoG, line 1: Please change complete header to: 
/*******************************************************************************
 * Copyright (c) 2020 Eurotech and/or its affiliates and others
 *
 * All rights reserved. This program and the accompanying materials
 * are made available under the terms of the Eclipse Public License v1.0
 * which accompanies this distribution, and is available at
 * http://www.eclipse.org/legal/epl-v10.html
 *
 * Contributors:
 *     Eurotech - initial API and implementation

LeoNerdoG (changes requested):
hey @zanpizmoht,
this is first par of my review - only headers. Please take a look and fix them 😄

qa/integration/src/test/resources/features/sso/SsoLoginUsingGenericProvider.feature
```diff
@@ -0,0 +1,18 @@
...
+# Contributors:
+#     Eurotech - initial API and implementation
+###############################################################################
+@sso
+@integration
+
+Feature: Sso Login using Generic Provider
+
+  Scenario: Configure environment for SSO with the Generic provider
+    Given Set environment variables for generic provider
```
gbarbon, line 18: Hi @zanpizmoht , I think that the tests scenarios defined in `SsoLoginUsingKeycloakProvider.feature` should be added also here, in order to carry the tests also with the generic provider. I'm not 100% sure, but maybe we can find a way to reuse the same scenarios without simply copying the code here.

sso/test-steps/src/main/java/org/eclipse/kapua/service/sso/steps/SsoServiceSteps.java
```diff
@@ -0,0 +1,211 @@
...
+    public void setEnvVariablesForKeycloakProvider() throws UnknownHostException {
+        InetAddress address = InetAddress.getLocalHost();
+//        System.setProperty("sso.keycloak.uri", address.getHostAddress() + ":9090");
+        System.setProperty("sso.keycloak.uri", "http://localhost:9090");
+        System.setProperty("site.home.uri", "http://localhost:8080");
+        System.setProperty("sso.keycloak.realm", "kapua");
+        System.setProperty("sso.openid.client.id", "console");
+        System.setProperty("sso.provider", "keycloak");
+        System.setProperty("authentication.registration.service.enabled", "true");
+    }
```
gbarbon: I would prefer setting the `authentication.registration.service.enabled` property as `true` only for the scenarios that actually use the `RegistrationServiceImpl`, otherwise this kind of registration is enabled for each scenario.

qa/integration/src/test/resources/features/sso/SsoLoginUsingKeycloakProvider.feature
```diff
@@ -0,0 +1,167 @@
...
+      | boolean | infiniteChildEntities  | true  |
+      | integer | maxNumberChildEntities | 5     |
+    When A generic user
+      | name      | displayName | email           | phoneNumber     | status  | userType |
+      | test-user | Test User   | test-user@co.co | +386 31 323 444 | ENABLED | INTERNAL |
+    And I add credentials
+      | name      | password          | enabled |
+      | test-user | ToManySecrets123# | true    |
+    Then I logout
+    Then Get access token for user with username "test-user" and password "TestPassword123#"
```
gbarbon: I think that here the password should be the same as in the "I add credentials" step (just to be coherent with the scenario, even if it shouldn't modify the expected result)

Coduz:
Ping ping on the comments from @gbarbon

qa/integration/pom.xml
zanpizmoht: I changed all the requested headers.

sso/test-steps/src/main/java/org/eclipse/kapua/service/sso/steps/SsoServiceSteps.java
zanpizmoht: This should now be resolved. I added this line to the other method so it is only used when `RegistrationServiceImpl `is used.

qa/integration/src/test/resources/features/sso/SsoLoginUsingKeycloakProvider.feature
zanpizmoht: I changed the password so it is the same through the whole scenario.

qa/integration/src/test/resources/features/sso/SsoLoginUsingGenericProvider.feature
zanpizmoht: I will look into the tests for generic provider when this PR will be merged.

sso/test-steps/src/main/java/org/eclipse/kapua/service/sso/steps/SsoServiceSteps.java
gbarbon: Hi @zanpizmoht , ok. Just a question concerning the `configureTheSso()` method, I saw that this is used also when the `RegistrationService` (or the `SimpleRegistrationProcessor`) is not mentioned in the scenario description. Is it because this kind of registration is used also for other tests?

zanpizmoht: Which scenario do you have in mind?

gbarbon: For instance _'Using normal login using Keycloak credentials Create user using keycloak Login as external user using normal login method Exception should be thrown'_, _'Logging out and logging back in with SSO Login as external user, logout Login again, no exception should be thrown'_ and _'Login using normal login after logging out with SSO Login as external user, logout Login again, this time with normal login method, exception should be thrown'_ . Do these scenarios also require the `RegistrationServiceImpl` ?

zanpizmoht: Yes both scenarios are requiring the `RegistrationServiceImpl` class because they use external user login.

gbarbon: But the `RegistrationServiceImpl` should be needed only for testing the `SimpleRegistrationProcessor` (automatic creation of the user in Kapua at the first log-in attempt). In the other cases we would like to test the external user login with existing (external) users, and such login method shouldn't require the `RegistrationServiceImpl`.

zanpizmoht: Yes I know, this is also covered in the scenario 'Create external user using the the UserService'. This is just the basic scenario, I will add more when this PR is merged.

gbarbon: Ok. If you are using the `RegistrationServiceImpl` also in the scenarios above, it would be better to mention that the external user is created via `SimpleRegistrationProcessor`, like you did in another scenario. Since the `RegistrationServiceImpl` can trigger the creation of new users in Kapua that can login even if they didn't exist before (I guess this is the case for the test scenario _'Logging out and logging back in with SSO Login as external user, logout Login again, no exception should be thrown'_), I would prefer to explicitly state when it is used. In this way we will be able to distinguish these scenarios from the ones that won't use the `RegistrationServiceImpl`.

zanpizmoht: You are right, I will add this to the descriptions of those scenarios.

zanpizmoht:
Hey @Coduz, I made all the changes that were discussed. Travis tests passed. So I think this PR is ready to be filed to CQ bacause it has more than 1k changed lines. For now I am leaving all commits (so it is easier to find changes) and will later squash them all into one.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/JwtAuthenticatingRealm.java
```diff
@@ -190,6 +190,9 @@ protected AuthenticationInfo doGetAuthenticationInfo(AuthenticationToken authent
     private String extractExternalId(String jwt) {
         final String id;
         try {
+            if (jwtProcessor == null) {
+                init();
```
gbarbon: Hi @zanpizmoht , just a question about this init, are you sure we need it? The `onInit()` method of a shiro realm is called while initialising all the realms, so if the JwtAuthenticationRealm is declared in the `shiro.ini` file (as it is in our case), this method should also be called. Thus we should never find the `jwtProcessor` set to `null`.

zanpizmoht: Hey Gianluca, I remember I had problems with this because the jwtProcessor was not initialised. So I discussed this with Alberto and he said I can just add this checker to prevent the jwtProcessor being null. I will double check again if this is needed now (maybe something changed) and will get back to you.

qa/integration/src/test/resources/shiro.ini
```diff
@@ -27,13 +27,20 @@ kapuaUserPassAuthenticatingRealm = org.eclipse.kapua.service.authentication.shir
...
+jwtAuthenticatingRealm = org.eclipse.kapua.service.authentication.shiro.realm.JwtAuthenticatingRealm
+
 ########################
 #Authorization section #
 ########################
 # Authorization
 kapuaAuthorizingRealm = org.eclipse.kapua.service.authorization.shiro.KapuaAuthorizingRealm
+
+# Jwt
+jwtAuthenticatingRealm = org.eclipse.kapua.service.authentication.shiro.realm.JwtAuthenticatingRealm
```
gbarbon: Hi @zanpizmoht , I think this second `jwtAuthenticatingRealm` definition is redundant, since there's also one at line 31.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/JwtAuthenticatingRealm.java
gbarbon: Yes, I remember some discussion about it. However while testing with a real console I never encountered a similar problem. I wonder if maybe at the time you tried there were no `jwtAuthenticatingRealm` defined in the `shiro.ini` file of the QAs (while now it is defined), so this might be the cause.

zanpizmoht: I have checked it and those lines are not needed as you said so I removed them now.

qa/integration/src/test/resources/shiro.ini
zanpizmoht: Thanks Gianluca, I have fixed that and am now resolving this comment.

sonja-ct:
Hey @Coduz,
my new PR #3122 covers these tests, so you can close this PR when you have time. Thanks

PR #2981, opened 2020-05-17
Sort order parameter for data messages request

xshiart:
The new parameter for /data/messages endpoint to control the sorting order of returned results.
This change improves REST API usage in case of iterating over messages.

### DIFF
--- a/rest-api/resources/src/main/java/org/eclipse/kapua/app/api/resources/v1/resources/DataMessages.java
+++ b/rest-api/resources/src/main/java/org/eclipse/kapua/app/api/resources/v1/resources/DataMessages.java
@@ -12,7 +12,6 @@
 package org.eclipse.kapua.app.api.resources.v1.resources;
 
 import com.google.common.base.Strings;
-
 import org.eclipse.kapua.KapuaException;
 import org.eclipse.kapua.app.api.resources.v1.resources.model.CountResult;
 import org.eclipse.kapua.app.api.resources.v1.resources.model.DateParam;
@@ -28,15 +27,19 @@ import org.eclipse.kapua.service.datastore.MessageStoreService;
 import org.eclipse.kapua.service.datastore.client.model.InsertResponse;
 import org.eclipse.kapua.service.datastore.internal.mediator.ChannelInfoField;
 import org.eclipse.kapua.service.datastore.internal.mediator.MessageField;
+import org.eclipse.kapua.service.datastore.internal.schema.MessageSchema;
 import org.eclipse.kapua.service.datastore.model.DatastoreMessage;
 import org.eclipse.kapua.service.datastore.model.MessageListResult;
 import org.eclipse.kapua.service.datastore.model.query.AndPredicate;
 import org.eclipse.kapua.service.datastore.model.query.MessageQuery;
 import org.eclipse.kapua.service.datastore.model.query.RangePredicate;
+import org.eclipse.kapua.service.datastore.model.query.SortField;
+import org.eclipse.kapua.service.datastore.model.query.SortDirection;
 import org.eclipse.kapua.service.datastore.model.query.StorableFetchStyle;
 import org.eclipse.kapua.service.datastore.model.query.StorablePredicate;
 import org.eclipse.kapua.service.datastore.model.query.StorablePredicateFactory;
 import org.eclipse.kapua.service.datastore.model.query.TermPredicate;
+
 
 import javax.ws.rs.Consumes;
 import javax.ws.rs.DefaultValue;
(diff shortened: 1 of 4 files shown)


rest-api/resources/src/main/java/org/eclipse/kapua/app/api/resources/v1/resources/DataMessages.java
```diff
@@ -88,7 +93,8 @@
             @QueryParam("metricMin") String metricMinValue, //
             @QueryParam("metricMax") String metricMaxValue, //
             @QueryParam("offset") @DefaultValue("0") int offset,//
-            @QueryParam("limit") @DefaultValue("50") int limit) throws KapuaException {
+            @QueryParam("limit") @DefaultValue("50") int limit,
+            @QueryParam("order") @DefaultValue("desc") String order) throws KapuaException {
```
lorthirk: A few of things:

- Instead of a `String`, you could use the `SortDirection` type itself so it's automatically parsed, and invalid values are rejected earlier. Also, to have it case-insensitive, please implement the `fromString()` method in `SortDirection`. You can draw inspiration from #2891
- Also, for coherence with #2891, I would rather name the parameter `sortDir`, and the default should be in uppercase
- This is just a stylistic preference, but I'd like to keep `offset` and `limit` to end of the signature, so could you please move `sortDir` just below `metricMax`?

lorthirk (changes requested):
Thanks for the submission @xshiart!

I would kindly ask for a few changes, and also it would be great if you could update the OpenAPI specification under `rest-api/resources/src/main/resources/openapi`.

I'll gladly merge this as soon as it's ready!

rest-api/resources/src/main/java/org/eclipse/kapua/app/api/resources/v1/resources/DataMessagesJson.java
```diff
@@ -91,7 +91,8 @@
             @QueryParam("metricMin") String metricMinValue,
             @QueryParam("metricMax") String metricMaxValue,
             @QueryParam("offset") @DefaultValue("0") int offset,
-            @QueryParam("limit") @DefaultValue("50") int limit) throws KapuaException {
+            @QueryParam("limit") @DefaultValue("50") int limit,
+            @QueryParam("order") @DefaultValue("desc") String order) throws KapuaException {
```
lorthirk: Of course, here you should reflect all the changes from `DataMessages.java`

lorthirk (approved):

rest-api/resources/src/main/resources/openapi/dataMessage/dataMessage-scopeId.yaml
```diff
@@ -67,6 +67,15 @@ paths:
...
             type: string
+        - description: The sort direction. Can be ASCENDING, DESCENDING (default). Case-insensitive.
+          name: sortDir
+          in: query
+          schema:
+            type: string
+            enum:
+              - ASCENDING
+              - DESCENDING
+            default: DESCENDING
```
lorthirk: Aren't these `ASC` and `DESC`?

lorthirk (changes requested):

Coduz (approved):

rest-api/resources/src/main/resources/openapi/dataMessage/dataMessage-scopeId.yaml
```diff
@@ -67,6 +67,15 @@ paths:
           in: query
           schema:
             type: string
+        - description: The sort direction. Can be ASCENDING, DESCENDING (default). Case-insensitive.
```
lorthirk: Whoops... this slipped in as well. `ASC` and `DESC` here too. Just one to go!

lorthirk (changes requested):
Just one last change! Sorry for not catching it earlier!

rest-api/resources/src/main/resources/openapi/dataMessage/dataMessage-scopeId.yaml
greatbalin: It is a general description, those two have the general meaning.

xshiart: This is a generic description and the meaning of those two words is generic as well.
They are not connected to actual enum members.

Coduz: Hi! 

That makes sense. Since they are generic description maybe those should be lowercase? So they will not be confused as values that can be used!

Thanks,

_Alberto_

lorthirk: I agree with @Coduz. I would rather write them lowercase, and also remove the reference of them being case insensitive.

lorthirk (approved):
Thanks for the submission!

Coduz:
All looks good so I'll proceed to merge this.
This might be something that we will make available into 1.2.x Kapua

PR #2982, opened 2020-05-20
Security plugin cleanup

riccardomodanese:
Security broker filter and related objects cleanup

**Related Issue**
none

KapuaConnectionContext and KapuaSecurityContext have partially overlapped role inside the authentication feature. Will be merged into one object (KapuaSecurityContext).

**Screenshots**
none

Some internal methods signature is changed since those based on KapuaConnectionContext are changed using the KapuaSecurityContext.


LeoNerdoG (commented):
Hey @riccardomodanese,
I noticed that there are plenty of files where header is not changed to "2020" - can you change this so we have files updated to date? (I think tehre is no point in writing comments in all the files sepparatelly, so I am writing here).

riccardomodanese:
I'll check, I missed some file may be.
Don't take in account the changes coming from commit 613adc4. It will be removed before merging the pr.

riccardomodanese:
moved to a new branch

PR #2992, opened 2020-05-29
Fixed missing OpenID Provider session invalidation

### LINKED ISSUE #2971: SSO provider session not invalidated at Kapua session invalidation
**Describe the bug**
When the session expires, the user is logged out from the Kapua console. However, if the user is an external one logged through SSO, he can login again by simply clicking on the 'SSO Login' button, without the need to insert again his credentials. This is due to the SSO session not yet invalidated, while the Kapua Console one already is.

One possible solution would be to simply reduce the time of the SSO session to a lower value than the Kapua Console one. Note that the Kapua Console session time is set though the `session-timeout` tag in the `web.xml` in `console/web/src/main/webapp/WEB-INF`. However, this is not the best option, since there is no synchronisation between the two session. Also, it requires to configure the SSO Provider session timeout at each installation. A better option would be to invalidate the SSO session using the SSO OpenId Connect logout endpoint when the Kapua one is invalidated.

**To Reproduce**
1. Login to Kapua with an external SSO user (an SSO provider is required)
2. Wait 30 minutes (the Kapua session timeout), after this amount of time the user should be logged out
3. Click on the 'SSO Login' button
4. The user is logged in without the need to insert credentials.

**Expected behavior**
The SSO session should be invalidated at the same time as the Kapua one.

**Screenshots**
_N/A_

**Version of Kapua**
1.2.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[x] Docker
[ ] Others

**Main component affected**
[ ] REST API
[ ] Message Broker
[ ] - Others

**Additional context**
_N/A_

gbarbon:
Logout from the OpenID Connect Provider in case the Kapua Console session is expired.

**Related Issue**
This PR fixes #2971 

Added an HttpSessionListener that listens to the Kapua Console session invalidation. When the session is invalidated, it performs an HTTP request to the OpenID Provider logout endpoint.

Note that this does not follow the same flow as the normal OpenID logout. Moreover, this feature can be disabled with a dedicated property called `console.sso.openid.session_listener.logout.enabled`.

**Screenshots**
_N/A_

Also updated comments and removed unused variables that are part of the SSO code.

This PR requires PR #2996 to be merged!

### DIFF
--- a/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/CredentialsFactoryImpl.java
+++ b/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/CredentialsFactoryImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2016 Eurotech and/or its affiliates and others
+ * Copyright (c) 2011, 2020 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
@@ -39,8 +39,8 @@ public class CredentialsFactoryImpl implements CredentialsFactory {
     }
 
     @Override
-    public JwtCredentials newJwtCredentials(String jwt) {
-        return new JwtCredentialsImpl(jwt);
+    public JwtCredentials newJwtCredentials(String jwt, String idToken) {
+        return new JwtCredentialsImpl(jwt, idToken);
     }
 
     @Override
--- a/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/JwtCredentialsImpl.java
+++ b/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/JwtCredentialsImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2017 Eurotech and/or its affiliates and others
+ * Copyright (c) 2011, 2020 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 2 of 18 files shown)


service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/CredentialsFactoryImpl.java
```diff
@@ -39,8 +39,8 @@ public ApiKeyCredentials newApiKeyCredentials(String apiKey) {
     }
 
     @Override
-    public JwtCredentials newJwtCredentials(String jwt) {
-        return new JwtCredentialsImpl(jwt);
+    public JwtCredentials newJwtCredentials(String jwt, String idToken) {
```
LeoNerdoG, line 42: Hey @gbarbon,
can you change the header of this file to " * Copyright (c) 2011, 2020 Eurotech and/or its affiliates and others" ? 
Because last change is from the current year 😄

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/JwtCredentialsImpl.java
```diff
@@ -19,9 +19,11 @@
     private static final long serialVersionUID = -5920944517814926028L;
 
     private String jwt;
+    private String idToken;
```
LeoNerdoG, line 22: Hey @gbarbon,
can you change the header of this file to " * Copyright (c) 2011, 2020 Eurotech and/or its affiliates and others" ? 
Because last change is from the current year 😄

LeoNerdoG (commented):
Hey @gbarbon sorry to bother you again - there are some more headers to fix 😄

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/CredentialsFactoryImpl.java
gbarbon: Sure, done!

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/JwtCredentialsImpl.java
gbarbon: Sure, done!

PR #3008, opened 2020-07-02
Added Junit Tests

sonja-ct:
Added Junit tests for EventStoreDAO class
Added Junit tests for EventStoreFactoryImpl class
Added Junit tests for EventStoreRecordImpl class
Added Junit tests for EventStoreServiceImpl class

Signed-off-by: Sonja <sonja.matic@comtrade.com>

**Related Issue**
/

/

**Screenshots**
/

/

### DIFF
--- a/commons/src/test/java/org/eclipse/kapua/commons/service/event/store/internal/EventStoreRecordImplTest.java
+++ b/commons/src/test/java/org/eclipse/kapua/commons/service/event/store/internal/EventStoreRecordImplTest.java
@@ -1,0 +1,437 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Eurotech and/or its affiliates and others
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Eurotech - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.kapua.commons.service.event.store.internal;
+
+import org.eclipse.kapua.KapuaException;
+import org.eclipse.kapua.commons.model.id.KapuaEid;
+import org.eclipse.kapua.commons.service.event.store.api.EventStoreRecord;
+import org.eclipse.kapua.event.ServiceEvent;
+import org.eclipse.kapua.model.id.KapuaId;
+import org.eclipse.kapua.model.id.KapuaIdStatic;
+import org.eclipse.kapua.qa.markers.junit.JUnitTests;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.mockito.Mockito;
+
+import java.math.BigInteger;
+import java.util.Date;
+
+@Category(JUnitTests.class)
+public class EventStoreRecordImplTest extends Assert {
+
+    EventStoreRecord eventStoreRecord;
+
+    @Before
+    public void initialize() {
+        eventStoreRecord = Mockito.mock(EventStoreRecord.class);
+
+        Mockito.when(eventStoreRecord.getContextId()).thenReturn("contextId");
+        Mockito.when(eventStoreRecord.getTimestamp()).thenReturn(new Date());
+        Mockito.when(eventStoreRecord.getUserId()).thenReturn(new KapuaEid(BigInteger.TEN));
+        Mockito.when(eventStoreRecord.getService()).thenReturn("service");
+        Mockito.when(eventStoreRecord.getEntityType()).thenReturn("entityType");
+        Mockito.when(eventStoreRecord.getEntityId()).thenReturn(new KapuaEid(BigInteger.TEN));
+        Mockito.when(eventStoreRecord.getOperation()).thenReturn("operation");
+        Mockito.when(eventStoreRecord.getInputs()).thenReturn("input");
+        Mockito.when(eventStoreRecord.getOutputs()).thenReturn("output");
+        Mockito.when(eventStoreRecord.getStatus()).thenReturn(ServiceEvent.EventStatus.TRIGGERED);
+        Mockito.when(eventStoreRecord.getNote()).thenReturn("note");
+    }
+
+    @Test
+    public void eventStoreRecordImplTest1() {
+        EventStoreRecordImpl eventStoreRecordImpl = new EventStoreRecordImpl();
+        assertEquals("Expected and actual values should be the same.", ServiceEvent.EventStatus.TRIGGERED, eventStoreRecordImpl.getStatus());
+    }
+
+    @Test
+    public void eventStoreRecordImplTest2() {
+        KapuaId[] scopeIdList = {null, new KapuaIdStatic(BigInteger.ONE)};
+
+        for (KapuaId scopeId : scopeIdList) {
+            EventStoreRecordImpl eventStoreRecordImpl = new EventStoreRecordImpl(scopeId);
+            assertEquals("Expected and actual values should be the same.", scopeId, eventStoreRecordImpl.getScopeId());
+        }
+    }
+
+    @Test
+    public void eventStoreRecordImplTest3() throws KapuaException {
+        EventStoreRecordImpl eventStoreRecordImpl = new EventStoreRecordImpl(eventStoreRecord);
+
+        assertEquals("Expected and actual values should be the same.", "contextId", eventStoreRecordImpl.getContextId());
+        assertEquals("Expected and actual values should be the same.", new Date().toString(), eventStoreRecordImpl.getTimestamp().toString());
+        assertEquals("Expected and actual values should be the same.", new KapuaEid(BigInteger.TEN), eventStoreRecordImpl.getUserId());
+        assertEquals("Expected and actual values should be the same.", "service", eventStoreRecordImpl.getService());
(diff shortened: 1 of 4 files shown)


LeoNerdoG (approved):
Hey @Coduz,
I have checked this PR and it looks OK - can you please recheck and merge when possible? 
Tests have passed and codeCov has no complaints.

commons/src/test/java/org/eclipse/kapua/commons/service/event/store/internal/EventStoreRecordImplTest.java
```diff
@@ -0,0 +1,464 @@
...
+        assertThat("EventStoreRecordImpl object expected.", eventStoreRecordImpl, IsInstanceOf.instanceOf(EventStoreRecordImpl.class));
+    }
+
+    @Test
+    public void eventStoreRecordImplTest2() {
+        KapuaId[] scopeIdList = {null, new KapuaIdStatic(BigInteger.ONE), new KapuaIdStatic(BigInteger.TEN), new KapuaIdStatic(BigInteger.ZERO)};
+
+        for (KapuaId scopeId : scopeIdList) {
+            EventStoreRecordImpl eventStoreRecordImpl = new EventStoreRecordImpl(scopeId);
+            assertNotNull("Null not expected.", eventStoreRecordImpl);
```
Coduz: This test is always true. 
Is not possible to have a `null` object after a `new`, and if `new EventStoreRecordImpl(scopeId)` goes on exception, this `assertNotNull` is not executed.

commons/src/test/java/org/eclipse/kapua/commons/service/event/store/internal/EventStoreRecordImplTest.java
```diff
@@ -0,0 +1,464 @@
...
+        String contextId = "ContextId";
+        KapuaId[] scopeIdList = {null, new KapuaIdStatic(BigInteger.ONE), new KapuaIdStatic(BigInteger.TEN), new KapuaIdStatic(BigInteger.ZERO)};
+
+        for (KapuaId scopeId : scopeIdList) {
+            EventStoreRecordImpl eventStoreRecordImpl1 = new EventStoreRecordImpl(scopeId);
+            assertNull("Null expected.", eventStoreRecordImpl1.getContextId());
+            eventStoreRecordImpl1.setContextId(contextId);
+            assertEquals("Expected and actual values should be the same.", contextId, eventStoreRecordImpl1.getContextId());
+            eventStoreRecordImpl1.setContextId(null);
+            assertNull("Null expected.", eventStoreRecordImpl1.getContextId());
```
Coduz: I think that testing for different values of scopeId is not meaningful.
We want to test only once test and get.

commons/src/test/java/org/eclipse/kapua/commons/service/event/store/internal/EventStoreRecordImplTest.java
```diff
@@ -0,0 +1,464 @@
...
+        assertEquals("Expected and actual values should be the same.", contextId, eventStoreRecordImpl2.getContextId());
+        eventStoreRecordImpl2.setContextId(null);
+        assertNull("Null expected.", eventStoreRecordImpl2.getContextId());
+
+        EventStoreRecordImpl eventStoreRecordImpl3 = new EventStoreRecordImpl(eventStoreRecord);
+        assertEquals("Expected and actual values should be the same.", "contextId", eventStoreRecordImpl3.getContextId());
+        eventStoreRecordImpl3.setContextId(contextId);
+        assertEquals("Expected and actual values should be the same.", contextId, eventStoreRecordImpl3.getContextId());
+        eventStoreRecordImpl3.setContextId(null);
+        assertNull("Null expected.", eventStoreRecordImpl3.getContextId());
```
Coduz, line 111: The clone method has been already been tested in `eventStoreRecordImplTest3`, if I'm correct.
So also this is not needed.

qa/integration/src/test/java/org/eclipse/kapua/integration/misc/EventStoreDAOTest.java
```diff
@@ -0,0 +1,296 @@
...
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.List;
+
+@Category(JUnitTests.class)
+public class EventStoreDAOTest extends Assert {
+
+    @Test
+    public void eventStoreDAOTest() throws Exception {
+        Constructor<EventStoreDAO> eventStoreDAO = EventStoreDAO.class.getDeclaredConstructor();
```
Coduz: EventStoreDAO and other EntityDAO are static and testing the instantiation is not needed.

qa/integration/src/test/java/org/eclipse/kapua/integration/misc/EventStoreDAOTest.java
```diff
@@ -0,0 +1,296 @@
...
+import javax.persistence.criteria.CriteriaBuilder;
+import javax.persistence.criteria.Root;
+import java.lang.reflect.Constructor;
+import java.math.BigInteger;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.List;
+
+@Category(JUnitTests.class)
+public class EventStoreDAOTest extends Assert {
```
Coduz, line 49: Why mock DAO methods when there is an actual implementation available and working?

Coduz (changes requested):
Hi @sonja-ct ,

I've made some comments. I've commented only first occurrence of many things, but I would like to have that fixed on all places (i.e: all places where:
```
EventStoreRecordImpl eventStoreRecordImpl = new EventStoreRecordImpl(scopeId);
assertNotNull("Null not expected.", eventStoreRecordImpl);
```
is done.

There is also a question on why `EntityManager` has been mocked.

Thanks,

- Alberto

commons/src/test/java/org/eclipse/kapua/commons/service/event/store/internal/EventStoreRecordImplTest.java
LeoNerdoG: @sonja-ct I suggest that you check here if the created object is of right instance (for example we are expecting "X" instance and if in future this changes so the method returns instance of an object "Y" we will have a test that will check this. 
assertNotNull is generally used only when you ahve two outcomes, Null and some not null value. It can also be a part of some assertion list (several assertions that check various things). I agree with @Coduz on this one.

commons/src/test/java/org/eclipse/kapua/commons/service/event/store/internal/EventStoreRecordImplTest.java
LeoNerdoG: I agree. We can test only with "null" and BigInteger.ONE for example.

qa/integration/src/test/java/org/eclipse/kapua/integration/misc/EventStoreDAOTest.java
LeoNerdoG: @Coduz we should always mock as much methods as possible to isolate unit test as much as possible. 
We don't want actual implementation of other methods to imterfere with our Unit tests.

commons/src/test/java/org/eclipse/kapua/commons/service/event/store/internal/EventStoreRecordImplTest.java
LeoNerdoG: @Coduz here we are checking if the contextId is null and in the "eventStoreRecordImplTest3" we check if the expected and actual values are the same, so these are two different cases. If I am seeing right.

LeoNerdoG (changes requested):
Hey @Coduz I have finished my review. 
Please look at the comments, if you have something to add, please let us know. Otherwise @sonja-ct  you can start fixing the tests or rather the comments that have not been addressed yet.

commons/src/test/java/org/eclipse/kapua/commons/service/event/store/internal/EventStoreRecordImplTest.java
Coduz: But a result of a `new` is always a `notNull` object, unless Exception, which is not covered anyway by this `assertNotNull`. This is the same thing as writing `assertTrue(true)`. 
If `new` returns a `null`, well Java must be a bit stoned or something 😂

commons/src/test/java/org/eclipse/kapua/commons/service/event/store/internal/EventStoreRecordImplTest.java
Coduz: Yes good

commons/src/test/java/org/eclipse/kapua/commons/service/event/store/internal/EventStoreRecordImplTest.java
Coduz: I wouldn't also mind to check for the right instance type.
Doing:
```
EventStoreRecordImpl eventStoreRecordImpl = new EventStoreRecordImpl(scopeId);
assertTrue(eventStoreRecordImpl instence of EventStoreRecordImpl)
```
is useless for the same reasons.
Also this:
```
EventStoreRecordImpl eventStoreRecordImpl = new EventStoreRecordImpl(scopeId);
assertTrue(eventStoreRecordImpl instence of EventStoreRecord)
```
is useless. In both cases, if that isn't true ClassCastException would be `throw`n when assigning the `new` result to the local variable.

sonja-ct:
Hey @Coduz,
I made all the requested changes, if you can take a look, I hope everything is ok now. 
Thanks, Sonja

Coduz (approved):

PR #3011, opened 2020-07-03
Cache manager review the cache instantiation process

riccardomodanese:

Brief description of the PR.
Keep CachingProvider and CacheManager instance reference in KapuaCacheManager class

**Related Issue**
none

The process to create a cache instance involves the creation of a CachingProvider and a CacheManager.
The previous implementation creates a new instance of both of them for each cache instantiation.
Now the CachingProvider and CacheManager are instantiated once and kept as reference for the further getCache invocations.
This is perfectly allowed and shouldn't change any cache behavior since from CachingProvider.getCacheManager method javadoc (JCache specification):
"Requests a CacheManager configured according to the implementation specific URI that uses the provided ClassLoader for loading underlying classes.
Multiple calls to this method with the same URI and ClassLoader must return the same CacheManager instance, except if a previously returned CacheManager has been closed."

**Screenshots**
none

none

### DIFF
--- a/commons/src/main/java/org/eclipse/kapua/commons/service/internal/cache/KapuaCacheManager.java
+++ b/commons/src/main/java/org/eclipse/kapua/commons/service/internal/cache/KapuaCacheManager.java
@@ -11,8 +11,11 @@
  *******************************************************************************/
 package org.eclipse.kapua.commons.service.internal.cache;
 
+import org.apache.commons.lang3.StringUtils;
 import org.eclipse.kapua.KapuaErrorCodes;
+import org.eclipse.kapua.KapuaException;
 import org.eclipse.kapua.KapuaRuntimeException;
+import org.eclipse.kapua.commons.metric.MetricServiceFactory;
 import org.eclipse.kapua.commons.setting.KapuaSettingException;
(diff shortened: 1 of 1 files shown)


commons/src/main/java/org/eclipse/kapua/commons/service/internal/cache/KapuaCacheManager.java
```diff
@@ -97,39 +119,74 @@ private static URI getCacheConfig() {
...
+    }
+
     /**
      * Utility method to cleanup the whole cache.
      */
     public static void invalidateAll() {
-        CACHE_MAP.forEach((cacheKey, cache) -> cache.clear());
+        CACHE_MAP.forEach((cacheKey, cache) -> {
+            cache.clear();
+            registeredCache.dec();
```
stefanomorson, line 188: I don't think registeredCache.dec() is enough, it should be set to zero, am I wrong ?

stefanomorson: My fault it's in the loop.

stefanomorson (approved):
Looks good for me.

Issue #1537, opened 2018-03-23
Why my RESTful API performance so slow


Beritra:
Am I doing something wrong?

It's my post code:

> GET /v1/AQ/accounts?offset=0&amp;limit=50 HTTP/1.1
Host: 172.16.0.151:8081
Content-Type: application/json
Accept: application/json
Authorization: Bearer eyJhbGciOiJSUzI1NiJ9.eyJpc3MiOiJodHRwczovL3d3dy5lY2xpcHNlLm9yZy9rYXB1YSIsImlhdCI6MTUyMTc3MjI1OSwiZXhwIjoxNTIxNzc0MDU5LCJzdWIiOiJBUSIsInNJZCI6IkFRIn0.oPwse_iPKttHOjCdb9Wm-dt_Du-ceLg-7i6W5NS-GgaXF4uK0zTOwEmYefsZfJZZ2iYWmQlc8wU3sWIF0EPCdx71kwKB2-Yz3p0xQlxtM8CKPpSzvV_pRbVBvrhuYS9MRdy6WXGn3VEU4aRU-yX-euAuLjfTpOxwvw_bWBKleffRXtIN3UFSQSrCt0PiQasUwI8W-eSNhvvV5XGngJqIXqwrYLVWoIILTz5c7PQ1f40uHyG_9xJfipYPnIZ76IFYBNGi4mGqNnhQw7qw-BjOXIM0wPPqodFqDxFux5DHRKMPBHC5FVhDRLY-t92DrJz-hYLNARIv1uVbbcuqVxRXx0LQ4bF4ha2LmXTdQc4U1vjTdajhlppOLiq1IN3DpLCPvp2v4jBs-WBD9mABnU9bhWb5RiWi13OwpEovLK7l06GLo5-AnivjcDk2Yd49Xug23vKgnTzvjr9Xj1eqplV1tNMx_r7SL_H7mo33qcjgpdudFWukmXrtyPdQffgkkrwRr52JObshqN7KBci_V9wXIBj4QZc-_EuOANnXsWjjXEBslCoArBBoMSKlQLCeedrXQ0Y-KSPEK68mWH31wvzsdo1d9T4ysndYGgpXmG54-44TuW6jL_dbAnY-iFmWcX3QYxnD_fKf5vXa50CcG4ExT5R0_sG-W-V8ZRr_WAYmndI
Cache-Control: no-cache
Postman-Token: f023ed89-ed24-4030-9d10-6a91a293a8be

[...]

PR #3022, opened 2020-07-09
Use index to search for AccessTokens

lorthirk:
This PR introduces an index in the `ATHT_ACCESS_TOKEN` table that is then used when looking for a specific Access Token, specifically during REST API authentication check and token refresh

**Related Issue**
This should really help #1537 but it's not 100% sure that this is the only cause of that issue. We may think to close it after we gather some feedback from the community

Changing the query we do when looking for a specific AccessToken, filtering out already expired and invalidated tokens, will allow the DBMS to use the newly created index to perform the search, cutting down the search time.

**Screenshots**
This from a brief test on my local machine:

Without the index and the optimized query:

```
kapuadb> SELECT * FROM atht_access_token WHERE token_id = '...'
[2020-07-09 11:33:05] 1 row retrieved starting from 1 in 6 s 714 ms (execution: 6 s 676 ms, fetching: 38 ms)
```

With the index and the optimized query:

```
kapuadb> SELECT * FROM atht_access_token WHERE invalidated_on IS NULL AND expires_on > NOW() AND token_id = '...'
[2020-07-09 11:32:29] 1 row retrieved starting from 1 in 60 ms (execution: 13 ms, fetching: 47 ms)
```

### DIFF
--- a/service/security/authentication/api/src/main/java/org/eclipse/kapua/service/authentication/token/AccessTokenAttributes.java
+++ b/service/security/authentication/api/src/main/java/org/eclipse/kapua/service/authentication/token/AccessTokenAttributes.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2016 Eurotech and/or its affiliates and others
+ * Copyright (c) 2011, 2020 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
@@ -22,5 +22,7 @@ public class AccessTokenAttributes extends KapuaUpdatableEntityAttributes {
 
     public static final String TOKEN_ID = "tokenId";
     public static final String USER_ID = "userId";
+    public static final String EXPIRES_ON = "expiresOn";
+    public static final String INVALIDATED_ON = "invalidatedOn";
 
 }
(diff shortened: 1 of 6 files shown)


service/security/authentication/api/src/main/java/org/eclipse/kapua/service/authentication/token/AccessTokenAttributes.java
```diff
@@ -22,5 +22,7 @@
 
     public static final String TOKEN_ID = "tokenId";
     public static final String USER_ID = "userId";
+    public static final String EXPIRES_ON = "expiresOn";
```
LeoNerdoG, line 25: Please update the header of the file to "2020".

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/AccessTokenAuthenticatingRealm.java
```diff
@@ -11,6 +11,8 @@
  *******************************************************************************/
 package org.eclipse.kapua.service.authentication.shiro.realm;
 
+import java.util.Date;
```
LeoNerdoG, line 14: Please update the header of the file to "2020".

service/security/shiro/src/main/resources/liquibase/changelog-authentication-master.xml
```diff
@@ -19,5 +19,6 @@
     <include relativeToChangelogFile="true" file="./0.3.0/changelog-authentication-0.3.0.xml"/>
     <include relativeToChangelogFile="true" file="./1.0.0/changelog-authentication-1.0.0.xml"/>
     <include relativeToChangelogFile="true" file="./1.2.0/changelog-authentication-1.2.0.xml"/>
+    <include relativeToChangelogFile="true" file="./1.3.0/changelog-authentication-1.3.0.xml"/>
```
LeoNerdoG, line 22: Please update the header of the file to "2020".

LeoNerdoG (changes requested):
Hey @lorthirk,
thare are some headers to be fixed, can you please check and fix them?

Coduz (approved):

Coduz (approved):

PR #3026, opened 2020-07-17
Remove REST Error Pages and /test resource reference

lorthirk:
This PR removes HTML Error Pages for REST APIs, and a reference to a non existing `/test` REST resource in REST API shiro.ini

**Related Issue**
No related issues


rest-api/web/src/main/webapp/WEB-INF/web.xml
```diff
@@ -79,33 +79,4 @@
         </cookie-config>
     </session-config>
 
-    <!-- | | Error Pages | -->
```
LeoNerdoG, line 82: Please update header to 2020.

LeoNerdoG (changes requested):

PR #3046, opened 2020-08-03
Added check on OpenId idToken on the AuthenticationServiceShiroImpl

### LINKED ISSUE #3041: The OpenID idToken should never be 'null'
The OpenID idToken can never be `null`. The `login` method of the `GwtAuthorizationServiceImpl` class already performs a check. However, if the flow does not pass through that method (e.g. in case of tests), the idToken can be `null`. 

Add a check on the `login` method of the `AuthenticationServiceShiroImpl` class, and throw an exception if the `getIdToken()` method returns a `null` value. The ideal portion of code where the check should be added is [here](https://github.com/eclipse/kapua/blob/890e9bdaa7c53b32cecd824876d591757059c9ce/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java#L133).

_N/A_

**Additional context**
The OpenID idToken is returned by the OpenID Provider with the accessToken. We use it in Kapua in order to perform the logout from the OpenID Session on the OpenID Provider side (see [here](https://github.com/eclipse/kapua/blob/890e9bdaa7c53b32cecd824876d591757059c9ce/console/core/src/main/java/org/eclipse/kapua/app/console/core/server/util/OpenIDLogoutListener.java#L50) for further information).

gbarbon:
Added check on OpenId idToken on the `AuthenticationServiceShiroImpl` to be sure that the token is not `null`.

**Related Issue**
This PR fixes/closes _#3041_

A check is added on the login method of the `AuthenticationServiceShiroImpl` class. It throws an exception if the `getIdToken()` method returns a `null` or an empty value. 

**Screenshots**
_N/A_

_N/A_

### DIFF
--- a/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
+++ b/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
@@ -14,6 +14,7 @@ package org.eclipse.kapua.service.authentication.shiro;
 import java.util.Date;
 import java.util.UUID;
 
+import com.google.common.base.Strings;
 import org.eclipse.kapua.KapuaEntityNotFoundException;
 import org.eclipse.kapua.KapuaException;
 import org.eclipse.kapua.KapuaRuntimeException;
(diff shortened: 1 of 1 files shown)


service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
```diff
@@ -131,9 +131,11 @@ public AccessToken login(LoginCredentials loginCredentials) throws KapuaExceptio
         } else if (loginCredentials instanceof ApiKeyCredentialsImpl) {
             shiroAuthenticationToken = new ApiKeyCredentialsImpl(((ApiKeyCredentialsImpl) loginCredentials).getApiKey());
         } else if (loginCredentials instanceof JwtCredentialsImpl) {
-            shiroAuthenticationToken = new JwtCredentialsImpl(((JwtCredentialsImpl) loginCredentials).getJwt(),
-                    ((JwtCredentialsImpl) loginCredentials).getIdToken());
             openIDidToken = ((JwtCredentialsImpl) loginCredentials).getIdToken();
+            if (openIDidToken == null || openIDidToken.isEmpty()) {
```
Coduz: When checking `String`s emptyness please use `Strings.isNullOrEmpty(...)` :)

Coduz (changes requested):

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
gbarbon: Done!

Coduz (approved):

PR #3047, opened 2020-08-03
Merged SSO 'audience' and 'client_id' properties into a single one

gbarbon:
The `sso.openid.client.id` and the `sso.generic.openid.jwt.audience.allowed` are merged into only one, since they basically represent the same value. 

**Related Issue**
This PR fixes/closes _#3038_

`sso.openid.client.id` is the chosen name for the merged property, while `sso.generic.openid.jwt.audience.allowed` has been removed.

**Screenshots**
_N/A_

The `sso.md` documentation file has also been updated accordingly.

### DIFF
--- a/assembly/console/entrypoint/run-console
+++ b/assembly/console/entrypoint/run-console
@@ -42,14 +42,12 @@ if [ -n "$KAPUA_CONSOLE_URL" ] && [ -n "$OPENID_JWT_ISSUER" ] && [ -n "$OPENID_A
    echo "  Token Endpoint: $OPENID_TOKEN_ENDPOINT"
 
    : OPENID_CLIENT_ID=${OPENID_CLIENT_ID:=console}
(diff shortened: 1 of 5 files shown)


gbarbon:
Temporarily converted to a draft in order to perform further investigation on the two properties.

gbarbon:
`sso.openid.client.id` is used as parameter in the requests to the OpenID Provider, while `sso.generic.openid.jwt.audience.allowed` is used by the `JwtProcessor`in order to validate the token received from the OpenID Provider. These two should correspond to the same `clientId` (as described in the corresponding issue). 

However, I recently discovered an OpenID Provider implementation that was using different values between clientId and audience, thus the only way to make it work was to use two different values for these properties. Since the generic-provider should be more 'permissive', I prefer maintaining two distinct properties (I will however improve the related documentation). I'm thus closing this PR.

PR #3052, opened 2020-08-14
Correctly populate Child Accounts Menu

### LINKED ISSUE #3049: Account list not properly shown
**Describe the bug**
Sometimes (approximately half of the times), when the list of available accounts is shown, it appears with only one row with the kapua-sys account:

imagen

In these cases, also sometimes the list is just there but only with a one row height view, allowing to choose the account but in a very uncomfortable way.

**To Reproduce**
It is not plenty reproducible as it happens arbitrarily. However, when it happens the steps would be just click on the user and in the Switch to account menu.

**Expected behavior**
The list should appear properly loaded on time, or wait for the data to be properly shown. Another option would be to show a _Loading..._ row while the data is prepared.

**Version of Kapua**
1.2.4

**Type of deployment**
- [ ] Local Vagrant deployment
- [x] Docker
- [ ] Openshift (in its variants) 
- [ ] Others

**Main component affected**
- [x] Console -> Mozilla Firefox
- [ ] REST API
- [ ] Message Broker
- [ ] - Others

lorthirk:
This PR adds a "Loading Child Accounts“ fake menu item in the Menu when loading child accounts, and forces a refresh after the list is loaded.

**Related Issue**
Fixes #3049 

The proposed solution isn't 100% clean, since in theory the menu should refresh itself when adding new items but for some reasons it doesn't. Through tests and reviews (@pintify feel free to test and report) are more than welcome.

### DIFF
--- a/console/core/src/main/java/org/eclipse/kapua/app/console/core/client/NorthView.java
+++ b/console/core/src/main/java/org/eclipse/kapua/app/console/core/client/NorthView.java
@@ -11,26 +11,8 @@
  *******************************************************************************/
 package org.eclipse.kapua.app.console.core.client;
 
-import com.extjs.gxt.ui.client.Style;
-import com.extjs.gxt.ui.client.event.BaseEvent;
-import com.extjs.gxt.ui.client.event.Events;
-import com.extjs.gxt.ui.client.event.Listener;
-import com.extjs.gxt.ui.client.event.MenuEvent;
-import com.extjs.gxt.ui.client.event.SelectionListener;
-import com.extjs.gxt.ui.client.widget.ContentPanel;
-import com.extjs.gxt.ui.client.widget.LayoutContainer;
-import com.extjs.gxt.ui.client.widget.button.Button;
-import com.extjs.gxt.ui.client.widget.layout.TableData;
-import com.extjs.gxt.ui.client.widget.layout.TableLayout;
-import com.extjs.gxt.ui.client.widget.menu.Menu;
-import com.extjs.gxt.ui.client.widget.menu.MenuItem;
-import com.extjs.gxt.ui.client.widget.menu.SeparatorMenuItem;
-import com.google.gwt.core.client.GWT;
-import com.google.gwt.user.client.Element;
-import com.google.gwt.user.client.Window;
-import com.google.gwt.user.client.rpc.AsyncCallback;
-import com.google.gwt.user.client.ui.SimplePanel;
-import com.google.gwt.user.client.ui.Widget;
+import java.util.List;
+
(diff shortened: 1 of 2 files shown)


LeoNerdoG:
Hey @lorthirk, I have checked this PR and found something interesting.
I have created almost 1,5k accounts and the loading sometimes takes up to a minute if user wants to change accounts (delay is expected but one minute is quite high).
Nontheless, I saw, that the account list is not properly shown sometimes (see screenshot bellow). This happens if userwants to switch from kapua-sys to some other account or other way around. 
To make things worse, this doesn't happen always... I think this is not a blocker, but I am just posting it here so we don't forget about it and if somebody else has similar problems, to be aware of this. Maybe we could post this under "known issues" until we resolved it for sure. 

**Screenhsot No.1:** 
Screenshot 2020-08-25 at 10 50 35

**Screenshot No. 2:** 
Screenshot 2020-08-25 at 10 50 59

lorthirk:
Yeah... I suspected that some corner case could come around. Probably there's something obscure happening in the GWT/GXT internals I can't identify. I honestly don't know if we want to merge this or not... It looks like an improvement, although it's not perfect. What do you guys think?

Coduz (approved):

LeoNerdoG:
Hey @lorthirk,
honestly I would ask @pintify, because he opened the issue - What do you think, is this an improvement, should we merge it?

pintify:
Sorry, I don't quite get the problem in the screenshots you pass, the names shouldn't get all the "OOOO"s? Or is it that the width of the list is not correct?

In any case, what was the behaviour prior to this PR? If it didn't showed anything (like the case I present in the issue, then this is an improvement although not perfect, isn't it?

LeoNerdoG:
hey @pintify,
my test was to create 1500 accounts named "OOOOOOOOOOOOOOOOOOOOOOO" to see if they are shown correctly or not and it it takes a normal amount of time to show them. Sometimes, the names were not properly shown, but as you said yourself - this PR fixes the original issue, so from this standpoint, this can be merged.

pintify:
Then it is only the width of the list what is not properly set? I guess it is also part of the initial issue, although the dimensions that worried me were only in vertical.

LeoNerdoG:
Yes, the problem is sometimes in horizontal list presentation - it is not wide enough. However, this does not happen often, but there are some corner cases when it does. You can try it yourself if you have the resources.

pintify:
Unfortunately I don't have enough resources at hand to launch a stable Kapua. The only place I could launch it is in the servers we use but only closed versions.

Regarding the PR, the improve is good enough for me. However if due to quality standards or similar culture you require the solution of the width issue, I could understand it and wait for a better approach.

PR #3060, opened 2020-08-25
Add startDate and endDate parameters for Device Events REST

### LINKED ISSUE #3057: Retrieve device events with time interval filter
Normal paging with limit and offset is not suitable for device events browsing.

The endpoint to retrieve the device events should provide parameters to filter the events according to a time interval.
Something like startDate and endDate should be ok (same params used in the dataMessage endpoint) 

None.

**Additional context**
Involved endpoint 

/{scopeId}/devices/{deviceId}/events

lorthirk:
Add `startDate` and `endDate` when querying for Device Events via REST

**Related Issue**
Fixes #3057

### DIFF
--- a/rest-api/resources/src/main/java/org/eclipse/kapua/app/api/resources/v1/resources/DeviceEvents.java
+++ b/rest-api/resources/src/main/java/org/eclipse/kapua/app/api/resources/v1/resources/DeviceEvents.java
@@ -15,11 +15,13 @@ import com.google.common.base.Strings;
 import org.eclipse.kapua.KapuaEntityNotFoundException;
 import org.eclipse.kapua.KapuaException;
 import org.eclipse.kapua.app.api.resources.v1.resources.model.CountResult;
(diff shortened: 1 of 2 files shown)


rest-api/resources/src/main/java/org/eclipse/kapua/app/api/resources/v1/resources/DeviceEvents.java
```diff
@@ -15,11 +15,13 @@
 import org.eclipse.kapua.KapuaEntityNotFoundException;
 import org.eclipse.kapua.KapuaException;
 import org.eclipse.kapua.app.api.resources.v1.resources.model.CountResult;
+import org.eclipse.kapua.app.api.resources.v1.resources.model.DateParam;
```
LeoNerdoG, line 18: Please update header to "2020".

Coduz (approved):

PR #3107, opened 2020-10-14
Added Junit tests for broker module

sonja-ct:
Added Junit test for classes in:

broker/api module:
BrokerDomain class
BrokerDomains class

broker/core/message package:
CamelKapuaMessage class
CamelUtil class
MessageConstants class
DefaultSystemMessageCreator class

broker/core/plugin/metric package:
ClientMetric class
LoginMetric class
PublishMetric class
SecurityMetrics class
SubscribeMetric class

Signed-off-by: Sonja <sonja.matic@endava.com>

**Related Issue**
This PR has no related issue, but closes PR #3075.

**Description of the solution adopted**
/
**Screenshots**
/
**Any side note on the changes made**
/

PR #3109, opened 2020-10-15
Added Junit tests for pool and setting package

sonja-ct:
Added Junit tests for classes in broker/core module:

pool package:
JmsAssistantProducerPool class
JmsAssistantProducerWrapper class
JmsAssistantProducerWrapperFactory class
JmsConnectionFactory class
JmsProducerWrapper class
JmsConsumerWrapper class

setting package:
BrokerSetting class
BrokerSettingKeyTest class

Signed-off-by: Sonja <sonja.matic@endava.com>

**Related Issue**
/

/

**Screenshots**
/

Tests for create () method in JmsAssistantProducerWrapperFactory class were not written because it took too long to create the connection. This also applies to tests in JmsConsumerWrapper class.

### DIFF
--- a/broker/core/src/test/java/org/eclipse/kapua/broker/core/pool/JmsAssistantProducerPoolTest.java
+++ b/broker/core/src/test/java/org/eclipse/kapua/broker/core/pool/JmsAssistantProducerPoolTest.java
@@ -1,0 +1,61 @@
+/*******************************************************************************
+ * Copyright (c) 2021 Eurotech and/or its affiliates and others
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Eurotech - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.kapua.broker.core.pool;
+
+import org.eclipse.kapua.qa.markers.junit.JUnitTests;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(JUnitTests.class)
+public class JmsAssistantProducerPoolTest extends Assert {
+
+    @Test
+    public void jmsAssistantProducerPoolTest() {
+        String[] destinations = {"test", "destination1234", "","destination test",
+                "destinationdestinationdestinationdestinationdestinationdestinationdestinationdestinationdestinationdestinationdestinationdestinationdestinationdestinationdestinationdestinationdestinationdestinationdestinationdestinationdestinationdestinationdestinationd",
+                "asd#$$%%/(@", "123", "-23", null};
+        for (String destination : destinations) {
+            JmsAssistantProducerPool pool = new JmsAssistantProducerPool(new JmsAssistantProducerWrapperFactory(destination));
+            assertEquals("Expected and actual values should be the same.", 10, pool.getMaxTotal());
+            assertEquals("Expected and actual values should be the same.", 10, pool.getMaxIdle());
+            assertEquals("Expected and actual values should be the same.", 5, pool.getMinIdle());
+        }
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void jmsAssistantProducerPoolNullTest() {
+       new JmsAssistantProducerPool(null);
+    }
+
+    @Test
+    public void getIOnstanceTest() {
+        JmsAssistantProducerPool pool = JmsAssistantProducerPool.getIOnstance(JmsAssistantProducerPool.DESTINATIONS.NO_DESTINATION);
+        assertEquals("Expected and actual values should be the same.", "org.eclipse.kapua.broker.core.pool.JmsAssistantProducerWrapperFactory<org.eclipse.kapua.broker.core.pool.JmsAssistantProducerWrapper>", pool.getFactoryType());
+    }
+
+    @Test
+    public void getIOnstanceNullTest() {
+        JmsAssistantProducerPool pool = JmsAssistantProducerPool.getIOnstance(null);
+        assertNull("Null expected.", pool);
+    }
+
+    @Test
+    public void closePools() {
+        try {
+            JmsAssistantProducerPool.closePools();
+        } catch (Exception e) {
+            fail("Exception should not be thrown");
+        }
+    }
+}
--- a/broker/core/src/test/java/org/eclipse/kapua/broker/core/pool/JmsAssistantProducerWrapperFactoryTest.java
+++ b/broker/core/src/test/java/org/eclipse/kapua/broker/core/pool/JmsAssistantProducerWrapperFactoryTest.java
@@ -1,0 +1,149 @@
+/*******************************************************************************
+ * Copyright (c) 2021 Eurotech and/or its affiliates and others
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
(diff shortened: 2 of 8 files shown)


LeoNerdoG (approved):
Hey @Coduz,
I have checked this PR and it looks OK - can you please recheck and merge it when you have time? Thanks!

riccardomodanese:
Hi @LeoNerdoG I was doing a quick look to this pr and a couple others (like 3107 and 3104).
From my point of view most of these unit tests are unuseful.
Testing objects like JMSProducerWrapper and the MetricsHelper(s) that are components used internally in this module and are not meant to be exposed outside I think has no utilty.
May be better focus into integration test to prove these components are working fine. And more or less we already have tests to cover these components also.

I have a question. Why is not used Cucumber?

LeoNerdoG:
Hey @riccardomodanese,
the answer to both parts of the question is basically the same - achieving code coverage. 
Firstly, we wanted to write unit tests to cover classes or rather methods in these classes, becasue most of the methods did not have even basic unit tests to check their functionality. Nonetheless, it is very hard to create good tests, becasue we don't have the specifications from which we could see what is the correct functionality so we create conclusions based on the written code - this is not the right way to go, but lack of documentation leaves us no other choice. 
We are working closely with @Coduz if we need additional info and with whom we have created a test backlog document in which everything is documented (and updated).
The second part of our work is (or rather will be) writing integration tests to see, if the components are working well with each other. We aren't even close to this stage yet, becasue we are don't have enough people on the project to move forward as we would want to. But I am sure you can understand this (we still have a long way to go with testing - REST API, frontend, performance...).

The answer why aren't we using cucumber is the same as above - we had a lot of problems with achieving code coverage with cucumber (as achieving code coverage is our primary goal here as discussed with @Coduz). We did not investigate why (again due to lack of people and time), so we used plain JUnit instead (if you ask me it is several times faster, cleaner, easier to write/implement and to run).
We can transform these test to cucumber with minimal effort, becasue all the implementation for tests is there, we would just have to write .feature files, but I cannot guarantee that coverage would stay high, it might even drop significantly. 

`May be better focus into integration test to prove these components are working fine. `
In every TDD or BDD or any kind of development you simply have to have unit tests for even smallest chunks of code. There is no other way around it to see, if the implemented methods work properly. Simply writing integration tests does not cover all the necessary cases, as this is only one part of the testing effort (unit -> integration -> system -> functional -> performance -> frontend -> ...). If you skip even one of them, the product will not be as stable as it could be.

`And more or less we already have tests to cover these components also.`
There may be some (integration) tests for these components, but as stated above, simply writing a broad integration test is not enough and will not give the right picture about the stability and reliability of the methods, classes and project as a whole.

riccardomodanese:
I'll give you some examples about what I mean.
The JmsProducerWrapper is a wrapper over a JMS connection/session. So doing a unit test ([like this](https://github.com/eclipse/kapua/blob/84919924e23ccbe3c4b9837ef6c2ca04fae4a34a/broker/core/src/test/java/org/eclipse/kapua/broker/core/pool/JmsAssistantProducerWrapperTest.java)) for this class means more or less testing the contructor of the class. Ok you can add this test but from my point of view in unuseful since the real issues using this object are coming when real connections/sessions are attached/used.
This [test](https://github.com/eclipse/kapua/blob/84919924e23ccbe3c4b9837ef6c2ca04fae4a34a/broker/core/src/test/java/org/eclipse/kapua/broker/core/pool/JmsAssistantProducerWrapperTest.java#L128) for example, is testing more or less this [line](https://github.com/eclipse/kapua/blob/84919924e23ccbe3c4b9837ef6c2ca04fae4a34a/broker/core/src/main/java/org/eclipse/kapua/broker/core/pool/JmsAssistantProducerWrapper.java#L55) so, with a mocked producer I will not expect any exception at all.
Moreover, I'm quite sure If you use these [topics](https://github.com/eclipse/kapua/blob/84919924e23ccbe3c4b9837ef6c2ca04fae4a34a/broker/core/src/test/java/org/eclipse/kapua/broker/core/pool/JmsAssistantProducerWrapperTest.java#L120) with a real instance, the test will fail because MQTT wildcards are not allowed in the topic when publishing a message.
Anyway the connection pool is used in a lot of qa since it's used by the broker in each device login.

About MetricHelpers there are a lot of unuseful test. For example simple I don't understand why there are tests [like these](https://github.com/eclipse/kapua/pull/3107/files#diff-baa28f32357e29f169263eb7f8df2e3c0aaa471cb34f648474f3b70c5ddba314R47).
A method with a signature like
SomeObject getSomething()
obviously will return an instance of SomeObject or null instance or throws exception.
So, if the method will throw exception, the test will not track correctly the failure. Otherwise is more or less like doing AssertTrue(true) assuming that the method will not return null.

I see a lot of accessing to private filelds. If the code is too difficult to test we can discuss if it's possible to change the code to make it more clear and easy to test, otherwise private methods/fields should be tested through public methods from my point of view.
In other words, sometimes I agree an access to a private field/method can be the only way to properly test a class/feature, but it shouldn't be abused.

LeoNerdoG:
Hey @riccardomodanese,
thanks for the detailed analysis. I understood exactly what you meant from the original post, but nonetheless I have to point out couple of things. 

> The JmsProducerWrapper is a wrapper over a JMS connection/session. So doing a unit test (like this) for this class means more or less testing the contructor of the class

I agree that testing a constructor of a class might be useless, but for achieving the primary goal (which is coverage) we are trying to test every line, every method, every constructor. I can point out hundreds of tests that might be useless (at first sight) but are actually not (for example - why would we test enums? It's just stupid if you think briefly, but if you really get into it - you write tests to check certain functionality or behaviour, maybe if soemthing is missing or not... Tests can point out that something is wrong when changing one part and maybe forgetting on something else). 

> This test for example, is testing more or less this line so, with a mocked producer I will not expect any exception at all.
Moreover, I'm quite sure If you use these topics with a real instance, the test will fail because MQTT wildcards are not allowed in the topic when publishing a message.
Anyway the connection pool is used in a lot of qa since it's used by the broker in each device login.

As already said in the answer above, I understand that some tests might be useless, but we can't just skip entire unit testing. Testing methods also gives us a very good insight into the product itself, so when we will do integration tests, they will be better. Some things can't be properly unit tested, so we either skipped them or tried to test them to best of our ability, understanding and knowledge. It is absolutely clear that testing if something returns an obvious instance of something is stupid, but the architeture or methods or soemthing else could change in the future and these tests will maybe give us some clues on what is wrong etc. We could skip this, but then the coverage would fall and we would not have the knowledge and experience that we posses now. 

Regarding private fields/methods: I am not a fan of accessing private fields and/or methods, but some private methods are so complex, that we just have to make them accessible. There is a difference in private methods if one just returns an object of something and some other that has 100 lines of logic and it's private (and the method that calls this private method is not that complex). 

One more thing about understanding the code - if you ask me, we would have to have a long session about all the stuff we might not understand (I am talking days here), create a roadmap for testing and (if possible) development, select technologies and our goals (which should never be code coverage if you ask me) and then start testing. Currently we are just trying our best to achieve only goal we have with people and time that are available (we currently have only 2 engineers working on Kapua - me and Sonja) - we can't even test the PRs that are being created from your side, that's how tight we are on time and people. 

One last tought from my side is that you have to make a decision what you actually want from us, becasue we can change our focus from unit to integration tests or something else easily. Just let us know how you decided and we can proceed accordingly.

LeoNerdoG (approved):
Hey @Coduz,
I have checked this branch after all the latest changes and it is ready to be merged. 
Can you please take a look when you can and merge it, so we "clean" our PRs a bit? 😄 
Thanks!

LeoNerdoG (approved):

Coduz (approved):

PR #3121, opened 2020-10-27
Configurable Minimum Password Length

lorthirk:
Users Passwords can have their minimum length configured per account. A new configuration has been created in `CredentialService`, named `password.minLength`. Such value can be specified for any account or left blank; in this case, the value will be read from a default value, again configurable with the `authentication.credential.userpass.password.minlength` system property, with the following restrictions:

- The System Default value cannot be less than 12 (hardcoded)
- The custom value for any account cannot be less than the System Default
- A maximum length of 255 is now enforced at password creation

Changing the value at any time won't affect existing passwords. The value is only checked when creating a new password.

From now on, the Password RegEx does not contain the limit of 12 characters anymore, and will only check the presence of the character classes (uppercase letters, lowercase letters, numbers and symbols). The length will be checked by the business logic of the service itself.


riccardomodanese (approved):

LeoNerdoG:
@Coduz @lorthirk please do not merge this just yet, as testing is not completed - I am testing this but was not able to finish today. I will notify you when everything is finished.

PR #3122, opened 2020-10-28
Added Integration Tests For SSO Using Keycloak Provider

sonja-ct:
Tests are added for SSO login using Keycloak provider.

Signed-off-by: Sonja sonja.matic@endava.com

In this PR are added integration tests for SSO using Keycloak provider.

**Related Issue**
/

There are a few changes in this PR:

- New testing methods for Docker are added to configure keycloak containers,
- Keycloak configuration is added to kapua-realm.json file,
- Added test scenarios and methods that are testing SSO login using Keycloak provider,
- Modified tests for creating user.

**Screenshots**
/

The author of all these tests is @zanpizmoht, and they are in #2978 . I made a new PR because it was necessary to do a rebase with develop.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -935,6 +935,11 @@
             </dependency>
             <dependency>
                 <groupId>org.eclipse.kapua</groupId>
+                <artifactId>kapua-sso-test-steps</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.eclipse.kapua</groupId>
                 <artifactId>kapua-sso-provider</artifactId>
                 <version>${project.version}</version>
             </dependency>
--- a/qa/common/src/main/java/org/eclipse/kapua/qa/common/cucumber/CucUser.java
+++ b/qa/common/src/main/java/org/eclipse/kapua/qa/common/cucumber/CucUser.java
@@ -45,6 +45,8 @@ public class CucUser {
 
     private String expirationDate;
 
+    private String externalId;
+
     public String getName() {
         return name;
     }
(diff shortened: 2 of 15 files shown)


gbarbon:
Hi @sonja-ct ,

We (me and @coduz ) took a look at the PR, and we think that the PR could be improved in such a way to considerably reduce the number of lines, and possibly avoid the CQ too, by getting rid of the big `kapua-realm.json` file. In Kapua you can find a way to deploy Keycloak directly on docker in an automatic way (see  [deployment/commons/sso/keycloak](https://github.com/eclipse/kapua/tree/develop/deployment/commons/sso/keycloak)). Thus, instead of using a json file to configure Keycloak, the Keycloak docker image is built using a Dockerfile and it incudes a script that configures Keycloak at the first execution.

FYI, this is how the things work. The `build` shell script is responsible for building the new Keycloak docker image by calling the Dockerfile that is in the same directory. The Dockerfile downloads the chosen Keycloak image and adds to this image the `run-keycloak` shell script. This `run-keycloak` script is launched when a Keycloak docker container, based on the new Keycloak image, is created and started. It configures the realm and the client inside Keycloak, and also creates a test user at the end.

In order to integrate this with Cucumber, the idea is to call the `build` script before the tests. Then, hook the modified Keycloak image at line 465 of the DockerSteps.java file ([here](https://github.com/eclipse/kapua/pull/3122/files#diff-02c594fa19f7bea539ec2682c69d09f58365157c2f892c13bf6b02088f5c4e78R465)), where there is this `.image("jboss/keycloak:8.0.1")`. The new image name and tag to be used are `keycloak:sso` (they are assigned by the `build` script).

Now, the only thing that remains to do is retrieving the ID of the user created in Keycloak. To do this, you can use the Keycloak REST APIs. Fist, you will need to authenticate to Keycloak in order to obtain an `access_token`. To do this you can use the OpenID REST API for authentication (I tested this locally by performing a POST to `http://localhost:9090/auth/realms/master/protocol/openid-connect/token`, note that my local instance of Keycloak is `http://localhost:9090`). Then, you ony need to perform a `GET \users` to retrieve the data of the users (I tested by performing it to `http://localhost:9090/auth/admin/realms/kapua/users/`) by passing the `access_token` in the header. Then you will finally obtain the user `id` in the response of this GET.

PR #3152, opened 2020-11-24
Extend accountSelfEdit support in Configurations

lorthirk:
Extending support for `allowSelfEdit` in Service configurations. A field marked with `allowSelfEdit="true"` in the Service Metadata XML can be edited by users of the account itself, not only from the parent account. As a first use case, all the configurations of `CredentialsService` are now self-editable.

An user can edit its own `allowSelfEdit="true"` configurations in the "Settings" view of the console, or via REST API.

**Related Issue**
No related issues

### DIFF
--- a/commons/src/main/java/org/eclipse/kapua/commons/configuration/AbstractKapuaConfigurableService.java
+++ b/commons/src/main/java/org/eclipse/kapua/commons/configuration/AbstractKapuaConfigurableService.java
@@ -15,6 +15,7 @@ package org.eclipse.kapua.commons.configuration;
 
 import javax.validation.constraints.NotNull;
 import javax.xml.bind.JAXBException;
+import javax.xml.namespace.QName;
 import javax.xml.stream.XMLStreamException;
 
 import org.eclipse.kapua.KapuaEntityNotFoundException;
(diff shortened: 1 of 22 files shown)


commons/src/main/java/org/eclipse/kapua/commons/configuration/AbstractKapuaConfigurableService.java
```diff
@@ -391,18 +393,35 @@ public void setConfigValues(KapuaId scopeId, KapuaId parentId, Map<String, Objec
...
+                    // current configuration does not allow self edit...
+                    !allowSelfEdit &&
+                    // a configuration for the current logged account is about to be changed...
+                    KapuaSecurityUtils.getSession().getScopeId().equals(scopeId) &&
+                    // and the new value is different from the other one...
+                    !originalValues.get(ad.getId()).equals(values.get(ad.getId()));
+
+            if (preventChange) {
+                // ... prevent the change!
+                throw KapuaException.internalError(String.format("The configuration \"%s\" cannot be changed from an user of the account", ad.getId()));
```
LeoNerdoG: I think that this message should be "The configuration \"%s\" cannot be changed by this user in this account" or something similar. What do you think?

LeoNerdoG (changes requested):
Hey @lorthirk,
I have checked this PR and it all looks fine except the message as commented bellow and this permission error. I think that it should be handled more gracefuly? What do you think?

Screenshot 2020-11-25 at 07 46 13

commons/src/main/java/org/eclipse/kapua/commons/configuration/AbstractKapuaConfigurableService.java
lorthirk: Yes, this could work

lorthirk:
Well... In a way I agree with you, the error should be handled more gracefully. Anyway this is a general problem of how missing permissions are reported: you will get the same message if, let's say, you go change a setting in a child account for a service you only have a `read` permission. So I wouldn't link the issue to this PR, but rather open a new one for the specific case.

Coduz (approved):

LeoNerdoG (approved):

PR #3166, opened 2020-12-02
MFA authentication and CRUD operations via REST API

### LINKED ISSUE #3144: Multi Factor Authentication can be enabled also for an external SSO user
**Describe the bug**
The Multi Factor Authentication can be enabled also for an external Single Sign On (OpenID Connect) user. This does not make any sense, since the user credentials (and possibly an MFA) are handled by the external OpenID Connect provider.

**To Reproduce**
Enable SSO, then create an external user. Then, login with the external user and access the MFA panel.

**Expected behavior**
It should not be possible to enable MFA for an external user.

**Screenshots**
_n/a_

**Version of Kapua**
1.4.0-SNAPSHOT

**Type of deployment**
[x] Local Vagrant deployment
[x] Docker
[x] Others

**Main component affected**
[ ] REST API
[ ] Message Broker
[ ] - Others

**Additional context**
_n/a_

### LINKED ISSUE #3168: MFA: kapua-sys user can enable the MfaOption also for other users
**Describe the bug**
The `kapua-sys` user (thus a user with `admin` role) can enable the MfaOption also for other users, while this should not be possible (only the 'disable' operation should be possible).

**To Reproduce**
1. Login in Kapua
2. Create a new user
3. Select the new user
4. Enable Mfa for the selected user

**Expected behavior**
The `kapua-sys` user, and any other 'admin' user, should not be able to enable the MfaOption for other users. 

**Screenshots**
_n/a_

**Version of Kapua**
[1.4.0-SNAPSHOT]

**Type of deployment**
[x] Local Vagrant deployment
[x] Docker
[ ] Others

**Main component affected**
[ ] REST API
[ ] Message Broker
[ ] - Others

**Additional context**
The solution can probably be applied in the `if` statement of the `updateUIComponents` method in the `MfaManagementPanel` class. However, this solution only applies to the Web UI, while Rest API would remain affected by the issue.

gbarbon:
This PR introduces authentication and CRUD operations for Multi Factor Authentication via REST API. 

Moreover, this PR also fixes open issues concerning MFA : 
- the MFA should not be enabled for other users (thus, should only be self enabled);
- the MFA creation should be disabled for SSO user;

and improves the security of the trust key used for the MFA.

**Related Issue**
Fix #3168 , fix #3144 

The following MFA endpoints are provided for CRUD operations:
- `GET /{scopeId}/users/{userId}/mfa` to get the `MfaOption` entity of the given user;
- `POST /{scopeId}/users/{userId}/mfa` to enable the MFA for the provided user;
- `DEL /{scopeId}/users/{userId}/mfa` to disable the MFA for the provided user;
- `DEL /{scopeId}/users/{userId}/mfa/disableTrust` to disable the machine trust of the `MfaOption` for the provided user.

The MFA 'create' will return an `MfaOption` object which also contain the QR code image, the MFA secret in plaintext and the list of scratch codes (in plaintext too). Note that the MFA secret and the scratch codes are stored in the DB in an encrypted format, and the creation is the only moment in which they are visible to the user. An 'update' endpoint is not provided, since an `MfaOption` cannot be modified (but only deleted).

Moreover, a new authentication endpoint is provided:
- `POST /authentication/mfa` in order to authenticate through username, password and MFA authentication code (or machine trust key).

This endpoint also allows to enable the machine trust, it the `enableTrust=true` query parameter is provided.

The normal authentication endpoint (username + pwd) has been improved in such a way to return the 'REQUIRE_MFA_CREDENTIALS' Kapua error code in case the user has the MFA enabled. Note also that the machine trust key can be obtained in plain text only through the access token when the MFA login Rest API is performed with `enableTrust=true`. The  GET of  an `MfaOption` will always return encrypted values, for both the `MfaOption` secret and the trust key (if set).

The `GwtScratchCode` classes and services have been removed, moreover the whole management of scratch codes in the Web UI has been simplified and the scratch code generation has been incorporated in the `create` method of the `MfaOptionServiceImpl` class. These changes have been required in order to make it possible to generate the `MfaOption` via REST API in a single step.

The trust key has been improved with BCrypt encryption on the DB, and it's returned in plain text only with the access token when the `enableTrust` is triggered, in both the Rest APIs and in the GWT based Web UI. Moreover, to address security concerns, now only a single machine at a time can be trusted (if a new browser/machine is trusted, the old one will lose the trust). Multiple machine trust will be introduced again with a dedicated issue/PR, with an implementation that allows using a unique trust key for each machine.
  
**Screenshots**
_n/a_

_n/a_


rest-api/resources/src/main/java/org/eclipse/kapua/app/api/resources/v1/resources/MfaOptions.java
```diff
@@ -0,0 +1,205 @@
...
+import javax.ws.rs.GET;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+@Path("{scopeId}/mfaoptions")
```
Coduz: @Path("{scopeId}/users/{userId}/mfa")

rest-api/resources/src/main/java/org/eclipse/kapua/app/api/resources/v1/resources/MfaOptions.java
```diff
@@ -0,0 +1,205 @@
...
+     * @param limit   The result set limit.
+     * @return The {@link MfaOptionListResult} of all the MfaOptions associated to the current selected scope.
+     * @throws KapuaException Whenever something bad happens. See specific {@link KapuaService} exceptions.
+     * @since 1.4.0
+     */
+    @GET
+    @Produces({MediaType.APPLICATION_JSON, MediaType.APPLICATION_XML})
+    public MfaOptionListResult simpleQuery(
+            @PathParam("scopeId") ScopeId scopeId,
+            @QueryParam("userId") EntityId userId,
```
Coduz: @PathParam("userId") EntityId userId,

rest-api/resources/src/main/java/org/eclipse/kapua/app/api/resources/v1/resources/MfaOptions.java
```diff
@@ -0,0 +1,205 @@
...
+     * Disable trusted machine for a given {@link MfaOption}.
+     *
+     * @param scopeId     The ScopeId of the requested {@link MfaOption}.
+     * @param mfaOptionId The id of the {@link MfaOption}
+     * @return HTTP 200 if operation has completed successfully.
+     * @throws KapuaException Whenever something bad happens. See specific {@link KapuaService} exceptions.
+     * @since 1.4.0
+     */
+    @POST
+    @Path("{mfaOptionId}/disabletrust")
```
Coduz: @Path("trust")

rest-api/resources/src/main/java/org/eclipse/kapua/app/api/resources/v1/resources/MfaOptions.java
```diff
@@ -0,0 +1,205 @@
...
+    /**
+     * Disable trusted machine for a given {@link MfaOption}.
+     *
+     * @param scopeId     The ScopeId of the requested {@link MfaOption}.
+     * @param mfaOptionId The id of the {@link MfaOption}
+     * @return HTTP 200 if operation has completed successfully.
+     * @throws KapuaException Whenever something bad happens. See specific {@link KapuaService} exceptions.
+     * @since 1.4.0
+     */
+    @POST
```
Coduz: `@DELETE`

Coduz (approved):

Coduz (approved):

PR #3183, opened 2020-12-22
Cache exception handling improvement

gbarbon:
This PR introduces cache exception handling in the `EntityCache` class and its subclasses. 

**Related Issue**
_n/a_

Cache errors are usually handled by the cache client (that implements the JCache interface). However, in some cases the exceptions are thrown by the cache client itself (e.g. a timeout exception). Since we are agnostic w.r.t. the underlying JCache implementation, we cannot catch precise exceptions. Thus, this PR introduces a way to treat such exceptions properly, in a conservative way w.r.t. the data, and to log them via the logger and the metrics.

**Screenshots**
_n/a_

_n/a_

### DIFF
--- a/commons/src/main/java/org/eclipse/kapua/commons/service/internal/cache/EntityCache.java
+++ b/commons/src/main/java/org/eclipse/kapua/commons/service/internal/cache/EntityCache.java
@@ -42,6 +42,7 @@ public class EntityCache {
     protected Counter cacheMiss;
     protected Counter cacheHit;
     protected Counter cacheRemoval;
+    protected Counter cacheError;
 
     /**
      * The constructor initializes the {@link #idCache} and the {@link #listsCache}.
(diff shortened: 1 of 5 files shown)


commons/src/main/java/org/eclipse/kapua/commons/service/internal/cache/EntityCache.java
```diff
@@ -146,7 +176,7 @@ protected KapuaEntity checkResult(KapuaId scopeId, KapuaEntity entity) {
      * This mimics the checks that are performed in the 'find' method of the {@link org.eclipse.kapua.commons.service.internal.ServiceDAO} class.
      *
      * @param scopeId a {@link KapuaId} representing the scopeId
-     * @param entity the {@link KapuaListResult} entity to be checked
+     * @param entity  the {@link KapuaListResult} entity to be checked
```
riccardomodanese: please remove the space :)

commons/src/main/java/org/eclipse/kapua/commons/service/internal/cache/EntityCache.java
```diff
@@ -122,7 +152,7 @@ public KapuaListResult removeList(KapuaId scopeId, Serializable id) {
      * This mimics the checks that are performed in the 'find' method of the {@link org.eclipse.kapua.commons.service.internal.ServiceDAO} class.
      *
      * @param scopeId a {@link KapuaId} representing the scopeId
-     * @param entity the {@link KapuaEntity} to be checked
+     * @param entity  the {@link KapuaEntity} to be checked
```
riccardomodanese: please remove the space :)

riccardomodanese (approved):
the pr looks good to me.
If you can just remove the 2 spaces in a couple of javadocs will be good

riccardomodanese (approved):
the pr looks good to me.
If you can just remove the 2 spaces in a couple of javadocs will be good

commons/src/main/java/org/eclipse/kapua/commons/service/internal/cache/EntityCache.java
gbarbon: done!

commons/src/main/java/org/eclipse/kapua/commons/service/internal/cache/EntityCache.java
gbarbon: done!

PR #3197, opened 2021-01-18
Fixed Console Http Session Fixation security issue

Coduz:
This PR fixes the Http Session fixation security issue.

**Related Issue**
_None_

Added invalidation of the HTTP session on each login, before login method is invoked.

**Screenshots**
_None_

Improved `GwtAuthorizationServiceImpl` and `KapuaExceptionHandler` code

### DIFF
--- a/console/core/src/main/java/org/eclipse/kapua/app/console/core/server/GwtAuthorizationServiceImpl.java
+++ b/console/core/src/main/java/org/eclipse/kapua/app/console/core/server/GwtAuthorizationServiceImpl.java
@@ -72,6 +72,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpSession;
 import java.util.concurrent.Callable;
 
 public class GwtAuthorizationServiceImpl extends KapuaRemoteServiceServlet implements GwtAuthorizationService {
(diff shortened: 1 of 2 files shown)


console/core/src/main/java/org/eclipse/kapua/app/console/core/server/GwtAuthorizationServiceImpl.java
```diff
@@ -72,122 +72,140 @@
...
-
-            // Get the session infos
-            if (gwtJwtIdToken == null || gwtJwtIdToken.getIdToken().isEmpty()) {
-                // in this specific case the gwtJwtIdToken cannot be empty
-                throw new KapuaIllegalNullArgumentException("gwtJwtIdToken");
+            // Check Credentials Values
+            ArgumentValidator.notNull(gwtAccessTokenCredentials, "loginCredentials");
+            ArgumentValidator.notNull(gwtAccessTokenCredentials.getAccessToken(), "loginCredentials.accessToken");
+            ArgumentValidator.notNull(gwtJwtIdToken, "jwtIdToken");
+            ArgumentValidator.notNull(gwtJwtIdToken.getIdToken(), "jwtIdToken.idToken");
```
gbarbon: Hi @Coduz , I think we should use  `ArgumentValidator.notEmptyOrNull` for the 'jwtIdToken.idToken' in order to mimic the behaviour that was on the old `if` statement.

gbarbon (changes requested):
Hi @Coduz , I think there's only a `notEmptyOrNull` to add, the rest seems ok to me!

gbarbon (approved):
Everything seems ok to me now!

PR #3219, opened 2021-02-03
Commented "ExecuteOnDeviceConnec" annotation for skipping tests

LeoNerdoG:
I have commented out the "triggerService", "jobs" and "integration" annotations in ExecuteOnDeviceConnectI9n.feature file, which will cause the tests to skip.
Because some CommandExecuteOnDeviceConnect tests are failing for no apparent reason (they are passing locally when running in intelliJ, terminal and on Travis) only on GHA, I have decided to comment this out, as i was unable to fix them. The tests themselves don't need fixing as they are working just fine (as mentioned before), so there is a problem with GHA setup.
The only thing that changed after merging PR 3213 was the Virtual Machine version, which was bumped from 20201210.0  to 20210123.1 and this is actually the only thing that is not configurable in GHA. 

I have tried the following:
 - Changing cache key and/or changing location of .m2/ repository
 - Changing Ubuntu version (using 16.04 instead 18.04.5)
 - Creating new repository (kapua-test) to see if something else is problem (if caches not cleared correctly)
 - Changing Maven, Java versions on local machine (java 1.8.281, maven 3.6.3 - on local machine worked fine with bothe versions of JAVA (1.8_211 and 1.8_281) and maven (3.5.0 and 3.6.3).
Following results of this work I found out that it has to be the VM version problem, which we cannot revert, as this is "hardcoded" in GHA.
After we find a solution for this, I will uncomment the tests.

@Coduz Please merge this PR so we can create all others and "clean-up" the unit testing part of our effort. 

Signed-off-by: Leonardo Gaube <leonardo.gaube@endava.com>

**Related Issue**
/

**Screenshot No. 1 (tests passing in IntelliJ):**
intelliJ-tests-passing

**Screenshot No. 2 (tests passing in Terminal):**
Screenshot 2021-02-03 at 14 20 27

**Screenshot No. 3 (Tests passing of Travis for PR 3218):**
Screenshot 2021-02-03 at 13 18 04

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ install:
 
 env:
   global:
-    - BUILD_OPTS="" 
+    - BUILD_OPTS=""
     - CONFIG_OVERRIDES="-Dcommons.db.schema=kapuadb -Dcommons.settings.hotswap=true -Dbroker.host=localhost"
     - MAVEN_OPTS="-Xmx4096m"
 
--- a/qa/integration/src/test/resources/features/jobScheduling/ExecuteOnDeviceConnectI9n.feature
+++ b/qa/integration/src/test/resources/features/jobScheduling/ExecuteOnDeviceConnectI9n.feature
@@ -10,9 +10,9 @@
 # Contributors:
 #     Eurotech - initial API and implementation
 ###############################################################################
-@jobs
-@triggerService
-@integration
+#@jobs
+#@triggerService
+#@integration
 
 Feature: JobEngineService execute job on device connect
 


Coduz:
Hi @LeoNerdoG ,

I think that we can close this PR without merging it, since I don't think that commenting failing test is the way to go.
I think we had some test that were commented at a certain point and then went forgotten.

It that ok for you?

LeoNerdoG:
Hey,
yeah, it's ok for me. Closing.

PR #3231, opened 2021-02-11
Upgraded docker-deploy.sh script, added REST API tests

LeoNerdoG:
This PR introduces framework for REST API testing. 
The PR consists of three parts:
 - Adding Newman runner for running tests in CI (GHA and Travis - this tool is available in GitHub Marketplace)
 - Enhancing docker-deploy.sh script
 - Adding JSON file with Rest Api tests

The idea is this: We will write tests in Postman, as this is much faster that adding tests via Rest Assured or something else (it is also a much better collaboration tool than any others, so we can divide testing team into smaller fragments and also check the progress of others, etc.). After we add some tests, we will export this to a file "KapuaApiTestingPostman_collection.json", which is located in rest-api/test/ folder. 

The tests will be run after every push and PR (I added a test-restApi stage into GHA yml file), but they can also be run locally:
1. If you don't have Newman on you machine yet, install it with `npm install -g newman`
2. In terminal go to rest-api/test folder
3. Run command `newman run KapuaApiTestingPostman_collection.json`

Output should be something like this: 
Screenshot 2021-02-11 at 13 47 09

 Although Travis is becoming obsolete I have upgraded Travis.yml file, so the rest-api tests will be run also there. 
There are just a couple of tests added in the JSON file, becasue rest-api tests tend to get very long (a lot of lines) and I wanted to avoid CQ.

**Related Issue**
This PR is not related to any specific issue, but is rather a initial commit, which sets up the framework for rest api testing. 

**Description of the solution adopted:**
/

**Screenshots**
/

I have added a "smart" timeout to the docker-deploy.sh script because there were some trobules wtih running Rest API tests.
The problem is, that once you run this script, docker containers get status "Running" although they are not properly started just yet - they need a minute or so to properly start. Without this addition, the Rest API tests will start as soon the docker-deploy.sh scripts completes and the tests will fail, because API container is not deployed. 
Therefore this timeout checks API container log for a certain string every 5 seconds (when this string appears the container is up and running), to see if the container is properly running or not. I have tested this and it works fine.

Signed-off-by: Leonardo Gaube <leonardo.gaube@endava.com>

### DIFF
--- a/.github/workflows/github-actions.yaml
+++ b/.github/workflows/github-actions.yaml
@@ -25,6 +25,28 @@ jobs:
       - run: mvn -v
       - run: mvn -B ${BUILD_OPTS} -DskipTests clean install
       - run: bash <(curl -s https://codecov.io/bash)
+  test-restApi:
+    needs: build-kapua
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-java@v1
+        with:
+          java-version: '8'
+      - uses: actions/cache@v2
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      # deploying docker containers because Rest Api calls need real endpoints
+      - run: ./deployment/docker/unix/docker-deploy.sh -xe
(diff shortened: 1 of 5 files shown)


LeoNerdoG:
Because some of the commits have been done with my old e-mail (I cannot verif myself with it naymore, as it is inactive), I am closing this PR and opening new one, which will pass ECA validation.

PR #3238, opened 2021-02-12
Standalone Job Engine

lorthirk:
Introducing a standalone Job Engine container so that it doesn't live in the Console container anymore.

**Related Issue**
No related issues

The container holds a Jersey REST API application (similar to the main REST API one) that is intended to be called only from other Kapua components (i.e. Broker, Console and REST API). That's why, in the Docker Compose deployment, the HTTP port in not exposed outside but only inside the Compose network.

Since only one Job Engine instance is allowed, bringing the Job Engine outside of the console now allows to deploy multiple Console instances.


qa/integration/pom.xml
```diff
@@ -260,6 +260,10 @@
             <groupId>org.eclipse.kapua</groupId>
             <artifactId>kapua-stream-internal</artifactId>
         </dependency>
+        <dependency>
```
riccardomodanese, line 263: Could you check if rest-api-core is really needed?

riccardomodanese (approved):

Coduz (approved):

PR #3239, opened 2021-02-12
Fixed JobEngineService error messages bundle configuration

Coduz:
This PR fixes the configuration for the Error Messages Bundle for Job Engine Service which is not mapped.

**Related Issue**
_None_

Just added the correct mapping.

**Screenshots**
_None_

_None_


lorthirk:
This will have an impact on #3238. How do we want to proceed?

PR #3242, opened 2021-02-16
Improved AuthenticationServiceShiro validation of credentials

Coduz:
This PR improves the validation of the `LoginCredentials` and `SessionCredential` to avoid that `null` values reach the Realms logic. 
If a credential is `null`, we can immediately return an AuthenticationException for invalid credentials. 

**Related Issue**
_None_

Added checks in `AuthenticationService.login(LoginCredentials)` and `AuthenticationService.authenticate(SessionCredentials)` according to the Credentials type. 
Improved the parsing of the object to avoid instantiation of `new` objects if the credentials are of the correct type.

**Screenshots**
_None_

_None_

### DIFF
--- a/service/security/shiro/src/test/java/org/eclipse/kapua/service/authentication/shiro/AccessTokenCredentialsImplTest.java
+++ b/service/security/shiro/src/test/java/org/eclipse/kapua/service/authentication/shiro/AccessTokenCredentialsImplTest.java
@@ -22,59 +22,58 @@ import org.junit.experimental.categories.Category;
 public class AccessTokenCredentialsImplTest extends Assert {
 
     @Test(expected = NullPointerException.class)
-    public void accessTokenCredentialImplCloneConstructorNull() {
+    public void accessTokenCredentialsImplCloneConstructorNullTest() {
         new AccessTokenCredentialsImpl((AccessTokenCredentials) null);
     }
 
     @Test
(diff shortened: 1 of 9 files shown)


service/security/shiro/src/test/java/org/eclipse/kapua/service/authentication/shiro/AccessTokenCredentialsImplTest.java
```diff
@@ -17,20 +17,9 @@
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 
-import java.lang.reflect.Constructor;
-import java.lang.reflect.Modifier;
-
 @Category(JUnitTests.class)
 public class AccessTokenCredentialsImplTest extends Assert {
 
-    @Test
```
gbarbon, line 26: Hi @Coduz , just a minor comment, I think we should add a unit test for the clone constructor too.

gbarbon (approved):

service/security/shiro/src/test/java/org/eclipse/kapua/service/authentication/shiro/AccessTokenCredentialsImplTest.java
Coduz: Done in c7a1322ef12730464753497a60fb9d7a62075d4a

PR #3260, opened 2021-02-25
Fixed OpenID Connect tokens  for SSO

### LINKED ISSUE #3257: SSO: ID Token should be used in place of the Access Token in the OIDC authentication step
**Describe the bug**
The OpenID Connect (OIDC) Access Token is currently used for authenticating the user with the SSO. However, with some OIDC providers the authentication process fails during the token validation. This is due to the fact that some providers do not include some claims in the Access Token. 

This is actually correct, since the official OIDC specification does not define any claims for the Access Token, while it does that for the ID Token. Moreover, the Access Token should be only a simple string (see below for further information). Thus, we should retrieve the information corresponding to those claims form the ID Token.

**To Reproduce**
Try to authenticate using an OpenID Connect provider that does not provide 'aud', 'iat' or 'sub' claims in the access token. The authentication process will fail in the token validation step, because the required claims are not found.

**Expected behavior**
The ID Token, wich we already obtain from the OIDC provider, should be used in place of the Access Token. In this way the token validation would be fine.

**Screenshots**
_n/a_

**Version of Kapua**
1.5.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[ ] Docker
[x] Others (all)

**Main component affected**
[ ] REST API
[ ] Message Broker
[ ] - Others

**Additional context**
The Access Token used in OIDC (see [here](https://openid.net/specs/openid-connect-core-1_0.html) for the official OIDC specification) is defined in OAuth 2 (since OIDC is defined on top of OAuth 2), see [here](https://tools.ietf.org/html/rfc6749#section-7.1), and no claims are defined in it. Moreover, note that  [here](https://tools.ietf.org/html/rfc6749#section-1.4) the spec says: `An access token is a string representing an authorization issued to the client. The string is usually opaque to the client.`

The current SSO implementation in Kapua is using the Access Token in order to retrieve the 'sub' string, but in the official specification the 'sub' claim is not required to be in this token (even if it's often included by various providers, e.g. Keycloak). Thus it would be better to rely on the ID Token instead.

gbarbon:
This PR fixes the OIDC (OpenID Connect) authentication step by using the Id Token instead of the Access Token.

**Related Issue**
Fixes #3257

The OIDC Id Token is now used in place of the Access Token, because the claims that where extracted and/or checked in the Access Token are not always present in it (while they are present in the Id Token). The claims in the Access Token are not defined by the official OIDC specification, thus they change depending on the OIDC Provider. On the other hand, the Id Token claims are precisely defined in the spec.

Note that now the `idToken` is used in place of the `accessToken` in the JUnit tests when needed, and renames all the instances of `jwt` to `accessToken` (since `jwt` was ambiguous). Moreover, the JWT claim definition is moved to `JwtProcessor. This allows providing a custom claim name in a provider. 

This PR also adds OIDC env var to the `docker-compose.yml` file.

**Screenshots**
_n/a_

_n/a_

### DIFF
--- a/plug-ins/sso/openid-connect/api/src/main/java/org/eclipse/kapua/plugin/sso/openid/JwtProcessor.java
+++ b/plug-ins/sso/openid-connect/api/src/main/java/org/eclipse/kapua/plugin/sso/openid/JwtProcessor.java
@@ -35,4 +35,11 @@ public interface JwtProcessor extends AutoCloseable {
      * @throws OpenIDJwtException if JWT processing fails.
      */
     JwtContext process(final String jwt) throws OpenIDException;
+
+    /**
+     * Return the claim to be extracted form the JWT as user identifier.
+     *
+     * @return the claim in the form of a String
+     */
+    String getExternalIdClaimName();
 }
--- a/assembly/README.md
+++ b/assembly/README.md
@@ -48,10 +48,13 @@ For further information take a look at the `sso.md` manual located in the `docs/
 It is also possible to use a generic OpenID Connect provider, by providing to the console the following environment 
 variables:
 
-- `OPENID_JWT_ISSUER` : the base URL to the OpenID server provider.
-- `OPENID_AUTH_ENDPOINT` :  the endpoint URL to the authentication API.
-- `OPENID_TOKEN_ENDPOINT` : the endpoint URL to the token API.
-- `OPENID_LOGOUT_ENDPOINT` : the URL to the logout endpoint.
+- `OPENID_JWT_ISSUER` : the base URL to the OpenID Connect server provider.
+- `OPENID_CLIENT_ID` : the client id (the default value is `console`).
+- `CLIENT_SECRET` : the client secret (optional).
+- `JWT_AUDIENCE` : the JWT audience (the default value is `console`).
+- `OPENID_AUTH_ENDPOINT` : the endpoint URL to the authentication API (optional, already retrieved via well-known document).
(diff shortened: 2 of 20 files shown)


plug-ins/sso/openid-connect/api/src/main/java/org/eclipse/kapua/plugin/sso/openid/JwtProcessor.java
```diff
@@ -35,4 +35,11 @@
      * @throws OpenIDJwtException if JWT processing fails.
      */
     JwtContext process(final String jwt) throws OpenIDException;
+
+    /**
+     * Return the claim to be extracted form the JWT as user identifier.
+     *
+     * @return the claim in the form of a String
+     */
+    String getClaimName();
```
lorthirk: I would prefer a more descriptive name, since "claim name" is too generic

lorthirk (changes requested):

plug-ins/sso/openid-connect/api/src/main/java/org/eclipse/kapua/plugin/sso/openid/JwtProcessor.java
gbarbon: Method changed to `getExternalIdClaimName`

lorthirk (approved):

Coduz (approved):

PR #3299, opened 2021-04-23
Remove Elasticsearch Transport Client

lorthirk:
Remove Elasticsearch Transport Client since it has been deprecated

**Related Issue**
No related issues


lorthirk:
Apparently we cannot remove `org.elasticsearch.client:transport` and `org.elasticsearch.client:transport-netty4-client` because they're used in `org.eclipse.kapua:kapua-service-elasticsearch-server-embedded`. I think we should consider to not to use embedded servers for integration tests anymore. Removing the ES Embedded Server would allow us to completely remove the ES Transport dependency.

gbarbon:
> Apparently we cannot remove `org.elasticsearch.client:transport` and `org.elasticsearch.client:transport-netty4-client` because they're used in `org.eclipse.kapua:kapua-service-elasticsearch-server-embedded`. I think we should consider to not to use embedded servers for integration tests anymore. Removing the ES Embedded Server would allow us to completely remove the ES Transport dependency.

I agree. I think that @riccardomodanese have almost removed all the embedded servers from the integration tests in his work concerning external consumers.

PR #3300, opened 2021-04-26
Issuer name added in MFA QR code image

### LINKED ISSUE #3256: MFA: the 'Kapua' name should be added in the respective authenticator app entry
When stored in an Authenticator app (e.g. Google Authenticator), the Kapua OTP entry is only showing the name of the user and its account, but not the name of the application (Kapua) itself, while other applications, like Google, are showing it. This means that this information is not present in the QR code that we produce.

The information with the name of Kapua should be included in the QR code.

_n/a_

**Additional context**
Consider this extreme case: Kapua and another application, both with MFA enabled and with an entry on the authenticator app on the phone, have the same user and account. With the current implementation it is possible to distinguish the Kapua OTP code from the one of the other application only thanks to the name of the other application.

gbarbon:
This PR adds the 'issuer' name in the Authenticator app entry containing the user QR second factor.

**Related Issue**
Fix #3256 

The implementation follows the spec detailed [here](https://github.com/google/google-authenticator/wiki/Key-Uri-Format). The 'organization' name is used as issuer and added to the URI string used to build the QR code image. String building has been replaced with `URIBuilder` from Apache.

Note that this fix also allows preventing [Conflicting Accounts](https://github.com/google/google-authenticator/wiki/Conflicting-Accounts) between different apps entries in the Authenticator App.

**Screenshots**
_n/a_

This is not a breaking change, since the code generated in the Authenticator App is only dependent on the key (and not on the issuer). This means that entries in the Authenticator App that have been created before this PR is merged will be still valid, thus they produce still valid authentication codes.

### DIFF
--- a/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/mfa/shiro/MfaOptionServiceImpl.java
+++ b/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/mfa/shiro/MfaOptionServiceImpl.java
@@ -18,6 +18,7 @@ import com.google.zxing.client.j2se.MatrixToImageWriter;
 import com.google.zxing.common.BitMatrix;
 import com.google.zxing.qrcode.QRCodeWriter;
 import org.apache.commons.lang.time.DateUtils;
+import org.apache.http.client.utils.URIBuilder;
 import org.eclipse.kapua.KapuaEntityNotFoundException;
 import org.eclipse.kapua.KapuaException;
 import org.eclipse.kapua.commons.jpa.EntityManager;
@@ -71,6 +72,8 @@ import java.awt.Graphics;
 import java.awt.image.BufferedImage;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.Base64;
 import java.util.Date;
 import java.util.UUID;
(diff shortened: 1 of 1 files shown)


service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/mfa/shiro/MfaOptionServiceImpl.java
```diff
@@ -152,7 +155,7 @@ public MfaOption create(MfaOptionCreator mfaOptionCreator) throws KapuaException
 
             // generating base64 QR code image
             Account account = KapuaSecurityUtils.doPrivileged(() -> accountService.find(finalMfaOptionCreator.getScopeId()));
-            mfaOption.setQRCodeImage(generateQRCode(account.getName(), user.getName(), fullKey));
+            mfaOption.setQRCodeImage(generateQRCode(account.getOrganization().getName(), account.getName(), user.getName(), fullKey));
```
lorthirk, line 158: Have you checked for NPEs if organization or organization name is `null`?

gbarbon: The `create` method of the `AccountServiceImpl` performs the following check:  `ArgumentValidator.notEmptyOrNull(accountCreator.getOrganizationName(), "accountCreator.organizationName");`, thus the organization name is always null.

lorthirk (approved):

Coduz (approved):

PR #3303, opened 2021-05-05
Don't return mfaSecretKey when activating MFA via REST API

### LINKED ISSUE #3279: MFA: the encrypted `mfaSecretKey` should be removed when returned with a GET REST API
The `MfaOption` object also contains the encrypted `mfaSecretKey`. This is used in the authentication of the second factor in the `UserPassCredentialsMatcher`. However, the encrypted `mfaSecretKey` is also returned with a GET REST API (more precisely, by the `findMfa` method in the `Users` class of the REST APIs), even if, in this case, it is not needed at all. 

The encrypted `mfaSecretKey` should be replaced with `null` in the `findMfa`.

_n/a_

**Additional context**
Hypothetically, an attacker that is able to impersonate another user and use the REST APIs, could also obtain the encrypted `mfaSecretKey`. If he also possess the key used to encrypt the `mfaSecretKey`, he would be able to decrypt it (since the mfa keys are encrypted and decrypted with the AES algorithm). So, it would be better to return `null`.

MDeLuise:
Fixes #3279, i.e. the `mfaSecretKey` was returned by the method `findMfa` of `Users` class even if it was not needed.
For security reasons is better to not return it.

In the `findMfa` use the method `setMfaSecretKey` of `mfaOption` object in order to set the `mfaSecretKey` to null before returning the result.

### DIFF
--- a/rest-api/resources/src/main/resources/openapi/user/user.yaml
+++ b/rest-api/resources/src/main/resources/openapi/user/user.yaml
@@ -111,8 +111,6 @@ components:
             userId:
               allOf:
                 - $ref: '../openapi.yaml#/components/schemas/kapuaId'
-            mfaSecretKey:
-              type: string
             trustKey:
               description: |
                 The key of the trusted machine.
@@ -159,7 +157,11 @@ components:
         type: mfaOption
         id: "Tu4UpgmN0A4"
         scopeId: "AQ"
+        createdOn: "2021-05-12T08:04:21.962Z"
+        createdBy: "AQ"
+        modifiedOn: "2021-05-12T08:04:21.962Z"
+        modifiedBy: "AQ"
+        optlock: 1,
         userId: "AQ"
-        mfaSecretKey: "exampleMfaSecretKey"
         trustKey: "exampleTrustKey"
         trustExpirationDate: '2020-31-12T00:00:00.000Z'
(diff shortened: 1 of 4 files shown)


rest-api/resources/src/main/resources/openapi/user/user.yaml
```diff
@@ -111,17 +111,6 @@ components:
...
-              type: string
-            trustKey:
-              description: |
-                The key of the trusted machine.
-              type: string
-            trustExpirationDate:
-              description: |
-                The moment when the machine trust will expire.
-              type: string
-              format: 'date-time'
```
lorthirk: `trustKey` and `trustExpirationDate` are not involved in this PR, so the documentation about them should not be changed. You're probably not seeing them in your responses because you didn't activate any trusted machine, so they're null and as such not shown.

rest-api/resources/src/main/resources/openapi/user/user.yaml
```diff
@@ -159,7 +148,10 @@ components:
...
         scopeId: "AQ"
+        createdOn: "2021-05-12T08:04:21.962Z"
+        createdBy: "AQ"
+        modifiedOn: "2021-05-12T08:04:21.962Z"
+        modifiedBy: "AQ"
+        optlock: 1,
         userId: "AQ"
-        mfaSecretKey: "exampleMfaSecretKey"
-        trustKey: "exampleTrustKey"
-        trustExpirationDate: '2020-31-12T00:00:00.000Z'
```
lorthirk: Same as https://github.com/eclipse/kapua/pull/3303#discussion_r631650031

lorthirk (changes requested):

Coduz (approved):

lorthirk (approved):

Coduz:
Removing from projects since this is tracked by issue

PR #3335, opened 2021-06-18
Do not increase lockout counter error when credentials are correct but no otp is provided

MDeLuise:
**Description**
This PR fixes the following problem:
when mfa is enable for an account and the user try to login with correct username and password but without provide any otp code, the lockout counter error is increased.
This is not the intended behavior because this situation happen when a user provide the correct credentials but use the wrong rest api. Also this is the workflow used by the console in order to check if the mfa is enable for an account.

* In the `UserPassCredentialsMatcher` class is created a new method `doUsernameAndPasswordMatch` which check if provided password is correct for the selected user without consider the otp.
* In the method `assertCredentialsMatch` of the `UserPassAuthenticatingRealm` class it's added the case with correct credentials but no mandatory otp sent.
* New integration test are created.

### DIFF
--- a/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/UserPassAuthenticatingRealm.java
+++ b/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/UserPassAuthenticatingRealm.java
@@ -297,8 +297,8 @@ public class UserPassAuthenticatingRealm extends AuthenticatingRealm {
         }
         Subject currentSubject = SecurityUtils.getSubject();
         Session session = currentSubject.getSession();
-        session.setAttribute("scopeId", userId);
-        session.setAttribute("userId", scopeId);
+        session.setAttribute("scopeId", scopeId);
+        session.setAttribute("userId", userId);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/UserPassAuthenticatingRealm.java
```diff
@@ -211,6 +213,20 @@ protected void assertCredentialsMatch(AuthenticationToken authcToken, Authentica
...
+        KapuaId userId = kapuaInfo.getUser().getId();
+        KapuaId scopeId = kapuaInfo.getUser().getScopeId();
+        boolean hasMfa = false;
+        boolean userAndPasswordMatch = ((UserPassCredentialsMatcher) getCredentialsMatcher()).doUsernameAndPasswordMatch(authcToken, info);
+        try {
+            if (KapuaSecurityUtils.doPrivileged(() -> mfaOptionService.findByUserId(userId, scopeId) != null)) {
+                hasMfa = true;
+            }
+        } catch (KapuaException e) {
+            e.printStackTrace();
```
Coduz: `e.printStackTrace()` is evil.
Please use SLF4J `Logger` and throw the exception if needed.

Coduz (approved):

PR #3341, opened 2021-06-29
Fixed "limitExceed" value always false

MDeLuise:
**Description**
This PR fixes the following problem: when a rest api response contains the `limitExceeded` field, its value is always `false`.

* Added the method to set the correct `limitExceeded` value in the `AbstractRegistryFacade`, then use it in all concrete subclasses
* Added tests

### DIFF
--- a/qa/integration-steps/src/main/java/org/eclipse/kapua/qa/integration/steps/RestClientSteps.java
+++ b/qa/integration-steps/src/main/java/org/eclipse/kapua/qa/integration/steps/RestClientSteps.java
@@ -207,6 +207,27 @@ public class RestClientSteps extends Assert {
         Assert.assertEquals("Wrong response code.", expeted, restResponseCode);
     }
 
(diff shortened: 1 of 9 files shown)


service/commons/storable/internal/src/main/java/org/eclipse/kapua/service/storable/model/AbstractStorableListResult.java
```diff
@@ -53,6 +53,7 @@ public AbstractStorableListResult(List<E> storables, Long totalCount) {
 
         addItems(storables);
         setTotalCount(totalCount);
+        this.limitExceeded = totalCount > storables.size();
```
Coduz: This is not actually correct in the case of the last page. 
With:

- 120 total count
- 100 offset
- 50 limit

this will make `limitExceeded` true with only 20 results, so the `limitExceeded` should be false.

Coduz (changes requested):

Coduz (approved):

PR #3352, opened 2021-07-08
Return 409 Conflict in REST API for ENITY_UNIQUENESS exceptions

lorthirk:
Return 409 Conflict instead of 500 when a duplicate entity is created

**Related Issue**
No related issues


Coduz:
I would suggest to map also 
- `KapuaDuplicateNameException`
- `KapuaDuplicateNameInAnotherAccountError` - this one IDK how to map properly. It might be combined with `KapuaDuplicateNameException`

We might also consider `OptimistickLockException` as 409...

lorthirk:
Well... all of them make sense, actually. I'll think a bit about this, but I am tempted to map those exceptions as well.

Coduz:
We can also make the DuplicateName and DuplicateNameInAnotherAccount as an extension of Uniqueness with uniqueness fields fixed... Don't know if that can be done...

PR #3358, opened 2021-07-21
CORS filter is now configurable for users in kapua-sys child account

MDeLuise:
**Related Issue**
This PR fixes the following problem:
users of one kapua-sys child account cannot configure CORS filters even if they have the permission to do it.

* change the `CorsToolbarGrid`class in order to enable the buttons to add/edit/remove CORS filter correctly
* change the `EndpointInfoServiceImpl` in order to fix the permissions' check
* add integration test regarding CORS filter configuration

### DIFF
--- a/service/endpoint/internal/src/main/java/org/eclipse/kapua/service/endpoint/internal/EndpointInfoServiceImpl.java
+++ b/service/endpoint/internal/src/main/java/org/eclipse/kapua/service/endpoint/internal/EndpointInfoServiceImpl.java
@@ -95,7 +95,9 @@ public class EndpointInfoServiceImpl
 
         //
         // Check Access
-        AUTHORIZATION_SERVICE.checkPermission(PERMISSION_FACTORY.newPermission(EndpointInfoDomains.ENDPOINT_INFO_DOMAIN, Actions.write, null));
+        AUTHORIZATION_SERVICE.checkPermission(
+                PERMISSION_FACTORY.newPermission(EndpointInfoDomains.ENDPOINT_INFO_DOMAIN, Actions.write, endpointInfoCreator.getScopeId())
+        );
 
         //
         // Check duplicate endpoint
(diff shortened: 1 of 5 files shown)


service/endpoint/internal/src/main/java/org/eclipse/kapua/service/endpoint/internal/EndpointInfoServiceImpl.java
```diff
@@ -167,7 +180,11 @@ public EndpointInfo find(KapuaId scopeId, KapuaId endpointInfoId)
 
         //
         // Check Access
-        AUTHORIZATION_SERVICE.checkPermission(PERMISSION_FACTORY.newPermission(EndpointInfoDomains.ENDPOINT_INFO_DOMAIN, Actions.read, scopeId));
+        KapuaId scopeIdPermission = entityManagerSession.doAction(em -> EndpointInfoDAO.find(em, scopeId, endpointInfoId))
```
lorthirk: If `entityManagerSession.doAction(em -> EndpointInfoDAO.find(em, scopeId, endpointInfoId))` returns null, we have a `NullPointerException` here. Let's test this first.

service/endpoint/internal/src/main/java/org/eclipse/kapua/service/endpoint/internal/EndpointInfoServiceImpl.java
```diff
@@ -152,7 +160,12 @@ public void delete(KapuaId scopeId, KapuaId endpointInfoId) throws KapuaExceptio
 
         //
         // Check Access
-        AUTHORIZATION_SERVICE.checkPermission(PERMISSION_FACTORY.newPermission(EndpointInfoDomains.ENDPOINT_INFO_DOMAIN, Actions.delete, null));
+        KapuaId scopeIdPermission = entityManagerSession.doAction(em -> EndpointInfoDAO.find(em, scopeId, endpointInfoId))
```
lorthirk: If `entityManagerSession.doAction(em -> EndpointInfoDAO.find(em, scopeId, endpointInfoId))` returns null, we have a `NullPointerException` here. Let's test this first.

lorthirk (changes requested):

stefanomorson (approved):
Looks good to me.

PR #3368, opened 2021-08-31
external consumers - move application code/configuration to a 3rd module

### LINKED ISSUE #3367: External consumers - move application code/configuration to a 3rd module
External consumers assembly modules contain SpringBoot code and several configuration files that should be moved to an intemediate layer.
So a new module to pack the application (uber jar) should be introduced for each consumer.

riccardomodanese:
Brief description of the PR.
see related issue (#3367)

**Related Issue**
fixes #3367 

moved uber-jar crration step to a new module.

**Screenshots**
none

none


stefanomorson (commented):
A would change the name <consumer>-uber into something more general (e.g. <consumer>-app) so that if the packaging of the jar has to be changed in the future for any reason, the name of the project remains consistent. "app" is also consistent with other projects that we have (e.g. job-engine).

Coduz (approved):

PR #3550, opened 2022-05-06
Add backend validation for display name

MDeLuise:
Brief description of the PR.
This PR fixes the bypass of the display name validation check that can be done via rest api.


MDeLuise:
After a discussion with @Coduz it turned out that the `display name` field actually can be empty or null, thus I'm closing this PR and I will create another one in order to conform the console behavior with the rest API.

PR #3680, opened 2022-12-22
Java11

dseurotech:
Introduces the use of jdk11 alongside java8 (using maven toolchain).

*** Description:

It’s common understanding the there are several good reasons to upgrade the base Java version used by Kapua from the current v1.8 to a more recent one. The one that pushed this step as a top priority is related to the lifecycle of ActiveMQ Artemis which is the messaging platform that will be used by Kapua v2.0. Unfortunately not all the SW components that constitute Kapua are ready to fully move to Java v11 (e.g. the console) so a compromise solution is required.

Generally speaking we would like to port to the highest Java LTS that cope with the following:

ActiveMQ Artemis v2.20 or highest can run on the selected Java version.

All the other Kapua components build and run on the selected java version provided that the necessary updates of the libraries and toolchain is applied

There is a version of the 3rd party components used by Kapua that work with the selected Java version

Current tests and QA should suffice and should not need refactoring

Additional info:

The GWT Console component need to be built and distributed after the Angular Console will be Generally Available (GA) for the time necessary to complete the transition of the user base

Kapua versions using the old messaging platform (ActiveMQ Classic with in-process business logic) should not be migrated to reduce the impact and minimise the validation effort


assembly/api/pom.xml
```diff
@@ -19,14 +19,14 @@
     <parent>
         <artifactId>kapua-assembly</artifactId>
         <groupId>org.eclipse.kapua</groupId>
-        <version>2.0.0-SNAPSHOT</version>
+        <version>2.0.0-j11-SNAPSHOT</version>
```
riccardomodanese: remove j11 version

riccardomodanese (approved):

PR #3683, opened 2022-12-22
Break unwanted module dependencies - ExceptionMappers and Models

dseurotech:
A number of modules depended on rest-api just for exception handling and mapping, with no other correlation to rest apis. Such classes need to be moved in their own module(s) in order to break the dependencies.

image

image


stefanomorson (changes requested):
Can you please add the about.html file to each of the new modules (kapua-commons-rest-errors, kapua-commons-rest-model) before merging ?

dseurotech:
Added the about.html in the new modules, as requested

stefanomorson (approved):

PR #3685, opened 2022-12-23
Feature - Liquibase unlocker tool

Agnul97:
This PR introduces another tool inside the extra module which unlocks stucked beployments caused by a wrong execution of liquibase updates. In particular, when the containers leave the lock on the liquibase update process but  are actually out of the critical section (for example, in the case of a crash)

### DIFF
--- a/extras/liquibaseUnlocker/pom.xml
+++ b/extras/liquibaseUnlocker/pom.xml
@@ -21,7 +21,7 @@
         <version>2.0.0-SNAPSHOT</version>
     </parent>
 
-    <artifactId>liquibaseUnlocker</artifactId>
+    <artifactId>kapua-liquibaseUnlocker</artifactId>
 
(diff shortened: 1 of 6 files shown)


extras/liquibaseUnlocker/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-core</artifactId>
+        </dependency>
+
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
```
Coduz, line 68: Please remove this so that the `maven-compiler-plugin` from the root pom is used.

Coduz (changes requested):

Coduz (approved):

PR #3689, opened 2023-01-11
Fix - users that can modify expiration date of the account in which are defined

Agnul97:
Brief description of the PR.
This PR adds a new check in the update function of the Account Service class to control that a user defined in an account cannot modify its expiration date.  Of course a user of the parent account, provided he has the permissions, it's still able to update the expiration date of such account. 

I've moved the checked for the modification of the exp. date in the admin account above in the update function, this allows me to assume that in the new check I've inserted (the one that verifies in fact if the user defined in the account tries to update the exp. date of it) we are not in the admin account.

### DIFF
--- a/service/account/internal/src/main/java/org/eclipse/kapua/service/account/internal/AccountServiceImpl.java
+++ b/service/account/internal/src/main/java/org/eclipse/kapua/service/account/internal/AccountServiceImpl.java
@@ -174,7 +174,7 @@ public class AccountServiceImpl extends AbstractKapuaConfigurableResourceLimited
             //check if the updated account is an admin account
             if (setting.getString(SystemSettingKey.SYS_ADMIN_ACCOUNT).equals(account.getName())) {
                 //throw exception if trying to set an expiration date for an admin account
-                throw new KapuaIllegalArgumentException("notAllowedExpirationDate", account.getExpirationDate().toString());
+                throw new KapuaAccountException(KapuaAccountErrorCodes.OPERATION_NOT_ALLOWED, null, "Admin account cannot have an expiration date set");
             }
         }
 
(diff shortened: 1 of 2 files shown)


service/account/internal/src/main/java/org/eclipse/kapua/service/account/internal/AccountServiceImpl.java
```diff
@@ -169,13 +169,33 @@ public Account update(Account account) throws KapuaException {
             authorizationService.checkPermission(permissionFactory.newPermission(AccountDomains.ACCOUNT_DOMAIN, Actions.write, account.getScopeId()));
         }
 
+        if (account.getExpirationDate() != null) {
+            SystemSetting setting = SystemSetting.getInstance();
+            //check if the updated account is an admin account
+            if (setting.getString(SystemSettingKey.SYS_ADMIN_ACCOUNT).equals(account.getName())) {
+                //throw exception if trying to set an expiration date for an admin account
+                throw new KapuaIllegalArgumentException("notAllowedExpirationDate", account.getExpirationDate().toString());
```
Coduz: Please change the `throw` of this exception.
KapuaIllegalArgumentException should be used to mark invalid fields by their name. 
So either do

- throw new KapuaIllegalArgumentException("account.expirationDate", account.getExpirationDate().toString());

or

- throw new KapuaAccountException(KapuaAccountErrorCodes.OPERATION_NOT_ALLOWED, null, "Admin account cannot have an expiration date set");

The latter gives more information to the user on what's wrong instead of just only saying that the value of `account.expirationDate` is invalid.

Coduz (changes requested):

Coduz (approved):

PR #3697, opened 2023-01-17
Improve handling and checking of values in Birth Message

### LINKED ISSUE #1237: device registry - device lifecycle test scenario may fail for some network configurations
Testflow:
This bug happened running the integration tests (qa) in a machine connected to the network through a mobile internet router but it can also happen in other cases when the number of interfaces and ip address exceeds the size of the corresponding fields in the device table.
* Run class RunDeviceDataI9nTest in debug mode using Cucumber runner

Result:
The test fails, reporting the follwing error:
```
5:15:02.156 [Camel (camelContext) thread #3 - JmsConsumer[Consumer.eurotech:mainRoutelisteners:EXACTLY_ONCE.VirtualTopic.%3E]] WARN  eclipselink.logging.all - Exception [EclipseLink-4002] (Eclipse Persistence Services - 2.6.3.v20160428-59c81c5): org.eclipse.persistence.exceptions.DatabaseException
Internal Exception: org.h2.jdbc.JdbcSQLException: Value too long for column "CONNECTION_IP VARCHAR(64)": "'<more-than-64-chars-list-of-ip-address>' (87)"; SQL statement:
INSERT INTO dvc_device (accept_encoding, app_framework_version, app_identifiers, attributes, bios_version, client_id, connection_interface, connection_ip, created_on, custom_attribute_1, custom_attribute_2, custom_attribute_3, custom_attribute_4, custom_attribute_5, display_name, firmware_version, iccid, imei, imsi, jvm_version, model_id, modified_on, optlock, os_version, osgi_framework_version, properties, serial_number, status, connection_id, created_by, group_id, id, last_event_id, modified_by, scope_id) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?) [22001-192]
Error Code: 22001
```

The list of interfaces and the list of ip addresses are build by kapua-simulator-kura project in class BirthCertificateBuilder which correctly collects all the available information from the system where it is running. The simulator is used by the tests to generate and send a birth certificate to the Kapua messaging service which in turn tries to create a new registry entry and stick that info in a 64 char field.

Expected behaviour:
The test should succeed.

### LINKED ISSUE #2726: Device registration error when connectionIp exceeds field size
#### Description
If a birth message contains a *connectionIp* metric which has a value that exceeds the length of the field in the device registry DB table, the registration of the device fails.

#### Affected version
Kapua version: 1.0.6

#### Expected result
The length of the value of this metric can vary significantly based on the device, the infrastructure where the device is deployed etc. There should be some rule in place to properly manage the registration when the value contained in the birth message exceeds the db limits.

Coduz:
This PR improves the handling of fields of the Birth Message which create/update the information of a device.

Main improvements are:
- handling of long values for `connectionInterface` #1237
- handling of long values for `connectionIp` #2726
- validation of all fields to provide the correct exception instead of relying on DB exception
- added a bunch of testing

**Related Issue**
This PR fixes #1237 and fixes #2726 

- Added validation for all `DeviceCreator`/`Device` fields.
- for `connectionIp` and `connectionInterfaces` an "overflow" field has been created to handle big values.

**Screenshots**
_None_

A bit of code cleanup and javadoc cleanup

### DIFF
--- a/service/device/registry/api/src/main/java/org/eclipse/kapua/service/device/registry/KapuaDeviceRegistrySettingKeys.java
+++ b/service/device/registry/api/src/main/java/org/eclipse/kapua/service/device/registry/KapuaDeviceRegistrySettingKeys.java
@@ -39,7 +39,7 @@ public enum KapuaDeviceRegistrySettingKeys implements SettingKey {
      *
      * @since 2.0.0
      */
-    DEVICE_REGISTRY_LIFECYCLE_BIRTH_FIELDS_CLOB_LENGTH_MAX("device.registry.lifecycle.birth.fields.clob.length.max"),
+    DEVICE_LIFECYCLE_BIRTH_VAR_FIELDS_LENGTH_MAX("device.lifecycle.birth.var.fields.length.max"),
 
     /**
      * Gets the hard limit for {@link KapuaBirthMessage} {@link DeviceExtendedProperty#getValue()} fields length hdrd limit.
--- a/service/device/registry/api/src/main/java/org/eclipse/kapua/service/device/registry/KapuaDeviceRegistrySettingKeys.java
+++ b/service/device/registry/api/src/main/java/org/eclipse/kapua/service/device/registry/KapuaDeviceRegistrySettingKeys.java
@@ -16,11 +16,11 @@ import org.eclipse.kapua.commons.setting.SettingKey;
 import org.eclipse.kapua.message.device.lifecycle.KapuaBirthMessage;
 
 /**
- * {@link SettingKey}s for {@link DeviceRegistrySettings}.
+ * {@link SettingKey}s for KapuaDeviceRegistrySettings.
  *
  * @since 1.0.0
  */
-public enum DeviceRegistrySettingKeys implements SettingKey {
+public enum KapuaDeviceRegistrySettingKeys implements SettingKey {
 
     /**
      * The key value in the configuration resources.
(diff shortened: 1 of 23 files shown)


service/device/registry/api/src/main/java/org/eclipse/kapua/service/device/registry/KapuaDeviceRegistrySettingKeys.java
```diff
@@ -13,41 +13,61 @@
...
 
     /**
-     * Set up the {@code enum} with the key value provided
+     * Gets the hard limit for {@link KapuaBirthMessage} generic fields length.
      *
-     * @param key
-     *            The value mapped by this {@link Enum} value
+     * @since 2.0.0
+     */
+    DEVICE_REGISTRY_LIFECYCLE_BIRTH_FIELDS_CLOB_LENGTH_MAX("device.registry.lifecycle.birth.fields.clob.length.max"),
```
stefanomorson: For this enum item I would use a name that it is a bit more consistent with the next one and maybe more specific. Something like:

- DEVICE_LIFECYCLE_BIRTH_VAR_FIELDS_LENGTH_MAX (or similar)
or

- DEVICE_LIFECYCLE_BIRTH_CONNECTION_IP_LENGTH_MAX
- DEVICE_LIFECYCLE_BIRTH_CONNECTION_INTERFACE_LENGTH_MAX

Coduz: Changed to be `DEVICE_LIFECYCLE_BIRTH_VAR_FIELDS_LENGTH_MAX`

Coduz: Changed to be `DEVICE_LIFECYCLE_BIRTH_VAR_FIELDS_LENGTH_MAX`

PR #3711, opened 2023-02-02
Add endpoint for changing user password

### LINKED ISSUE #3710: Missing endpoint for changing user password
The REST API should provide a way to change the key for a user password. 

The change should only requires:
* The old password
* The new password

The change password functionality has to implement all the validations needed on the old and new password. If any validation fails the change must fail as well.
The change should be permitted only to the the user that owns the password credential entry.

MDeLuise:
This PR fixes #3710, i.e. adds an endpoint in the REST API for changing the user password.

**Notes**
For now only the changing of the current user is implemented, future improvement should bring the functionality to reset the password of another user (if provided enough permissions).

### DIFF
--- a/qa/common/src/main/java/org/eclipse/kapua/qa/common/MockedLocator.java
+++ b/qa/common/src/main/java/org/eclipse/kapua/qa/common/MockedLocator.java
@@ -83,7 +83,7 @@ public class MockedLocator extends KapuaLocator {
     @Override
     public <S extends KapuaService> S getService(Class<S> serviceClass) {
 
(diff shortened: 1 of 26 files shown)


Coduz (approved):

stefanomorson (commented):
Nice feature, a couple of comments my side. Should the API resource name be changed from UserCredential to UserCredentials (to be consistent with other resources)? 
A minor comment on the PasswordChangeRequest class, why not using currPassword or currentPassword rather than oldPassword? Seems to be more consistent.

PR #3715, opened 2023-02-08
Fix - updated swagger UI version

Agnul97:
This pr updates the swagger UI version. One major change of this new versions is the coloring of json showed in the web UI.


Coduz (approved):

Coduz:
IP review passed

PR #3717, opened 2023-02-14
Create endpoint to retrieve current system `version` and `build version`

### LINKED ISSUE #3716: Add endpoint to retrieve system info
It would be useful to have an endpoint used to retrieve information about the current system (_e.g. version, build version, etc._)

`version`, `build version` and other info should be handled automatically without the need to manually set them at the end of the build process.

MDeLuise:
This PR fixes #3716, i.e. creates an endpoint used to retrieve information about the current system.
For now, the info are: `version` and `build version`, but in the future it can be used to retrieve even other info.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -154,6 +154,7 @@
         <maven-war-plugin.version>3.3.2</maven-war-plugin.version>
         <maven-toolchains-plugin.version>3.1.0</maven-toolchains-plugin.version>
 
+        <buildnumber-maven-plugin.version>3.0.0</buildnumber-maven-plugin.version>
         <checkstyle.version>7.6.1</checkstyle.version>
         <dependency-check-maven.version>7.4.4</dependency-check-maven.version>
         <docker-maven-plugin.version>0.40.2</docker-maven-plugin.version>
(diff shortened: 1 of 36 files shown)


pom.xml
```diff
@@ -556,6 +573,22 @@
                 <artifactId>exec-maven-plugin</artifactId>
                 <version>${exec-maven-plugin.version}</version>
             </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>buildnumber-maven-plugin</artifactId>
+                <version>${buildnumber-maven-plugin.version}</version>
+                <configuration>
```
Coduz: Since the `configuration` and the `executions` are already defined in the plugin in `pluginManagement` they can be removed

pom.xml
```diff
@@ -134,6 +134,7 @@
         <spotify.version>8.15.1</spotify.version>
 
         <!-- Plugins versions -->
+        <buildnumber-maven-plugin.version>3.0.0</buildnumber-maven-plugin.version>
```
Coduz: Optional, just OCD, if you want to move this down to 157 so they are grouped for similar name and sorted

Coduz (changes requested):

Coduz (approved):

PR #3721, opened 2023-02-22
Add functionality to reset password

### LINKED ISSUE #3720: Missing functionality to reset password
The system should provide a way to reset the password credential of a user. Password reset is a different use case from password update in the sense that password change is allowed only for the password credential owned by the currently authenticated user. A user that wants to change his password must provide the current password (for identity verification) and the new password.

Password reset, instead, is executed by a user (typically an administrator user) on the password credential of another user. Password reset does not need to know the current password since this would be a violation of the security principles.

In the password reset use case, the new password can be provided by the user executing the reset or generated by the system. In the latter case there should be a way to safely deliver it to the user being reset.

MDeLuise:
This PR fixes #3720 , i.e. add the functionality of resetting the password credential of a user. In this case, the new password is provided with the reset request by a user with the right permission (usually a system admin).


service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/user/shiro/UserCredentialsServiceImpl.java
```diff
@@ -90,4 +105,58 @@ public Credential changePasswordRequest(PasswordChangeRequest passwordChangeRequ
             return credentialService.update(passwordCredential);
         });
     }
+
+
+    @Override
+    public Credential resetPassword(KapuaId scopeId, KapuaId credentialId, PasswordResetRequest passwordResetRequest) throws KapuaException {
+        Credential credential = credentialService.find(scopeId, credentialId);
```
Coduz: It is better to do first:
- argument validation: to throw proper KapuaIllegalArgumentException
- checkAccess: to avoid guessing of entity existence base on the response time of the request

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/user/shiro/UserCredentialsServiceImpl.java
```diff
@@ -90,4 +105,58 @@ public Credential changePasswordRequest(PasswordChangeRequest passwordChangeRequ
...
+        if (credential == null) {
+            throw new KapuaEntityNotFoundException(Credential.TYPE, credentialId);
+        }
+
+        //
+        // Argument Validation
+        ArgumentValidator.notNull(scopeId, "scopeId");
+        ArgumentValidator.notNull(credentialId, "credential.id");
```
Coduz: This should be `credentialId` since it is the parameter that appears in the docs

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/user/shiro/UserCredentialsServiceImpl.java
```diff
@@ -90,4 +105,58 @@ public Credential changePasswordRequest(PasswordChangeRequest passwordChangeRequ
...
+        ArgumentValidator.notNull(passwordResetRequest.getNewPassword(), "passwordResetRequest.newPassword");
+
+        //
+        // Check access
+        AuthorizationService authorizationService = locator.getService(AuthorizationService.class);
+        PermissionFactory permissionFactory = locator.getFactory(PermissionFactory.class);
+        authorizationService.checkPermission(permissionFactory.newPermission(AuthenticationDomains.CREDENTIAL_DOMAIN, Actions.write, scopeId));
+
+        // Validate Password length
+        String plainNewPassword = passwordResetRequest.getNewPassword();
```
Coduz: It would be possibile to avoid duplication of the password validation?

Coduz (changes requested):

service/security/authentication/api/src/main/java/org/eclipse/kapua/service/authentication/credential/CredentialService.java
```diff
@@ -70,9 +69,21 @@ CredentialListResult query(KapuaQuery query)
...
+
+    /**
+     * Check if the provided password meets all the password's requirements. Return exception if requirements not fulfilled.
+     *
+     * @param scopeId:          The scope ID in which to perform the check
+     * @param plainPassword:    Password to check requirement for
+     * @param argumentName:     Name of the field that contains the password
+     * @throws KapuaException   When something goes wrong
+     */
+    void validatePassword(KapuaId scopeId, String plainPassword, String argumentName) throws KapuaException;
```
Coduz: I don't think that is a good idea to provide the `argumentName` from the caller if this is a ServiceAPI method. 

I would prefer a signature like the following

`void validatePassword(String plainPassword) throws KapuaException`

or 

`boolean isValidPassword(String plainPassword)`

Then the caller (i.e. UserCredentialsServiceImpl) can catch the exception (or use the boolean reult) and throw proper KapuaIllegalArgumentException as needed, with the right argumentName.

What do you think?

PR #3730, opened 2023-03-03
Inserted missing checks on nested entities

Agnul97:
This PR inserts some missing checks of existence on nested entities. 
For example, in the provided commit, the delete function of AccessPermissions endpoints has been revisited in order to check if the provided accessPermission exists inside the provided accessInfo. This kind of control is in place for the find function but not for the delete one (the one modified here), so I wanted to fill the gap.

Other endpoints have this same problem and with this first commit I simply wanted to demonstrate the kind of modifications I would like to do.


rest-api/resources/src/main/java/org/eclipse/kapua/app/api/resources/v1/resources/AccessPermissions.java
```diff
@@ -223,6 +223,13 @@ public Response deleteAccessPermission(
             @PathParam("scopeId") ScopeId scopeId,
             @PathParam("accessInfoId") EntityId accessInfoId,
             @PathParam("accessPermissionId") EntityId accessPermissionId) throws KapuaException {
+
+        try {
+            find(scopeId,accessInfoId,accessPermissionId);
+        } catch (KapuaEntityNotFoundException e) {
+            throw e;
+        }
+
```
dseurotech, line 232: This catches and rethrows the exception - what is the intended improvement?

PR #3777, opened 2023-05-15
FIX - updated readme file to provide a functioning demo setup

Agnul97:
The docker-compose deployment file, in the absence of local docker images, tries to pull the consumers from the docker hub which are, for now, not present in the Kapua docker repo. This means that deploying Kapua following the "demo setup" section of the main project readme file is actually not possible.

In order to provide a functioning demo setup, I changed the instructions provided for that section, explaining that it is possible to exploit the hub in order to deploy tagged released versions but not for snapshot versions (for example, the actual development branch with the consumers), which require a local build. 
Furthermore, I updated the building section of the documentation, because I found some improvements there. I attached this update in this PR because the "quick demo" section actually depends on that document.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,11 +56,19 @@ scenarios.
 **Note:** the Docker Hub repository mentioned above is not the official project repository from Eclipse Foundation.
 ***
 
-Suppose the target is the current snapshot 2.0.0-SNAPSHOT.
-
 * Clone Eclipse Kapua&trade; into a local directory
 * Open an OS shell and move to Kapua project root directory
 * Start Docker runtime
+
+The Kapua repository mentioned above hosts only images of released versions. It is possible to test different versions of Kapua doing a checkout into
+the release branches (for example, "release-1.6.x") and to the tagged commits to select the specific version (for example, the commit tagged as "1.6.7"). Doing so, it is assured
+that the following step will pull proper images from the Docker Hub. If, on the other hand, your target is a SNAPSHOT version (for example, the 2.0.0-SNAPSHOT), a local build is required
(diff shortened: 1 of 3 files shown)


Coduz (approved):

Coduz (changes requested):
After further review also of the linked issue I think that this solution won't work.
Checking out `release-1.7.x` would have the same problem, since snapshot images are not hosted in Docker HUB.

The best solution would be to specify that only released version have Docker Images hosted in Docker HUB. For develop version we need to specify that the user needs to build the images locally if he wants to run a demo on `develop`. Otherwise it needs to checkout a tagged release.

README.md
```diff
@@ -56,11 +56,19 @@ scenarios.
...
 ***
 
-Suppose the target is the current snapshot 2.0.0-SNAPSHOT.
-
 * Clone Eclipse Kapua&trade; into a local directory
 * Open an OS shell and move to Kapua project root directory
 * Start Docker runtime
+
+The Kapua repository mentioned above hosts only images of released versions. It is possible to test different versions of Kapua doing a checkout into
+the release branches (for example, "release 1.6.x") and to the tagged commits to select the specific version (for example, the commit tagged as "1.6.7"). Doing so, it is assured
```
Coduz: release-1.6.x

README.md
```diff
@@ -56,11 +56,19 @@ scenarios.
...
+that the following step will pull proper images from the docker hub. If, on the other hand, your target is a SNAPSHOT version (for example, the 2.0.0 - SNAPSHOT), a local build is required
```
Coduz: Docker Hub
2.0.0-SNAPSHOT

Coduz (approved):

PR #3781, opened 2023-05-26
feat(jobs): Added JobStepProperty overrides on JobEngineStartOptions

Coduz:
This PR adds the capability of override on the fly `JobStepProperties` of the `Job` upon `JobEngineService.start` operation.

This is useful because once started once a Job configuration cannot be updated and you always required to delete the job and re-create which sometimes is not a preferred thing to do.

**Related Issue**
_None_

Added a `List` of `JobStepProperties` to `JobStartOptions` in which is is possible to add multiple `JobStepProperties` that will override any matching `JopStepProperty` defined in any of the `JobStep`.

**Screenshots**
_None_

Further improvements:
- JobStepProperty overrides for specific JobStep
- Report that a JobExecution has started with a different configuration

### DIFF
--- a/job-engine/client/src/main/java/org/eclipse/kapua/job/engine/client/JobStartOptionsClient.java
+++ b/job-engine/client/src/main/java/org/eclipse/kapua/job/engine/client/JobStartOptionsClient.java
@@ -76,7 +76,7 @@ public class JobStartOptionsClient implements JobStartOptions {
 
     @Override
     public void setStepPropertiesOverrides(List<JobStepProperty> stepPropertiesOverrides) {
-        this.stepPropertiesOverrides = stepPropertiesOverrides;
+        this.stepPropertiesOverrides = new ArrayList<>(stepPropertiesOverrides);
     }
 
     @Override
--- a/job-engine/client/src/main/java/org/eclipse/kapua/job/engine/client/JobStartOptionsClient.java
+++ b/job-engine/client/src/main/java/org/eclipse/kapua/job/engine/client/JobStartOptionsClient.java
@@ -14,8 +14,11 @@ package org.eclipse.kapua.job.engine.client;
 
 import org.eclipse.kapua.job.engine.JobStartOptions;
 import org.eclipse.kapua.model.id.KapuaId;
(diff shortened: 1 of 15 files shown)


job-engine/client/src/main/java/org/eclipse/kapua/job/engine/client/JobStartOptionsClient.java
```diff
@@ -56,6 +60,25 @@ public void removeTargetIdToSublist(KapuaId targetId) {
...
+    }
+
+    @Override
+    public void addStepPropertyOverride(JobStepProperty jobStepPropertyOverride) {
+        getStepPropertiesOverrides().add(jobStepPropertyOverride);
+    }
+
+    @Override
+    public void setStepPropertiesOverrides(List<JobStepProperty> stepPropertiesOverrides) {
+        this.stepPropertiesOverrides = stepPropertiesOverrides;
```
dseurotech: consider copying the elements into the inner list, to avoid changing the list implementation with a different one, and side effects from the list changing externally

job-engine/commons/src/main/java/org/eclipse/kapua/job/engine/commons/model/JobStepPropertiesOverrides.java
```diff
@@ -0,0 +1,65 @@
...
+import javax.xml.bind.annotation.XmlRootElement;
+import javax.xml.bind.annotation.XmlType;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+@XmlRootElement(name = "jobStepPropertiesOverrides")
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlType
+public class JobStepPropertiesOverrides implements Iterable<JobStepProperty> {
```
dseurotech, line 29: This looks like a simple wrapper for List<JobStepProperty> - what is the value added by having such wrapper? Why do we need this?

dseurotech: Belay that comment - JAXB is incapable of handling pure lists, hence the wrapper is needed. =(

Issue #3688, opened 2023-01-11
github action optimizations (builds repeated needlessly)


dseurotech:
The current github action "kapua-ci.yml" is heavily under-optimized. A full build is performed as a first step, then for each of the different subtasks a near-full build is performed again (to recreate docker images needed for tests), then a further maven verify is performed in order to execute the specific subset of tests. This is horribly inefficient. 
There is trace in a previous commit of a change in behaviour introduced to mitigate an error in docker layer caching:
https://github.com/eclipse/kapua/commit/9b457b1ceb24e8e1bbdbe5984370111cd2a669b7

That commit needs to be partially reverted, at least to regain the ability to build once and reuse previous caches (maven artifacts/docker layers).

The original github build action (satackey/action-docker-layer-caching@v0.0.11) has been forked and updated since, this could be worth exploring:
original: https://github.com/marketplace/actions/docker-layer-caching
fork: https://github.com/marketplace/actions/docker-layer-caching2

Alternatively, an approach based on artifacts could be considered: https://github.com/orgs/community/discussions/26723#discussioncomment-3253091

or maybe just using the cache:
https://stackoverflow.com/questions/71348621/docker-re-use-container-image-by-caching

Also lookup: https://github.com/marketplace/actions/multi-stage-docker-build

Agnul97:
I've created one PR that re-introduces the maven artifacts caching along with various tweaks, it's the one shown above. (#3785 )

[...]

PR #3785, opened 2023-05-30
FIX - Ci improvement - maven artifact caching and various tweaks

Agnul97:
Brief description of the PR.
This PR improves the CI process, with various tweaks and re-introducing the maven artifacts cache mechanism.

**Related Issue**
This PR addresses https://github.com/eclipse/kapua/issues/3688

I'll present the changes made with a list that covers the different "categories" of improvements, sorted according to the order of implementation of each of them:

1. Duplicated code removal
I've noticed that a lot of code for job definitions was repeated in the kapua-ci.yaml file, especially in the case of jobs that executes the different cucumber tests of the project. This redundancy of code causes difficulty in the maintainability of it, for example, a change that affects the jobs has to be repeated for each of them, with the difficulties associated with this. For example,  I noticed that some of these jobs were using the same GitHub Actions but with different versions. To solve this I used the GitHub composite Actions https://docs.github.com/en/actions/creating-actions/creating-a-composite-action, defining in a separate file the template of these jobs (in the action.yaml file under the "runTestsTaggedAs" folder) that can be "specialized" with the proper tag of cucumber tests to run. To this end, the job definitions inside the Kapua-ci.yaml file re-use this template passing the different tests tag. 

2. Execution of cucumber tests in the proper maven projects
In a past issue related to tests for Kapua, I noticed that cucumber tests reside in a very small set of folders (or, in other words, maven projects). I noticed that, if I ran tests with a naive "mvn verify <test arguments>" a lot of time was spent unnecessarily in the "traversal" of a lot of maven projects not containing tests at all. I wanted to solve this waste of time by launching the maven tests execution only in the projects containing the tests. So, instead of the naive "mvn verify..." I used something like "mvn verify -pl _list of projects containing cucumber tests_". This, obviously, works assuming that maven artifacts and docker images needed by those tests are present in the environment. To retrieve the project containing tests I used the command "grep -l -r --include \*.feature ." 

3. Removal of jobs executing tests no more present in the code-base
I noticed that the jobs _triggerServiceIntegration_ , _triggerServiceIntegrationbase_ , _triggerService_ were trying to launch tests that are no longer present in the code base. Discussing with @Coduz we concluded that probably they may have been moved to the job tests suites, in any case, I will conduct an investigation to try to understand where they have been moved (or lost). I removed those 3 jobs considering that they were using machine time uselessly. 

4. Maven Artifacts Caching
Before this PR, there was an attempt to cache artifacts, exploiting the GitHub Action "actions/cache" but it was incomplete and actually not used. This Action was doing a save-restore for the cache but, despite this, each different job was executing a full build and in this way not re-using the cached artifacts. Other than this, the _key_ parameter used to configure the caching mechanism (precisely _key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}_ ) did not limit the scope of the cache to the specific workflow run but rather to the changes made in the root pom file. The cause of this was that, to propose an example, one could have done a series of commits to the same PR, without modification to the pom file BUT with modifications in the code, and the CI process for each of them would retrieve the same artifacts cache, without noticing changes in the code. To this end, The _key_ parameter has been changed to scope the cache as correctly needed.
This caching allows re-using artifacts in each job. Therefore, a full build is no longer necessary before running the cucumber tests. Rather, it is sufficient to build the docker images only.

The caching of docker images would be another good idea to prevent image re-building for each job. This task presents other difficulties and trade-offs and will be addressed in a separate PR.
 

The execution of tests with the specification of fixed set of maven projects creates the necessity of future modification of the set, if other tests are added in different maven projects. This is the trade-off for the speed-up obtained with the point 2 of the presented list. I think that this won't be a big concern if we consider the fact that the same is, at the moment, also true for the cucumber tags used in the different jobs (one could add tests with different tags and has to remember to add the specific job for it). In other words, a dependency on the kapua-ci file is already present for test addition and removal. 

FINAL NOTES ON THE CORRECTNESS OF THIS PR
I considered from the start the requirement to maintain here the same tests that are run in the old CI process. In this regard, I verified manually that the new CI executes the same tests as the old one (and, also, the tests that are actually present in the code base). I report here the numbers of scenarios for each cucumber tag that I found in the code base and that are executed with this new CI

```
@brokerAcl: 59
@test-tag: 7 
@broker:  25
@device: 7
@devicemanagement: 22
@connection: 11 
@datastore: 56
@user : 30
@userIntegrationBase: 56
@userIntegration : 38
@security: 152 
@jobs,scheduler: 85
@jobsIntegrationBase: 82
@jobsIntegration: 8
@account,translator: 131
@jobEngineStepDefinitions: 5
@jobEngineStartOfflineDevice: 30
@jobEngineStartOnlineDevice: 25
@jobEngineRestartOfflineDevice: 30
@jobEngineRestartOnlineDevice: 14
@jobEngineRestartOnlineDeviceSecondPart: 12
@jobEngineServiceStop: 12
@role,group: 86
@deviceRegistry: 333
@endpoint: 66
```

### DIFF
--- a/.github/actions/runTestsTaggedAs/action.yaml
+++ b/.github/actions/runTestsTaggedAs/action.yaml
@@ -12,8 +12,6 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - name: Clones Kapua repo inside the runner
-      uses: actions/checkout@v3
     - name: Setup java
       uses: actions/setup-java@v3
       with:
@@ -28,15 +26,15 @@ runs:
         restore-keys: |
           ${{ runner.os }}-maven-
     - name: Docker images creation
-      if: ${{ inputs.needs-docker-images }}
+      if: ${{ inputs.needs-docker-images == 'true' }}
       run: mvn -B ${BUILD_OPTS} -Pdocker -DskipTests clean install
       shell: bash
     - name: Dns look-up containers needed for tests - message-broker
-      if: ${{ inputs.needs-docker-images }}
+      if: ${{ inputs.needs-docker-images == 'true' }}
       run: echo "127.0.0.1       message-broker" | sudo tee -a /etc/hosts
       shell: bash
     - name: Dns look-up containers needed for tests - job-engine
-      if: ${{ inputs.needs-docker-images }}
+      if: ${{ inputs.needs-docker-images == 'true' }}
       run: echo "127.0.0.1       job-engine" | sudo tee -a /etc/hosts
       shell: bash
     - name: Test execution step
--- a/.github/actions/runTestsTaggedAs/action.yaml
+++ b/.github/actions/runTestsTaggedAs/action.yaml
@@ -43,7 +43,7 @@ runs:
         timeout_minutes: 45
         retry_on: error
         max_attempts: 1
-        command: ./ci-output.sh mvn -B ${BUILD_OPTS} ${CONFIG_OVERRIDES} -Dgroups='!org.eclipse.kapua.qa.markers.junit.JUnitTests' -Dcucumber.filter.tags=${{ inputs.tag }} -pl ${TEST_PROJECTS} verify
+        command: ./ci-output.sh mvn -B ${BUILD_OPTS} ${CONFIG_OVERRIDES} -Dgroups='!org.eclipse.kapua.qa.markers.junit.JUnitTests' -Dcucumber.filter.tags="${{ inputs.tag }}" -pl ${TEST_PROJECTS} verify
     - name: Code-coverage results
       run: bash <(curl -s https://codecov.io/bash)
       shell: bash
--- a/.github/actions/runTestsTaggedAs/action.yaml
+++ b/.github/actions/runTestsTaggedAs/action.yaml
@@ -43,7 +43,7 @@ runs:
         timeout_minutes: 45
         retry_on: error
         max_attempts: 1
-        command: ./ci-output.sh mvn -B ${BUILD_OPTS} ${CONFIG_OVERRIDES} -Dgroups='!org.eclipse.kapua.qa.markers.junit.JUnitTests' -Dcucumber.filter.tags=${{ inputs.tag }} verify
+        command: ./ci-output.sh mvn -B ${BUILD_OPTS} ${CONFIG_OVERRIDES} -Dgroups='!org.eclipse.kapua.qa.markers.junit.JUnitTests' -Dcucumber.filter.tags=${{ inputs.tag }} -pl ${TEST_PROJECTS} verify
     - name: Code-coverage results
       run: bash <(curl -s https://codecov.io/bash)
       shell: bash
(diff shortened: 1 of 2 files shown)


.github/actions/runTestsTaggedAs/action.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Dns look-up containers needed for tests - message-broker
+      if: ${{ inputs.needs-docker-images == 'true' }}
+      run: echo "127.0.0.1       message-broker" | sudo tee -a /etc/hosts
+      shell: bash
+    - name: Dns look-up containers needed for tests - job-engine
+      if: ${{ inputs.needs-docker-images == 'true' }}
+      run: echo "127.0.0.1       job-engine" | sudo tee -a /etc/hosts
+      shell: bash
+    - name: Test execution step
+      uses: nick-fields/retry@v2.8.1
```
dseurotech: Can this be removed completely? Having a retry plugin with attempts=1 just adds confusion

.github/workflows/kapua-ci.yaml
```diff
@@ -31,710 +30,283 @@ jobs:
...
-          max_attempts: 1
-          command: ./ci-output.sh mvn -B ${BUILD_OPTS} ${CONFIG_OVERRIDES} -Dgroups='!org.eclipse.kapua.qa.markers.junit.JUnitTests' -Dcucumber.filter.tags="@triggerServiceIntegration" verify
-      - run: bash <(curl -s https://codecov.io/bash)
+      - name: Clones Kapua repo inside the runner
+        uses: actions/checkout@v3
+      - uses: ./.github/actions/runTestsTaggedAs
+        with:
+          tag: '@jobsIntegration'
+          needs-docker-images: 'true'
+  #the next 3 tests-suites have been disabled because in the code-base I can't find feature files tagged in this way
```
dseurotech: As a general rule: do not comment, delete. If we need to restore this in the future, we can retrive the old code from previous commits.

dseurotech (approved):

.github/actions/runTestsTaggedAs/action.yaml
Agnul97: I maintained it thinking that the timeout mechanism of 45 mins could be useful for some reason. I would wait what Riccardo has to say to this regard

riccardomodanese: If there is no ather way (or action) to handle the timeout of a single job I prefer to keep the retry action.
Did you check the job 'timeout-minutes' parameter: [timeout-parameter](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes)

Agnul97: I introduced the job timeout parameter, substituting the retry GitAction. Learning the provided doc. I understood that this has an equivalent result

riccardomodanese (approved):

Coduz (approved):

PR #3789, opened 2023-06-08
Fix URL regex marking as invalid legal URL

### LINKED ISSUE #3788: URL regex does not accept valid URL
**Describe the bug**
The existing URL regex does not accept some valid URL, marking them as invalid. https://github.com/eclipse/kapua/blob/2c701fe144152db2c09bc200e8287a4f20ba4ecb/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/util/validator/TextFieldValidator.java#L73

For example `http://172.16.0.2:1234/static/com.eurotech.opcua.driver_1.1.1.dp` is affected by this problem.

MDeLuise:
Some valid URLs was considered illegal, for example the following: `http://172.16.0.2:1234/static/com.eurotech.opcua.driver_1.1.1.dp`.
This PR is to fix this issue.

**Related Issue**
This PR fixes #3788.

Change the regex in order to include more URLs.

Added also the validation check on the backend side, in order to avoid console parameter validation bypass.

### DIFF
--- a/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/util/validator/TextFieldValidator.java
+++ b/console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/util/validator/TextFieldValidator.java
@@ -16,7 +16,6 @@ import com.extjs.gxt.ui.client.widget.form.Field;
 import com.extjs.gxt.ui.client.widget.form.TextField;
 import com.extjs.gxt.ui.client.widget.form.Validator;
 import com.google.gwt.core.client.GWT;
-
 import org.eclipse.kapua.app.console.module.api.client.messages.ValidationMessages;
 
 import java.util.MissingResourceException;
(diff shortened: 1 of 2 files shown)


console/module/api/src/main/java/org/eclipse/kapua/app/console/module/api/client/util/validator/TextFieldValidator.java
```diff
@@ -70,7 +69,7 @@ public enum FieldType {
         ALPHANUMERIC("alphanumeric", "^[a-zA-Z0-9_]+$"),
         NUMERIC("numeric", "^[+0-9.]+$"),
         PACKAGE_VERSION("package_version", "^[a-zA-Z0-9.\\-\\_]*$"),
-        URL("url", "(http(s)?:\\/\\/.)?(www\\.)?[-a-zA-Z0-9@:%._\\+~#=]{2,256}\\.[a-z]{2,6}\\b([-a-zA-Z0-9@:%_\\+.~#?&//=]*)");
+        URL("url", "^(https?:\\/\\/)?(www\\.)?[-a-zA-Z0-9@:%._\\+~#=//]{1,256}\\.[a-zA-Z0-9()//]{2,6}\\b(?:[-a-zA-Z0-9()@:%_\\+.~#?&\\/=]*)$");
```
Coduz, line 72: Isn't this making use of `https ` mandatory?

MDeLuise: It's ok to have a URL without the `https` with this regex, in fact the first part of it have the trailing `?` character (`^(https?:\\/\\/)?`).

Screenshot 2023-06-09 at 16 44 45

Coduz (approved):

PR #3795, opened 2023-06-22
Update `keycloak` docker image

### LINKED ISSUE #3794: Update keycloak docker image
Currently the `keycloak` image that is used is the `jboss/keycloak:7.0.0`.
This image is [very old](https://hub.docker.com/layers/jboss/keycloak/7.0.0/images/sha256-6e6b94949242c9f36bb52be680af3d7bc7903b42813103fc33ad0a811448bf96?context=explore) (7 years), and the jboss repository [is not updated anymore](https://keycloak.discourse.group/t/difference-between-jboss-keycloak-and-keycloak-keycloak/6927/6).

It would be good to move to the official `keycloak/keycloak` image, which is currently on the 21.1 version.

MDeLuise:
This PR is to update `keycloak` image from `jboss/keycloak:7.0.0` to `keycloak/keycloak:21.1`.

**Related Issue**
This PR fixes #3794.

Change the startup script in order to make it compliant with the new image.

### DIFF
--- a/deployment/commons/sso/keycloak/docker/Dockerfile
+++ b/deployment/commons/sso/keycloak/docker/Dockerfile
@@ -12,17 +12,20 @@
 #     Eurotech
 ###############################################################################
 
-FROM jboss/keycloak:7.0.0
+FROM registry.access.redhat.com/ubi9 AS ubi-micro-build
+RUN mkdir -p /mnt/rootfs && \
+    dnf install --installroot /mnt/rootfs curl coreutils --releasever 9 --setopt install_weak_deps=false --nodocs -y && \
+    dnf --installroot /mnt/rootfs clean all
 
-COPY ../entrypoint/run-keycloak /opt/jboss/run-keycloak
+
+FROM keycloak/keycloak:21.1
+
+COPY --from=ubi-micro-build /mnt/rootfs /
+
+COPY ../entrypoint/run-keycloak /opt/run-keycloak
 
 USER 0
 
-RUN chown -R jboss:0 /opt/jboss && \
-    chmod -R g=u /opt/jboss
+VOLUME /opt/keycloak/keycloak/standalone/data
 
-USER jboss
-
-VOLUME /opt/jboss/keycloak/standalone/data
-
-ENTRYPOINT ["/opt/jboss/run-keycloak"]
+ENTRYPOINT ["/opt/run-keycloak"]
--- a/deployment/commons/sso/keycloak/entrypoint/run-keycloak
+++ b/deployment/commons/sso/keycloak/entrypoint/run-keycloak
@@ -30,7 +30,7 @@ if [ ! -f /opt/keycloak/first-run ]; then
   #
   # Start and wait Keycloak
   echo "Starting Keycloak..."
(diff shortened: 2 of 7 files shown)


deployment/commons/sso/keycloak/docker/Dockerfile
```diff
@@ -12,17 +12,20 @@
 #     Eurotech
 ###############################################################################
 
-FROM jboss/keycloak:7.0.0
+FROM registry.access.redhat.com/ubi9 AS ubi-micro-build
+RUN mkdir -p /mnt/rootfs
```
Coduz: Ths is not a published imaged but it would be better to have only `RUN` command with everything concatenated with ` && \` to create fewer layer as possible.

Each RUN instruction creates a layer with makes the final image bigger

Coduz (changes requested):

deployment/commons/sso/keycloak/docker/Dockerfile
MDeLuise: Thanks for the feedback, agree with you.
Updated the `Dockerfile`.

deployment/commons/sso/keycloak/entrypoint/run-keycloak
```diff
@@ -28,9 +30,9 @@ if [ ! -f /opt/jboss/keycloak/standalone/data/first-run ]; then
   #
   # Start and wait Keycloak
   echo "Starting Keycloak..."
-  /opt/jboss/tools/docker-entrypoint.sh -b localhost &
+  KEYCLOAK_ADMIN=$KEYCLOAK_USER KEYCLOAK_ADMIN_PASSWORD=$KEYCLOAK_PASSWORD /opt/keycloak/bin/kc.sh start-dev --http-port=$KC_PORT &
```
dseurotech: Is start-dev the best option here? The main advantange of keycloak's development mode is having HTTP enabled and strict hostname resolution disabled - do we need those? Otherwise I'd prefer to enforce https, even if it is a local machine.

deployment/commons/sso/keycloak/entrypoint/run-keycloak
```diff
@@ -40,15 +42,11 @@ if [ ! -f /opt/jboss/keycloak/standalone/data/first-run ]; then
   #
   # Keycloak Admin credential configuration
   echo "Configuring Keycloak admin credentials..."
-  $KC config credentials \
-        --server http://localhost:8080/auth \
-        --realm master \
-        --user "$KEYCLOAK_USER" \
-        --password "$KEYCLOAK_PASSWORD" ||
-  {
```
dseurotech, line 48: The following bit of code used to inform users when admin configuration went wrong, are we sure we want to omit this?

deployment/docker/compose/sso/docker-compose.keycloak.yml
```diff
@@ -5,7 +5,7 @@ services:
     container_name: kapua-keycloak
     image: kapua/kapua-keycloak:${IMAGE_VERSION}
     ports:
-      - "9090:8080"
+      - "9090:9090"
```
dseurotech: KC_PORT is configurable in the keycloak image - is there any way to share the variable here?

Coduz (approved):

PR #3804, opened 2023-06-29
User/password login cache

riccardomodanese:
Brief description of the PR.
This pr enable the bcrypt caching for user/password authentication. This caching will have performance impact only when multiple login attempts using the same credentials are performed in a short time

**Related Issue**
none

Used local cache to store password and hash.

**Screenshots**
none

none

### DIFF
--- a/commons/src/main/resources/kapua-environment-setting.properties
+++ b/commons/src/main/resources/kapua-environment-setting.properties
@@ -120,6 +120,10 @@ commons.cache.provider.classname=org.eclipse.kapua.commons.service.internal.cach
 #
 commons.cache.local.tmetadata.maxsize=100
 
+commons.authentication-cache.config.size=1000
+#authentication cache ttl (in seconds)
+commons.authentication-cache.config.ttl=60
+
 deployment.name=default-deployment
 
 cluster.name=default-cluster
--- a/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/UserPassCredentialsMatcher.java
+++ b/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/UserPassCredentialsMatcher.java
@@ -12,12 +12,15 @@
  *******************************************************************************/
 package org.eclipse.kapua.service.authentication.shiro.realm;
 
+import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.shiro.ShiroException;
 import org.apache.shiro.authc.AuthenticationException;
 import org.apache.shiro.authc.AuthenticationInfo;
 import org.apache.shiro.authc.AuthenticationToken;
 import org.apache.shiro.authc.credential.CredentialsMatcher;
 import org.eclipse.kapua.KapuaException;
+import org.eclipse.kapua.commons.cache.Cache;
+import org.eclipse.kapua.commons.cache.LocalCache;
 import org.eclipse.kapua.commons.security.KapuaSecurityUtils;
 import org.eclipse.kapua.locator.KapuaLocator;
 import org.eclipse.kapua.service.authentication.ApiKeyCredentials;
@@ -30,11 +33,17 @@ import org.eclipse.kapua.service.authentication.credential.mfa.ScratchCode;
 import org.eclipse.kapua.service.authentication.credential.mfa.ScratchCodeListResult;
 import org.eclipse.kapua.service.authentication.credential.mfa.ScratchCodeService;
 import org.eclipse.kapua.service.authentication.mfa.MfaAuthenticator;
+import org.eclipse.kapua.service.authentication.shiro.AuthenticationServiceShiroImpl;
 import org.eclipse.kapua.service.authentication.shiro.exceptions.MfaRequiredException;
 import org.eclipse.kapua.service.authentication.shiro.mfa.MfaAuthenticatorServiceLocator;
+import org.eclipse.kapua.service.authentication.shiro.setting.KapuaAuthenticationSetting;
+import org.eclipse.kapua.service.authentication.shiro.setting.KapuaAuthenticationSettingKeys;
 import org.eclipse.kapua.service.user.User;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.security.crypto.bcrypt.BCrypt;
 
+import java.util.Base64;
 import java.util.Date;
 
 /**
@@ -44,11 +53,14 @@ import java.util.Date;
  */
 public class UserPassCredentialsMatcher implements CredentialsMatcher {
 
+    private final Logger logger = LoggerFactory.getLogger(AuthenticationServiceShiroImpl.class);
+
     private final KapuaLocator locator;
     private final MfaOptionService mfaOptionService;
     private final ScratchCodeService scratchCodeService;
     private final MfaAuthenticatorServiceLocator mfaAuthServiceLocator;
     private final MfaAuthenticator mfaAuthenticator;
+    private PasswordChecker passwordChecker;
 
     public UserPassCredentialsMatcher() {
         locator = KapuaLocator.getInstance();
@@ -56,6 +68,14 @@ public class UserPassCredentialsMatcher implements CredentialsMatcher {
         scratchCodeService = locator.getService(ScratchCodeService.class);
         mfaAuthServiceLocator = MfaAuthenticatorServiceLocator.getInstance();
         mfaAuthenticator = mfaAuthServiceLocator.getMfaAuthenticator();
+        if (KapuaAuthenticationSetting.getInstance().getBoolean(KapuaAuthenticationSettingKeys.AUTHENTICATION_CREDENTIAL_USERPASS_CACHE_ENABLE, true)) {
+            logger.info("Cache enabled. Initializing CachePasswordChecker...");
+            passwordChecker = new CachePasswordChecker();
+        }
+        else {
+            logger.info("Cache disabled. Initializing NoCachePasswordChecker...");
+            passwordChecker = new NoCachePasswordChecker();
+        }
     }
 
     @Override
@@ -80,7 +100,7 @@ public class UserPassCredentialsMatcher implements CredentialsMatcher {
         boolean credentialMatch = false;
         if (tokenUsername.equals(infoUser.getName()) &&
                 CredentialType.PASSWORD.equals(infoCredential.getCredentialType()) &&
-                BCrypt.checkpw(tokenPassword, infoCredential.getCredentialKey())) {
+                passwordChecker.checkPassword(tokenPassword, infoCredential)) {
 
             if (!mfaAuthenticator.isEnabled()) {
                 credentialMatch = true;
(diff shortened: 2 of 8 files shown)


riccardomodanese:
@Coduz @elbert3 @stefanomorson should I add a couple of parameters for cache size and cache ttl?

commons/src/main/resources/kapua-environment-setting.properties
```diff
@@ -120,6 +120,10 @@ commons.cache.provider.classname=org.eclipse.kapua.commons.service.internal.cach
 #
 commons.cache.local.tmetadata.maxsize=100
 
+commons.authentication-cache.config.size=1000
```
Coduz, line 123: This settings are authentication settings and are betters suit to live in the authentication module. 

Actually we have already a `authentication.credential.userpass.cache.ttl` property in the `kapua-authentication-setting.properties` setting file which we are not using. We might re-use that and add a new `authentication.credential.userpass.cache.size`

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/UserPassCredentialsMatcher.java
```diff
@@ -178,4 +187,18 @@ public boolean doCredentialsMatch(AuthenticationToken authenticationToken, Authe
         return credentialMatch;
     }
 
+    private boolean checkPassword(String tokenPassword, Credential infoCredential) {
+        String cachedCredential = CACHED_CREDENTIALS.get(tokenPassword);
```
Coduz: Perhaps we mught want to do at least a `SHA-xxx` on the key of the map so we don't have clear-text passwords stored in memory?

riccardomodanese: done, please see my changes

Coduz (approved):

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/UserPassCredentialsMatcher.java
```diff
@@ -179,3 +199,42 @@ public boolean doCredentialsMatch(AuthenticationToken authenticationToken, Authe
...
+
+    private static final Cache<String, String> CACHED_CREDENTIALS = new LocalCache<String, String>(
+            KapuaAuthenticationSetting.getInstance().getInt(KapuaAuthenticationSettingKeys.AUTHENTICATION_CREDENTIAL_USERPASS_CACHE_CACHE_SIZE, 1000),
+            KapuaAuthenticationSetting.getInstance().getInt(KapuaAuthenticationSettingKeys.AUTHENTICATION_CREDENTIAL_USERPASS_CACHE_CACHE_TTL, 60),
+            null);
+
+    public boolean checkPassword(String tokenPassword, Credential infoCredential) {
+        String hashedTokenPassword = Base64.getEncoder().encodeToString(DigestUtils.sha3_512(tokenPassword));
+        String cachedCredential = CACHED_CREDENTIALS.get(hashedTokenPassword);
+        if (cachedCredential!=null && cachedCredential.equals(infoCredential.getCredentialKey())) {
```
elbert3, line 219: In the event of a hash collision (where the wrong password is provided but it produces the same hash as the correct password), I assume that `cachedCredential.equals(infoCredential.getCredentialKey())` should return false here?

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/UserPassCredentialsMatcher.java
```diff
@@ -179,3 +199,42 @@ public boolean doCredentialsMatch(AuthenticationToken authenticationToken, Authe
...
+            return true;
+        }
+        else if (BCrypt.checkpw(tokenPassword, infoCredential.getCredentialKey())) {
+            CACHED_CREDENTIALS.put(hashedTokenPassword, infoCredential.getCredentialKey());
```
elbert3, line 223: It looks like we are using the cache when the correct password is used.  However, consider the case where devices have been configured with the wrong password (or it has been changed, has expired, etc.).  I think we should also be using the cache for when the wrong credentials are provided.

elbert3 (changes requested):

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/UserPassCredentialsMatcher.java
```diff
@@ -179,3 +199,42 @@ public boolean doCredentialsMatch(AuthenticationToken authenticationToken, Authe
...
 
 }
+
+interface PasswordChecker {
+
+    boolean checkPassword(String tokenPassword, Credential infoCredential);
+
+}
+
+class CachePasswordChecker implements PasswordChecker {
```
dseurotech, line 209: It would be better to have a password cacher which is independent from the actual password checker implementation - e.g.: a wrapper/decorator around another PasswordChecker. That we we can compose different strategies, and/or test independently the caching algorithm and the password checking one.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/UserPassCredentialsMatcher.java
dseurotech: If the hash algorithm used is appropriate (anche sha3_512 should be) the likelihood of hash collision on wrong passwords should be minimal - particularly if the "wrong" password is very close to the correct one. That said, the cached credential seems to be the hashed password, so there is no way to distinguish it from a match against the correct one.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/UserPassCredentialsMatcher.java
dseurotech: In other words, you'd cached the result of the match, is that right? Either way, we need to be wary of the cache refresh strategy of the cache.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/UserPassCredentialsMatcher.java
```diff
@@ -44,18 +53,29 @@
...
     private final MfaAuthenticator mfaAuthenticator;
+    private PasswordChecker passwordChecker;
 
     public UserPassCredentialsMatcher() {
         locator = KapuaLocator.getInstance();
         mfaOptionService = locator.getService(MfaOptionService.class);
         scratchCodeService = locator.getService(ScratchCodeService.class);
         mfaAuthServiceLocator = MfaAuthenticatorServiceLocator.getInstance();
         mfaAuthenticator = mfaAuthServiceLocator.getMfaAuthenticator();
+        if (KapuaAuthenticationSetting.getInstance().getBoolean(KapuaAuthenticationSettingKeys.AUTHENTICATION_CREDENTIAL_USERPASS_CACHE_ENABLE, true)) {
```
dseurotech, line 71: //TODO: move this logic to the module wiring

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/UserPassCredentialsMatcher.java
```diff
@@ -44,18 +53,29 @@
...
 public class UserPassCredentialsMatcher implements CredentialsMatcher {
 
+    private final Logger logger = LoggerFactory.getLogger(AuthenticationServiceShiroImpl.class);
+
     private final KapuaLocator locator;
     private final MfaOptionService mfaOptionService;
     private final ScratchCodeService scratchCodeService;
     private final MfaAuthenticatorServiceLocator mfaAuthServiceLocator;
     private final MfaAuthenticator mfaAuthenticator;
+    private PasswordChecker passwordChecker;
```
dseurotech, line 63: add "final"

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/UserPassCredentialsMatcher.java
dseurotech, line 209: This cache (like so many others in this project) needs to establish an eviction strategy. As in: if the credentials of the device change (through api/console/etc), we need to remove cached entries related to such device/credential.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/UserPassCredentialsMatcher.java
dseurotech, line 223: Wouldn't it be more efficient to store them the other way around? retrieve by credential//userId and compare to the password hash? Caching by password hash is more likely to create collisions than the other way around

service/security/test-steps/src/main/java/org/eclipse/kapua/service/authentication/steps/AuthenticationServiceSteps.java
```diff
@@ -95,6 +105,24 @@ public void setServices() {
         accessInfoService = locator.getService(AccessInfoService.class);
         accessInfoFactory = locator.getFactory(AccessInfoFactory.class);
         permissionFactory = locator.getFactory(PermissionFactory.class);
+        authenticationService = new AuthenticationServiceShiroImpl();
+    }
+
+    @When("I do a login loop of {int} iterations with user {string} and password {string}")
```
dseurotech, line 111: How is that testing the cache, exactly?

dseurotech (changes requested):
- separate caching logic from password check logic
 - establish cache eviction policy
 - (not directly related to this pr): Why is no one checking the expiration date of the credential? 
 - consider moving the mfa logic in a separate collaborator anyway

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/UserPassCredentialsMatcher.java
stefanomorson: Is the collision a possible event with bcrypt as well ?

riccardomodanese: every hash algorithm has a collision risk because if you would like to have an n-bits output whethever input length you have if the possible inputs are m with m>2^n collisions are expected regardless of the algorithm you can find

riccardomodanese: in other words if we have a password length of m bits and the hash produces n bits with n>m we could have no collisions (it depends on the hashing algorithm) otherwise there is no chance to have no collisions.

riccardomodanese:
replaced by 3811

PR #3805, opened 2023-06-30
FIX - Added exception mapper for duplicate pass exception

Agnul97:
This pr ads an exception mapper for the DuplicatedPasswordCredentialException, in this way the response returned for such exception is more verbose and helps to understand what is causing the authorization problem

### DIFF
--- a/commons-rest/errors/src/main/java/org/eclipse/kapua/commons/rest/errors/KapuaDuplicatePasswordCredentialExceptionMapper.java
+++ b/commons-rest/errors/src/main/java/org/eclipse/kapua/commons/rest/errors/KapuaDuplicatePasswordCredentialExceptionMapper.java
@@ -1,0 +1,47 @@
+/*******************************************************************************
+ * Copyright (c) 2017, 2022 Eurotech and/or its affiliates and others
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


commons-rest/errors/src/main/java/org/eclipse/kapua/commons/rest/errors/KapuaDuplicatePasswordCredentialExceptionMapper.java
```diff
@@ -0,0 +1,47 @@
...
+    @Inject
+    public KapuaDuplicatePasswordCredentialExceptionMapper(ExceptionConfigurationProvider exceptionConfigurationProvider) {
+        this.showStackTrace = exceptionConfigurationProvider.showStackTrace();
+    }
+
+    @Override
+    public Response toResponse(DuplicatedPasswordCredentialException kapuaDuplicatePassException) {
+        LOG.error(kapuaDuplicatePassException.getMessage(), kapuaDuplicatePassException);
+        return Response
+                .status(Status.UNAUTHORIZED)
```
Coduz: Since it is a "Duplicate" kind of exception [409](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409) would be the better code to use. 
409 is also returned for duplicate entities constrain violation

Coduz (changes requested):

Coduz (approved):

PR #3808, opened 2023-07-06
Fix - Inserted device package fields limits 

Agnul97:
This PR adds limits for some fields used in the device management packages service. In this way, a reasonable upper-bound is set in order to allow clients to set the needed fields and discourage a buffer overflow attack.

### DIFF
--- a/service/device/management/packages/internal/src/main/java/org/eclipse/kapua/service/device/management/packages/internal/DevicePackageManagementServiceImpl.java
+++ b/service/device/management/packages/internal/src/main/java/org/eclipse/kapua/service/device/management/packages/internal/DevicePackageManagementServiceImpl.java
@@ -151,6 +151,8 @@ public class DevicePackageManagementServiceImpl extends AbstractDeviceManagement
             throw new KapuaIllegalArgumentException("packageDownloadRequest.uri", packageDownloadRequest.getUri().toString());
         }
 
+        verifyOverflowPackageFields(packageDownloadRequest);
+
         //
         // Generate requestId
         KapuaId operationId = new KapuaEid(IdGenerator.generate());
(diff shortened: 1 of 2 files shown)


service/device/management/packages/internal/src/main/java/org/eclipse/kapua/service/device/management/packages/internal/DevicePackageManagementServiceImpl.java
```diff
@@ -574,4 +576,25 @@ public DevicePackageUninstallOperation uninstallStatus(KapuaId scopeId, KapuaId
         // Check response
         return checkResponseAcceptedOrThrowError(responseMessage, () -> responseMessage.getPayload().getDevicePackageUninstallOperation());
     }
+
+    private void verifyOverflowPackageFields(DevicePackageDownloadRequest packageDownloadRequest) throws KapuaIllegalArgumentException {
+        int uriLength = packageDownloadRequest.getUri().toString().length();
+        int nameLength = packageDownloadRequest.getName().length();
+        int versionLength = packageDownloadRequest.getVersion().length();
+
+        if (uriLength > 2048) {
```
Coduz: Please use `ArgumentValidator.lengthRange`. 
Min value should be passed `null`

service/device/management/packages/internal/src/main/java/org/eclipse/kapua/service/device/management/packages/internal/DevicePackageManagementServiceImpl.java
```diff
@@ -574,4 +576,25 @@ public DevicePackageUninstallOperation uninstallStatus(KapuaId scopeId, KapuaId
...
+            throw new KapuaIllegalArgumentException("packageDownloadRequest.uri", packageDownloadRequest.getUri().toString());
+        }
+
+        if (nameLength > 256) {
```
Coduz: Please use `ArgumentValidator.lengthRange`. 
Min value should be passed `null`

service/device/management/packages/internal/src/main/java/org/eclipse/kapua/service/device/management/packages/internal/DevicePackageManagementServiceImpl.java
```diff
@@ -574,4 +576,25 @@ public DevicePackageUninstallOperation uninstallStatus(KapuaId scopeId, KapuaId
...
+            throw new KapuaIllegalArgumentException("packageDownloadRequest.name", packageDownloadRequest.getName().toString());
+        }
+
+        if (versionLength > 256) {
```
Coduz: Please use `ArgumentValidator.lengthRange`. 
Min value should be passed `null`

Coduz (changes requested):
Please make use of `ArgumentValidator.lengthRange`.
Move the checks on the service before `AuthorizationService.checkPermission` so we can validate any "bad request" before the Auth check and possibly save one auth check. 

Actually, also `packageDownloadRequest.getUri().toURL()` check should be before auth check.

Coduz (approved):

PR #3811, opened 2023-07-10
feat(authentication): add local cache for user/password login

riccardomodanese:
Brief description of the PR.
This pr enable the bcrypt caching for user/password authentication. This caching will have performance impact only when multiple login attempts using the same credentials are performed in a short time.
Replace pr #3804 

Related Issue
none

Description of the solution adopted
Used local cache to store password and hash.

Screenshots
none

Any side note on the changes made
none

### DIFF
--- a/commons/src/main/java/org/eclipse/kapua/commons/setting/system/SystemSettingKey.java
+++ b/commons/src/main/java/org/eclipse/kapua/commons/setting/system/SystemSettingKey.java
@@ -341,6 +341,14 @@ public enum SystemSettingKey implements SettingKey {
      */
     CACHE_TTL("commons.cache.config.ttl"),
     /**
+     * Provide the Authentication Cache TTL
+     */
+    AUTHENTICATION_CACHE_TTL("commons.authentication-cache.config.ttl"),
+    /**
+     * Provide the Authentication Cache Size
+     */
+    AUTHENTICATION_CACHE_SIZE("commons.authentication-cache.config.size"),
+    /**
      * Provide the JCache Expiry Policy. Allowed values: MODIFIED, TOUCHED
      */
     JCACHE_EXPIRY_POLICY("commons.cache.config.expiryPolicy");
--- a/commons/src/main/java/org/eclipse/kapua/commons/setting/system/SystemSettingKey.java
+++ b/commons/src/main/java/org/eclipse/kapua/commons/setting/system/SystemSettingKey.java
@@ -341,14 +341,6 @@ public enum SystemSettingKey implements SettingKey {
      */
     CACHE_TTL("commons.cache.config.ttl"),
     /**
-     * Provide the Authentication Cache TTL
-     */
-    AUTHENTICATION_CACHE_TTL("commons.authentication-cache.config.ttl"),
-    /**
-     * Provide the Authentication Cache Size
-     */
-    AUTHENTICATION_CACHE_SIZE("commons.authentication-cache.config.size"),
-    /**
      * Provide the JCache Expiry Policy. Allowed values: MODIFIED, TOUCHED
      */
     JCACHE_EXPIRY_POLICY("commons.cache.config.expiryPolicy");
--- a/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/cache/CachedCredential.java
+++ b/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/cache/CachedCredential.java
@@ -12,26 +12,22 @@
  *******************************************************************************/
 package org.eclipse.kapua.service.authentication.credential.cache;
 
-import java.util.Base64;
 import java.util.Date;
-
-import org.apache.commons.codec.digest.DigestUtils;
 
 public class CachedCredential {
 
     private Date modifiedOn;
-    private String digest;
-    private String hashed;
+    private String tmpHash;
+    private String hash;
 
-    CachedCredential(Date modifiedOn, String digest, String hashed) {
+    CachedCredential(Date modifiedOn, String tmpHash, String hash) {
         this.modifiedOn = modifiedOn;
-        this.digest = digest;
-        this.hashed = hashed;
+        this.tmpHash = tmpHash;
+        this.hash = hash;
     }
 
-    public boolean isTokenMatches(String tokenPassword, String hashed) {
-        String digest = Base64.getEncoder().encodeToString(DigestUtils.sha3_512(tokenPassword));
-        return this.digest.equals(digest) && this.hashed.equals(hashed);
+    public boolean isTokenMatches(String tmpHash, String hash) {
+        return this.tmpHash.equals(tmpHash) && this.hash.equals(hash);
     }
 
     public boolean isStillValid(Date lastModifiedOn) {
--- a/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/cache/CachedCredential.java
+++ b/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/cache/CachedCredential.java
@@ -1,0 +1,41 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Eurotech and/or its affiliates and others
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Eurotech - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.kapua.service.authentication.credential.cache;
+
+import java.util.Base64;
+import java.util.Date;
+
+import org.apache.commons.codec.digest.DigestUtils;
+
+public class CachedCredential {
+
+    private Date modifiedOn;
+    private String digest;
+    private String hashed;
+
+    CachedCredential(Date modifiedOn, String digest, String hashed) {
+        this.modifiedOn = modifiedOn;
+        this.digest = digest;
+        this.hashed = hashed;
+    }
+
+    public boolean isTokenMatches(String tokenPassword, String hashed) {
+        String digest = Base64.getEncoder().encodeToString(DigestUtils.sha3_512(tokenPassword));
+        return this.digest.equals(digest) && this.hashed.equals(hashed);
+    }
+
+    public boolean isStillValid(Date lastModifiedOn) {
+        return (modifiedOn!=null && !modifiedOn.before(lastModifiedOn));
+    }
+
+}
--- a/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/cache/CachedPasswordMatcher.java
+++ b/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/cache/CachedPasswordMatcher.java
@@ -1,0 +1,53 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Eurotech and/or its affiliates and others
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Eurotech - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.kapua.service.authentication.credential.cache;
+
+import java.util.Base64;
+
+import org.apache.commons.codec.digest.DigestUtils;
+import org.eclipse.kapua.commons.cache.Cache;
+import org.eclipse.kapua.commons.cache.LocalCache;
+import org.eclipse.kapua.service.authentication.credential.Credential;
+import org.eclipse.kapua.service.authentication.shiro.setting.KapuaAuthenticationSetting;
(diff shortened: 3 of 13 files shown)


service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/UserPassCredentialsMatcher.java
```diff
@@ -44,18 +52,30 @@
...
+    //TODO inject????
+    private PasswordMatcher passwordMatcher;
 
     public UserPassCredentialsMatcher() {
         locator = KapuaLocator.getInstance();
         mfaOptionService = locator.getService(MfaOptionService.class);
         scratchCodeService = locator.getService(ScratchCodeService.class);
         mfaAuthServiceLocator = MfaAuthenticatorServiceLocator.getInstance();
         mfaAuthenticator = mfaAuthServiceLocator.getMfaAuthenticator();
+        if (KapuaAuthenticationSetting.getInstance().getBoolean(KapuaAuthenticationSettingKeys.AUTHENTICATION_CREDENTIAL_USERPASS_CACHE_ENABLE, true)) {
```
dseurotech, line 79: This logic will need to be moved to the DI wiring - I understand this is not possible at the time, but please mark this if/else with a comment like "Move construction to DI wiring" or something like that

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/KapuaAuthenticatingRealm.java
```diff
@@ -235,19 +235,28 @@ protected void increaseLockoutPolicyCount(LoginAuthenticationInfo loginAuthentic
...
-        credential.setFirstLoginFailure(null);
-        credential.setLoginFailuresReset(null);
-        credential.setLockoutReset(null);
-        credential.setLoginFailures(0);
-        try {
-            KapuaSecurityUtils.doPrivileged(() -> credentialService.update(credential));
-        } catch (KapuaException kex) {
-            throw new ShiroException("Unexpected error while looking for the lockout policy", kex);
+        //TODO find a proper way to update only if needed (obviously database update has a cost)
+        if (shouldResetCredentialLockout(credential)) {
```
dseurotech, line 239: Shouldn't deciding whether to perform the update be the responsibility of the CredentialService - so that the rules are centralized there? Within the credential service there should still be time to decide whether the db update is needed or not

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/cache/CachedPasswordMatcher.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.apache.commons.codec.digest.DigestUtils;
+import org.eclipse.kapua.commons.cache.Cache;
+import org.eclipse.kapua.commons.cache.LocalCache;
+import org.eclipse.kapua.service.authentication.credential.Credential;
+import org.eclipse.kapua.service.authentication.shiro.setting.KapuaAuthenticationSetting;
+import org.eclipse.kapua.service.authentication.shiro.setting.KapuaAuthenticationSettingKeys;
+import org.springframework.security.crypto.bcrypt.BCrypt;
+
+public class CachedPasswordMatcher implements PasswordMatcher {
```
dseurotech, line 47: This class violates Single Responsibility Principle. Can we split the caching logic in a wrapper, independent from the underlying algorithm?

commons/src/main/java/org/eclipse/kapua/commons/setting/system/SystemSettingKey.java
```diff
@@ -340,6 +340,14 @@ public enum SystemSettingKey implements SettingKey {
      * Provide the Cache TTL
      */
     CACHE_TTL("commons.cache.config.ttl"),
+    /**
```
dseurotech: In the previous PR it was already pointed out that at least one of these properties is already present (https://github.com/eclipse/kapua/pull/3804#discussion_r1250621583) . I'd suggest to either use that, or remove the old one in favour of there new ones.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/utils/AuthenticationUtils.java
```diff
@@ -77,60 +88,31 @@ public static String cryptCredential(CryptAlgorithm algorithm, String plainValue
...
-                    break;
-            }
-
-            //
-            // Return value
-            return salt + ":" + hashedValue;
-        } catch (NoSuchAlgorithmException e) {
-            throw new KapuaRuntimeException(KapuaAuthenticationErrorCodes.CREDENTIAL_CRYPT_ERROR, e);
+    public static String doSha(String plainValue) {
+        KapuaCryptoSetting settings = KapuaCryptoSetting.getInstance();
```
dseurotech, line 92: Please avoid instantiating singletons within the method's body. Instantiate them on top of the class - that makes replacing them with proper Injection much easier in the future

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/utils/AuthenticationUtils.java
```diff
@@ -43,10 +43,21 @@
 public class AuthenticationUtils {
```
dseurotech, line 43: As a general rule, static Util classes do more harm than good. Please consider using injectable (and replaceable, through the use of Interfaces) collaborators.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/cache/CachedPasswordMatcher.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private static final Cache<String, CachedCredential> CACHED_CREDENTIALS = new LocalCache<String, CachedCredential>(
```
dseurotech, line 51: What is the cache eviction policy here, other than the size and the TTL? What if someone changes the password though the api or console? Do they need to wait for 60 seconds to log in again? 
Also: is it 60 seconds or 60 minutes? Always include the measurement unit in constant names for better clarity.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/cache/CachedCredential.java
```diff
@@ -0,0 +1,41 @@
...
+        this.digest = digest;
+        this.hashed = hashed;
+    }
+
+    public boolean isTokenMatches(String tokenPassword, String hashed) {
+        String digest = Base64.getEncoder().encodeToString(DigestUtils.sha3_512(tokenPassword));
+        return this.digest.equals(digest) && this.hashed.equals(hashed);
+    }
+
+    public boolean isStillValid(Date lastModifiedOn) {
```
dseurotech, line 33: Do we really need this level of detail? While checking for token expiration is more than legitimate (actually, necessary), can't we just say that an expired token never matches? Or is there a logic that evicts from the cache expired credentials?

dseurotech (changes requested):
See individual comments for details, but overall I think we can improve in the following areas:
 - SRP violation in caching class
 - No cache eviction policy for outdated passwords
 - No tests
 - possibly duplicate configuration keys

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/utils/AuthenticationUtils.java
riccardomodanese: old code, I just removed SecureRandom instantiation inside the method since it's a costly operation

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/utils/AuthenticationUtils.java
riccardomodanese: old code, I just removed SecureRandom instantiation inside the method since it's a costly operation

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/cache/CachedCredential.java
riccardomodanese: in this way you will increment the failed login by ine and you could lock a user that is logging in with the right credentials

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/cache/CachedPasswordMatcher.java
riccardomodanese: why? the ttl unit is present in the property description in the property file and is clearly written in LocalCache javadoc. Unfortuately there are no factories or other caches available but in this way it's clear which is the unit.
I will never include the unit of measure.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/cache/CachedPasswordMatcher.java
riccardomodanese: this is a layer introduced just to allow a specific Shiro realm to enable/disable depending on the configuration parameter so there is no need to intriduce other layers since we are using only BCrypt algorithm and this is a specific (shiro) authentication/autorization module.
If you inject here a different algorithm you will break the functionality of this module since the BCrypt is used in other parts of this module.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/KapuaAuthenticatingRealm.java
riccardomodanese: the logic is already implemented here becasue if the login succeed then the reset lockout policy is executed. But if lockout policy is already reset this update is not needed. Unfortunately, if I'm right, the DAO layer doesn't perform any check if the entity had changed or not. I don't see any alternatives but, if you see one, please suggest me

commons/src/main/java/org/eclipse/kapua/commons/setting/system/SystemSettingKey.java
riccardomodanese: I thought I used the old one as you suggested, I'll double check the code thanks!

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/UserPassCredentialsMatcher.java
riccardomodanese: I already wrote the comment. see line 62

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/cache/CachedPasswordMatcher.java
dseurotech: the property description describes the meaning of the property to end users, the name of the static variable describes it to developers, making sure they never initialize a cache with the wrong measurement unit. Obviously we are not talking about a critical change, so leave it as it is if you feel it's more in line with how settings are defined elsewhere - we can always return on this point later.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/cache/CachedPasswordMatcher.java
dseurotech: The rest of the code being heavily dependent on a specific algorithm is bad enough, I agree. 
I'd still prefer to see a reusable caching wrapper which need to know nothing of the underlying mechanism - otherwise we are just compounding the problem (the day we need to change the encryption we need to change the invocation to BCrypt both here AND in the Default implementation.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/KapuaAuthenticatingRealm.java
dseurotech: The DAO layer might not perform checks, but we still have a service layer we can act upon. And whether to invoke the DAO (repository, in the future) or not should be the responsibility of such service, IMHO - otherwise we risk having to replicate this logic elsewhere, should the need to change any of those field arise in the future.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/UserPassCredentialsMatcher.java
dseurotech: seen it now!

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/utils/AuthenticationUtils.java
dseurotech: We will have to discuss this, I understand it's debatable.
While I can appreciate that in theory PRs should be as laser focused as possible, I believe that works only in a semi-sane environment - which is not our case.
The only way to get out of the mess of code we have at the moment is to fix things as soon as you have to touch them for whatever reason - otherwise we are always postponing fixing issues to an hypothetical future "big cleanup" - which will never be scheduled due to external pressure, as history confirms.

elbert3 (approved):
I think we should also cache cases where an invalid password is used.  But that can be addressed as a separate issue, so I'm okay proceeding with this implementation.

Issue #3469, opened 2021-12-01
Unable to create Put configuration batch job


pintify:
**Describe the bug**
When I try to create a Put Configuration step in a batch job, it appears an `Error during Persistence Operation`. The issue seems to be related with the length of the configuration as a short configuration (under 65535 characters) works fine.

**To Reproduce**
1. Create a new batch job
2. Click on new Step
3. Apply a valid but large configuration (over 65535 characters)
4. See error:

   <details>
      <summary>Stack Trace: java.sql.SQLException: Data too long for column 'property_value' at row 1</summary>

      ```
      Query: InsertObjectQuery(org.eclipse.kapua.service.job.step.definition.internal.JobStepPropertyImpl@21baaaff)
         at org.eclipse.persistence.exceptions.DatabaseException.sqlException(DatabaseException.java:333)
         at org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor.executeDirectNoSelect(DatabaseAccessor.java:908)
         at org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor.executeNoSelect(DatabaseAccessor.java:970)
         at org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor.basicExecuteCall(DatabaseAccessor.java:640)
         at org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor.executeCall(DatabaseAccessor.java:567)
         at org.eclipse.persistence.internal.sessions.AbstractSession.basicExecuteCall(AbstractSession.java:2099)
         at org.eclipse.persistence.sessions.server.ClientSession.executeCall(ClientSession.java:313)
         at org.eclipse.persistence.internal.queries.DatasourceCallQueryMechanism.executeCall(DatasourceCallQueryMechanism.java:277)
         at org.eclipse.persistence.internal.queries.DatasourceCallQueryMechanism.executeCall(DatasourceCallQueryMechanism.java:263)
         at org.eclipse.persistence.internal.queries.DatasourceCallQueryMechanism.insertObject(DatasourceCallQueryMechanism.java:413)
         at org.eclipse.persistence.internal.queries.StatementQueryMechanism.insertObject(StatementQueryMechanism.java:167)
         at org.eclipse.persistence.internal.queries.StatementQueryMechanism.insertObject(StatementQueryMechanism.java:182)
[...]

PR #3812, opened 2023-07-12
Add maximum length to `configuration put` step definition

MDeLuise:
Due to the fact that the database column `property_value` [is of type](https://github.com/eclipse/kapua/blob/bad8d2946040de44e97aecfd51eca0a764cfd6f3/service/job/internal/src/main/resources/liquibase/0.3.0/job_step_properties.xml#L36) `TEXT`, a configuration cannot have more than 65535 characters.
This because the column type `TEXT` can hold data [up to about 64 Kilobytes size](https://stackoverflow.com/a/13932834/20088695).

This PR adds a validation in the frontend that permits to illegal requests to fail fast without even sending them to the backend.

**Related Issue**
This PR relates to #3469.

A proper maximum length is forced to the textarea.


Coduz (changes requested):
Please add the `ArgumentValidation.lengthRange(...)` invocation in the backend service

Coduz (approved):

PR #3817, opened 2023-07-17
FIX - introduction of sonarCloud scans for pullrequests coming from external forks

Agnul97:
This PR changes both the Kapua-ci and sonarCloud workflows in order to allow sonarCloud scans for pull-requests opened from different forks. In fact, for those PRs, there was a problem reading the SONAR_TOKEN secret configured inside this repo and this PR fixes this, implementing a solution proposed here: https://community.sonarsource.com/t/how-to-use-sonarcloud-with-a-forked-repository-on-github/7363/32.

The idea is to leverage the GitHub actions Workflow_run event https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run. Basically, With this construct, it's possible to launch a workflow triggered by another one... considering that the triggered workflow resides in the default branch, it can read secrets, and here comes the solution to the before-mentioned problem. Considering that this triggered workflow is not "born" from the opened PR it doesn't have the commits of the PR, for this reason, the data from the PR (practically, the code) must be retrieved using other techniques. 

**How I tested this fix**
I tested this solution on my fork. First of all, I inserted in my develop branch the workflows here modified. Then, asking a contributor to open an external PR from his fork and seeing that his commits on the PR were triggering this sonarCloud scan, I concluded that the solution was working and that similar behavior will be expected for the eclipse/kapua repo and external forks.

for now, the scan works only for pull requests and is not working for push events on the default branch or release branches (as was before). This because I just would like to test the work done and the re-introduce this behavior

### DIFF
--- a/.github/workflows/kapua-ci.yaml
+++ b/.github/workflows/kapua-ci.yaml
@@ -1,4 +1,4 @@
-name: kapua-ci
+name: kapua ci
 on: [ push, pull_request ] # Triggers the workflow on push or pull request events
 
 env: #the last 2 env variables defines respectively the maven projects were cucumber tests resides and maven projects were consumers applications needed for docker images building resides.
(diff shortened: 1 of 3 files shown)


.github/workflows/kapua-ci.yaml
```diff
@@ -1,4 +1,4 @@
-name: Kapua CI
+name: kapua ci
```
Coduz: This can be reverted to the original

Coduz (approved):

PR #3823, opened 2023-07-19
Fix `GET /messages` api returns empty result if limit set over 10000

### LINKED ISSUE #3822: Datastore `GET /messages` REST API returns empty result if limit is set over 10000
**Describe the bug**
Datastore `GET /messages` REST API returns empty result if the limit parameter is set over 10000. 
10000 is the limit of Elasticsearch. 

**To Reproduce**
1. Send some messages through the platform
2. Call `GET /messages` REST API with a limit greater then 10000
3. See the response

**Expected behavior**
Messages should be returned or, if this is a limit of the platform (Elasticsearch component), then an appropriate error should be returned.

MDeLuise:
This PR introduces a check of the `limit` parameter in the `GET /messages` API. If the check fails, an appropriate error message is sent as response.

Consider a platform state with some stored messages.
Result of `GET /messages` with `limit = 10000` before the PR:
```
{
  "type": "storableListResult",
  "limitExceeded": false,
  "size": 0,
  "items": [],
  "totalCount": 0
}
```

Result of `GET /messages` with `limit = 10000` after the PR:
```
{
  "type": "illegalArgumentExceptionInfo",
  "httpErrorCode": 400,
  "message": "An illegal value was provided for the argument limit: Value over than allowed max value. Max value is 10000.",
  "kapuaErrorCode": "ILLEGAL_ARGUMENT",
  "argumentName": "limit",
  "argumentValue": "Value over than allowed max value. Max value is 10000"
}
```

**Related Issue**
This PR fixes #3822.

* Add the property `datastore.query.limit.max` which represents the maximum legal value for the `limit` parameter of the `GET /messages` API
* Check the consistency of the `limit` parameter in the `GET /messages` API

### DIFF
--- a/service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/MessageStoreServiceImpl.java
+++ b/service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/MessageStoreServiceImpl.java
@@ -67,6 +67,7 @@ public class MessageStoreServiceImpl extends KapuaConfigurableServiceBase implem
     protected PermissionFactory permissionFactory;
 
     protected static final Integer MAX_ENTRIES_ON_DELETE = DatastoreSettings.getInstance().getInt(DatastoreSettingsKey.CONFIG_MAX_ENTRIES_ON_DELETE);
+    protected static final Integer MAX_LIMIT_VALUE = DatastoreSettings.getInstance().getInt(DatastoreSettingsKey.MAX_LIMIT_VALUE);
     protected final MessageStoreFacade messageStoreFacade;
 
     /**
@@ -181,6 +182,9 @@ public class MessageStoreServiceImpl extends KapuaConfigurableServiceBase implem
     public MessageListResult query(MessageQuery query)
             throws KapuaException {
         checkDataAccess(query.getScopeId(), Actions.read);
+        if (query.getLimit() != null) {
+            ArgumentValidator.numRange(query.getLimit(), 0, MAX_LIMIT_VALUE, "limit");
+        }
         try {
             return messageStoreFacade.query(query);
         } catch (Exception e) {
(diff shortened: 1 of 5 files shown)


service/datastore/internal/src/main/resources/kapua-datastore-settings.properties
```diff
@@ -32,3 +32,7 @@ datastore.cache.metadata.local.size.maximum=1000
 
 # Datastore index prefix
 datastore.index.prefix=
+
+#
+#maximum value for limit parameter
+datastore.limit.max_value=10000
```
Coduz: I think that `datastore.query.limit.value.max` o `datastore.query.limit.max` would be better for the scope of the setting

service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/MessageStoreServiceImpl.java
```diff
@@ -166,6 +167,7 @@ public DatastoreMessage find(KapuaId scopeId, StorableId id, StorableFetchStyle
     public MessageListResult query(MessageQuery query)
             throws KapuaException {
         checkDataAccess(query.getScopeId(), Actions.read);
+        ArgumentValidator.numRange(query.getLimit(), 0, MAX_LIMIT_VALUE, "limit");
```
Coduz: Does the limit change from being nullable to be mandatory?

Coduz (changes requested):

service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/MessageStoreServiceImpl.java
MDeLuise: A request to `.../v1/_/data/messages?sortDir=DESC&offset=0` (so without the `limit` parameter) would still be ok, since the `limit` parameter has a default value of 50.

Coduz (approved):

Coduz (approved):

PR #3837, opened 2023-08-02
Refactorings/exterminate statics

### LINKED ISSUE #3266: 'static final' use should be avoided in class fields
**Describe the bug**
All the `static final` usages should be removed, since they are impacting with property overriding in tests. Factories, services and locators (which are often assigned to `static final` fields) are the same for every class, thus one will intuitively use the `static final`. However, since they are all initialised form the classloader, their initialisation might have an impact to other things, like configurations.

As an example, consider the class [UserPassCredentialsMatcher](https://github.com/eclipse/kapua/blob/develop/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/UserPassCredentialsMatcher.java). If one tries to override the property defined by the key `SystemSettingKey.DB_CONNECTION_ADDITIONAL_OPTIONS` (note that this property is not used in the `UserPassCredentialsMatcher` class), the property is not overridden because the `UserPassCredentialsMatcher` class (which is first initialised by Shiro) is first initialising all the services,. Thus the override of the property is coming too late, when the Db Connection has already been initialised by the services called in the class.

What is happening with test is that the `static` are referenced in a sort of chain, make the services being initialised before the test can set the properties that it wants to override.

**To Reproduce**
Define a test that overrides a given property (e.g. the one defined above), then try to run the test and debug it. The property will maintain its original value.

**Expected behavior**
`static final` should not be used. In such a way services are instantiated later w.r.t. properties, and properties can be correctly overridden.

**Screenshots**
_n/a_

**Version of Kapua**
1.5.0-SNAPSHOT

**Type of deployment**
[ ] Local Vagrant deployment
[ ] Docker
[x] Others (all)

**Main component affected**
[ ] REST API
[ ] Message Broker
[x] - Others (all tests)

**Additional context**
First noticed by @riccardomodanese

dseurotech:
Brief description of the PR.
Removes most static variables and manual singletons, in favour of a more pervasive Dependency Injection usage.

**Related Issue**
This PR fixes/closes [3266](https://github.com/eclipse/kapua/issues/3266)

[Exterminating Statics.pdf](https://github.com/eclipse/kapua/files/14063410/Exterminating.Statics.pdf)

* A lot of code cleanups
* Reworked the structure of data store classes (iteraction with Elastic Search)


Coduz:
IP review started! 
https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/12906

Coduz (approved):

PR #3850, opened 2023-08-22
fix KapuaPrincipal name

riccardomodanese:
Brief description of the PR.
The name of the KapuaPrincipal is not unique now since a user can connect more than one client so potentially more than one device could be connected with the same principal

**Related Issue**
none

Name now is composed by concatenation of username and client id making it unique in the system.

**Screenshots**
none

none

### DIFF
--- a/client/security/src/main/java/org/eclipse/kapua/client/security/bean/KapuaPrincipalImpl.java
+++ b/client/security/src/main/java/org/eclipse/kapua/client/security/bean/KapuaPrincipalImpl.java
@@ -45,7 +45,7 @@ public class KapuaPrincipalImpl implements KapuaPrincipal {
      * @param authResponse
      */
     public KapuaPrincipalImpl(AuthResponse authResponse) {
-        name = authResponse.getUsername();
+        name = authResponse.getUsername() + "@" + authResponse.getClientId();
         clientId = authResponse.getClientId();
         accessTokenId = authResponse.getAccessTokenId();
         userId = KapuaEid.parseCompactId(authResponse.getUserId());
@@ -60,10 +60,11 @@ public class KapuaPrincipalImpl implements KapuaPrincipal {
      * @param accountId
      * @param clientId
      */
-    public KapuaPrincipalImpl(KapuaId accountId, String clientId) {
+    public KapuaPrincipalImpl(KapuaId accountId, String name, String clientId) {
         internal = true;
-        this.clientId = clientId;
         this.accountId = accountId;
+        this.name = name + "@" + clientId;
+        this.clientId = clientId;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


client/security/src/main/java/org/eclipse/kapua/client/security/bean/KapuaPrincipalImpl.java
```diff
@@ -60,10 +60,11 @@ public KapuaPrincipalImpl(AuthResponse authResponse) {
      * @param accountId
      * @param clientId
      */
-    public KapuaPrincipalImpl(KapuaId accountId, String clientId) {
+    public KapuaPrincipalImpl(KapuaId accountId, String name, String clientId) {
```
elbert3: Maybe change this parameter to "username" to distinguish it from the principal "name"?

client/security/src/main/java/org/eclipse/kapua/client/security/bean/KapuaPrincipalImpl.java
```diff
@@ -128,13 +125,6 @@ public boolean equals(Object obj) {
             return false;
         }
         KapuaPrincipalImpl other = (KapuaPrincipalImpl) obj;
-        if (accountId == null) {
-            if (other.accountId != null) {
-                return false;
-            }
-        } else if (!accountId.equals(other.accountId)) {
```
elbert3, line 135: Wouldn't it be possible to have the same username and clientId in multiple accounts?

elbert3 (commented):
Hi @riccardomodanese, I added a couple questions about your changes.

client/security/src/main/java/org/eclipse/kapua/client/security/bean/KapuaPrincipalImpl.java
riccardomodanese: no because the username is unique in the platform (it's not possible to have the same username in 2, or more, different acocunt).
If a user connects with multiple clients each client should have different clientId (otherwise the broker disconnects the first client connected)

client/security/src/main/java/org/eclipse/kapua/client/security/bean/KapuaPrincipalImpl.java
riccardomodanese: done

client/security/src/main/java/org/eclipse/kapua/client/security/bean/KapuaPrincipalImpl.java
elbert3: Ah okay, I didn't realize that the username is unique, but I was able to confirm this.  In that case, I agree that just comparing the principal name would seem sufficient.

elbert3 (approved):

PR #3854, opened 2023-08-28
Security Broker plugin improvement

riccardomodanese:
Brief description of the PR.
Add a map to keep track of the last publish time for each address and lowering some log line since the current info log is too verbose and, as usual, prevent to see the important informations.
Other log lines should be set in debug level or removed but let's do it step by step.

**Related Issue**
none

see description

**Screenshots**
none

none

### DIFF
--- a/broker/artemis/plugin/src/main/java/org/eclipse/kapua/broker/artemis/plugin/security/AddressMap.java
+++ b/broker/artemis/plugin/src/main/java/org/eclipse/kapua/broker/artemis/plugin/security/AddressMap.java
@@ -1,0 +1,42 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Eurotech and/or its affiliates and others
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Eurotech - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.kapua.broker.artemis.plugin.security;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class AddressMap {
+
+    protected Map<String, Long> accessMap;
+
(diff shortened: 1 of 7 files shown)


broker/artemis/plugin/src/main/java/org/eclipse/kapua/broker/artemis/plugin/security/AddressMap.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public void remove(String address) {
+        accessMap.remove(address);
+    }
+
+    public Long get(String address) {
+        return accessMap.get(address);
+    }
+
+    public int sie() {
```
elbert3: Fix name to "size"

broker/artemis/plugin/src/main/java/org/eclipse/kapua/broker/artemis/plugin/security/AddressMap.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Contributors:
+ *     Eurotech - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.kapua.broker.artemis.plugin.security;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class AddressMap {
```
elbert3: To me "AddressMap" seems to imply that it is a "Map" implementation.  Maybe "AddressTracker" or "AddressAccessTracker" would be better?

broker/artemis/plugin/src/main/java/org/eclipse/kapua/broker/artemis/plugin/security/AddressMap.java
```diff
@@ -0,0 +1,42 @@
...
+
+    protected Map<String, Long> accessMap;
+
+    AddressMap() {
+        //TODO could be used a "normal" map?
+        accessMap = new ConcurrentHashMap<String, Long>();
+    }
+
+    public void refresh(String address) {
```
elbert3: "Refresh" is probably okay, but what do you think of "update" for this method?

elbert3 (approved):

PR #3858, opened 2023-09-04
feat(authentication): injectable credential helpers

Coduz:
This PR makes the hendling of `LoginCredentials` and `SessionCredentials` more flexible, to improve extensibility of the AuthenticationService implementation

**Related Issue**
This PR is based on #3840 made by @dseurotech as a starting point for the feature

Created new API in `kapua-security-shiro` module named `CredentialsHelper` from which to extend to provide many different Credentials to be used for authenticating

**Screenshots**
_None_

Added some missing javadoc around


service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
```diff
@@ -132,6 +133,13 @@ public class AuthenticationServiceShiroImpl implements AuthenticationService {
 
     private final UserService userService;
 
+    @Inject
```
dseurotech: Avoid mixing property injection and constructor injection - the latter is preferrable.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationModule.java
```diff
@@ -115,6 +121,27 @@ public ServiceModule authenticationServiceModule(AccessTokenService accessTokenS
                 ));
     }
 
+    @ProvidesIntoSet
+    public LoginCredentialsHandler usernamePasswordCredentialsHandler() {
```
dseurotech: consider adding @Singleton annotation for each of these objects.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/CredentialsHandler.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * {@link CredentialsHandler} definition.
+ * <p>
+ * It maps a {@link AuthenticationCredentials} to a specific implementation of Apache Shiro {@link AuthenticationToken}.
+ *
+ * @param <C> The type of {@link AuthenticationCredentials}
+ * @since 2.0.0
+ */
+public interface CredentialsHandler<C extends AuthenticationCredentials> {
```
dseurotech: The generics here add little to no value. Just pass AuthenticationCredentials to the methods canProcess and mapToShiro directly.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/LoginCredentialsHandler.java
dseurotech, line 1: These interface reifications add little to no value to the pattern, and add unneccessary complexity to the pattern. The idea is that CrendentialHandlers are indistinguishable from one another.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/SessionCredentialsHandler.java
dseurotech, line 1: These interface reifications add little to no value to the pattern, and add unneccessary complexity to the pattern. The idea is that CrendentialHandlers are indistinguishable from one another.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/JwtCredentialsHandler.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Override
+    public ImmutablePair<AuthenticationToken, Optional<String>> mapToShiro(LoginCredentials loginCredentials) throws KapuaAuthenticationException {
+        JwtCredentialsImpl jwtCredentials = JwtCredentialsImpl.parse((JwtCredentials) loginCredentials);
+        final String openIDidToken = Optional.ofNullable(jwtCredentials.getIdToken())
+                .filter(token -> !Strings.isNullOrEmpty(token))
+                .orElseThrow(() -> new KapuaAuthenticationException(KapuaAuthenticationErrorCodes.INVALID_LOGIN_CREDENTIALS));
+
+        return ImmutablePair.of(jwtCredentials, Optional.of(openIDidToken));
```
dseurotech: This is the only case where we use the right side of the pair, to store a value which is directly extracted from the left side. This smells like a forced abstraction.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
```diff
@@ -428,20 +421,30 @@ public AccessToken refreshAccessToken(String tokenId, String refreshToken) throw
     @Override
     public LoginInfo getLoginInfo() throws KapuaException {
         LoginInfo loginInfo = accessTokenFactory.newLoginInfo();
+
```
dseurotech, line 392: avoid empty comment lines if possible

service/security/authentication/api/src/main/java/org/eclipse/kapua/service/authentication/SessionCredentials.java
dseurotech, line 1: Redundant class. In the end the only thing that matters is that we have an AuthenticationCredentials underneath, is it not?

service/security/authentication/api/src/main/java/org/eclipse/kapua/service/authentication/LoginCredentials.java
dseurotech, line 1: Redundant class. In the end the only thing that matters is that we have an AuthenticationCredentials underneath, is it not?

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
```diff
@@ -132,6 +133,13 @@ public class AuthenticationServiceShiroImpl implements AuthenticationService {
 
     private final UserService userService;
 
+    @Inject
+    private Set<LoginCredentialsHandler> loginCredentialsHandlers;
+
+    @Inject
+    private Set<SessionCredentialsHandler> sessionCredentialsHandlers;
+
```
dseurotech: Why differentiate, if both sets in the end deal with instances of AuthenticationCredentials?

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/AccessTokenCredentialsHandler.java
```diff
@@ -0,0 +1,48 @@
...
+public class AccessTokenCredentialsHandler implements SessionCredentialsHandler {
+
+    @Override
+    public boolean canProcess(SessionCredentials loginCredentials) {
+        return loginCredentials instanceof AccessTokenCredentials;
+    }
+
+    @Override
+    public ImmutablePair<AuthenticationToken, Optional<String>> mapToShiro(SessionCredentials loginCredentials) throws KapuaAuthenticationException {
+        AccessTokenCredentialsImpl accessTokenCredentials = AccessTokenCredentialsImpl.parse((AccessTokenCredentials) loginCredentials);
```
dseurotech: why delegate the parsing logic to a static method elsewhere? THIS is the place to put it!

service/security/authentication/api/src/main/java/org/eclipse/kapua/service/authentication/LoginCredentials.java
Coduz: As it is the API right now, LoginCredentials and its extension can be used with:

- `AuthenticationService.login`
- `AuthenticationService.verifyCredentials`

Without this distinction it would be possible to provide AccessTokenCredentials to above methods, which is not supported/allowed

service/security/authentication/api/src/main/java/org/eclipse/kapua/service/authentication/SessionCredentials.java
Coduz: As it is the API right now, SessionCredentials and its extension can be used with:

- `AuthenticationService.authenticate`

Without this distinction it would be possible to provide UserPassCredentials (or other extensions of LoginCredentials) to above methods, which is not supported/allowed

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationModule.java
Coduz: I've matched what was on #3840 

If you think it is better to have those as `@Singleton`, I'll add it

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
Coduz: So you mean adding parameters to the Constructor of `AuthenticationServiceShiroImpl`? 

BTW I've added following #3840

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
Coduz: To distinguish the two types of authentication that we have.

The first login (handled by `login` with `LoginCredentials` and meant to create a session on EC side) and the continuation of that session (handled by `authenticate` with `SessionCredentials` to recover that session). 

So following the same logic of `AuthenticationCredentials` splitting into `LoginCredentials` and `SessionCredentials`.

This API design is something that we can change, if we want but maybe it would be better change in another PR to refactor the API and then apply the changes on this PR.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/AccessTokenCredentialsHandler.java
Coduz: On this, I've followed the #3840 changes. 

I can change for all CredentialsHandler

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/JwtCredentialsHandler.java
Coduz: That one comes out from how the SSO session specific task are handled. 

It would be much nicer if the KapuaSession is created from the CredentialHelper, so we can have different many types of Session according to the credentials given. This for sure has its own challenges, but I think it is outside of the scope of this PR

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
dseurotech: 3840 was just a proof of concept to be expanded upon =)
Yes, either use constructor base injection of property injection (prefer the first) for all collaborators, best to not mix the two strategies, otherwise one is never sure whether an object is fully built or not.

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationModule.java
dseurotech: Avoiding instantiating multiple copies of an object is always best from the memory perspective - as long as they are thread safe (and in this case, I think they are).

service/security/authentication/api/src/main/java/org/eclipse/kapua/service/authentication/SessionCredentials.java
dseurotech: Just to clarify the scenario in my head: why are LoginCredential and SessionCredential both extending AuthenticationCredentials? What is the difference between "authenticate" and "login" methods in AuthenticationService? Their implementation seems nearly identical.

EDIT: I see now that more details are provided below, see following comments

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
dseurotech: I'm not 100% sure regarding this mechanism, and would probably review it. Still, for the purpose of this PR I'd propose at least to split the method CredentialsHandler.canProcess into:

* canLogin
* canAuthenticate

and handle the difference there - you can then invoke each method from the appropriate AuthenticationService branch, without creating too many classes

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/AccessTokenCredentialsHandler.java
dseurotech: Yes, please do - credential parsing should never have been part of a data class in the first place. The correct thing to do is to move all logic within CredentialsHandler instances

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/realm/JwtCredentialsHandler.java
dseurotech: See the solution proposed in https://github.com/Coduz/kapua/pull/11 - subclassing AuthenticationToken to provide the additional "openIdToken" property is cleaner, because each implementation can specify whether such token CAN be provided or is a fixed-value Optional.empty()

That said, should the need to create different KapuaSession based on the given credentials (to the point that no common logic can be found, I mean) would definitely justify the approach you suggest - let's burn that bridge when we get there

dseurotech (approved):

dseurotech (approved):

PR #3890, opened 2023-10-17
fix consumers dlq configuration

riccardomodanese:
Brief description of the PR.
This pr cleans the Camel routes configuration (some old configuration was still present) and reorganizes dlq queues name
Introduces the $SYS prefix for internal use. This prefix is used for dlq and further internal feature and is not meant to be subscribed by external devices.

**Related Issue**
none

none

**Screenshots**
none

none, the effect of the changes is bound to internal components.
NO BACKPORT


assembly/events-broker/configurations/broker.xml
```diff
@@ -111,25 +111,10 @@
...
-                <auto-create-jms-topics>true</auto-create-jms-topics>
-                <default-address-routing-type>ANYCAST</default-address-routing-type>
-            </address-setting>
             <!--default for catch all-->
             <address-setting match="#">
-                <dead-letter-address>DLQ</dead-letter-address>
-                <expiry-address>ExpiryQueue</expiry-address>
+                <dead-letter-address>dlq</dead-letter-address>
+                <expiry-address>expired</expiry-address>
                 <redelivery-delay>0</redelivery-delay>
```
dseurotech, line 117: Missing $SYS/ prefix, is that right?

riccardomodanese: not forgotten, it's the service broker used for authentication so for now I preferred to leave the queues without prefixes.
Do you suggest to use the prefix here (service broker) also? (in the event broker I prefer to leave namespace as it is now)

elbert3: To me it doesn't seem like a bad idea to use this separate namespace for internal messages, though I realize it's for a separate broker and adds a little overhead.  Otherwise I think the PR is fine and will go ahead and approve.

elbert3 (approved):

dseurotech (approved):

assembly/events-broker/configurations/broker.xml
riccardomodanese: To be honest I'm not sure about which is the best solution. I mean, the telemetry broker handles both external load (messages from clients and Camel consumers) and internal (like the dlq for example)
On the other side the event and service brokers handle only internal load now (events are meant to be used only by internal components now and the authentication service is performed by a service consumer built for this scope).
Anyway having services and events deployed in the same broker could lead to a namespace conflicts.

riccardomodanese: I think what you are suggesting worth to be implemented now.
I'm proposing a refactoring like this one (I'm going to implement it in a new commit)

Telemetry broker (only dlq):
$SYS/dlq/default
$SYS/dlq/expired

Event-broker:
$SYS/svc/evt/EVENT_ADDRESS

Service-broker:
$SYS/svc/ath/req
$SYS/svc/ath/resp/brokerId

Event/Service-broker dlq: 
$SYS/svc/dlq/default
$SYS/svc/dlq/expired

I think it's a good proposal but I'm looking forward for your feedback!

elbert3: I agree that this separate namespace should help avoid conflicts and provide more flexibility with how components are deployed.  I think it would also be helpful to consider now how it should be structured/used.  For the telemetry dlq for example, should it also specify the service/consumer - i.e. "$SYS/telemetry/dlq/default", etc.?

riccardomodanese: I preferred to use MSG so the new addresses (last commit) are:

Telemetry broker
$SYS/MSG/dlq/default
$SYS/MSG/dlq/expired

Event-broker
$SYS/EVT/EVENT_ADDRESS

Service-broker
$SYS/SVC/auth/req
$SYS/SVC/auth/resp/brokerId

$SYS/SVC/dlq/default
$SYS/SVC/dlq/expired

I moved event prefix out of svc.
What do you think?

elbert3: Works for me!

PR #3894, opened 2023-10-31
FIX - Keycloak server execution failing at startup

Agnul97:
The startup of the keycloak server inside the kapua/kapua-keycloak image was failing due to missing of some parameters in the startup script. Basically, the server was started in production mode but this mode requires additional configurations of some aspects, as stated here https://www.keycloak.org/server/configuration-production, that was missing.

With this PR, I inserted the missing configuration parameters. In this way, with the kapua/kapua-keycloak image (so, with the --sso option on the deploy) we propose a demo image to test the sso feature. Finally, I updated the documentation, particularly in the part were it is explained how to run a stand-alone keycloak image for the case of a (real) production mode deployment.

Specifically, I worked on these aspects

1. Inserted missing cli parameters to fix the server startup issue and provide an HTTP deployment
2. Fixed the way SSL certificates were imported with the _volumes_ setting in the compose files: this was bugged for a typo in the path and certs were not imported at all
3. Used certs at point 1. to provide the capability of an HTTP + SSL deployment, with a dedicated env. variable to enable/disable SSL if needed (KEYCLOAK_SSL_ENABLE, like done in the console with KAPUA_SSL_ENABLE) and automatically set via the --ssl option with the deployment script
4. Updated the developer guide in the SSO section: this was not updated after the bump to the image version 21.1 (from the jboss 7 version) and I've fixed the gap, also I touched other aspects that were wrong
5. Updated the README.md file inside the assembly folder: I noticed that there was a lot of redundant content with respect to the developer guide, so I inserted some references to that guide in order to solve the redundancy


dseurotech (approved):

deployment/commons/sso/keycloak/entrypoint/run-keycloak
```diff
@@ -15,22 +15,25 @@
...
-  KC_PORT="9090"
-  KC_URL="http://localhost:$KC_PORT"
+  KC_URL="http://localhost:$KEYCLOAK_PORT_HTTP"
   KC=/opt/keycloak/bin/kcadm.sh
 
   echo "Kapua Keycloak Configuration:"
   echo "    Kapua Console URL: $KAPUA_CONSOLE_URL"
   echo "    Keycloak Realm: $REALM_NAME"
   echo "    Keycloak Admin User: $KEYCLOAK_USER"
+  echo "    Keycloak TLS disabled:  $KEYCLOAK_DISABLE_SSL"
```
Coduz: Please don't use this reverse logic ENV to enable component. 

It is far easier to understand `KEYCLOAK_SSL_ENABLE` when is defined and used. 
The default can be `false` to indicate that a feature is disabled by default.

Coduz (changes requested):
First, I want to suggest to switch to `KEYCLOAK_SSL_ENABLE` instead of `KEYCLOAK_DISABLE_SSL` which enables SSL when is `false` and disable SSL `true`. We can talk about this, if there is any doubt on this point

Second, it would be possible to have the Keycloak SSL enabled based on the `--ssl` option available in the Docker Compose deployment? "One SSL option to rule them all" it is easier to use.

Please also remove `KEYCLOAK_*` env defined in `deployment/docker/compose/docker-compose.yml` for the `kapua-console` component.

Agnul97:
> First, I want to suggest to switch to `KEYCLOAK_SSL_ENABLE` instead of `KEYCLOAK_DISABLE_SSL` which enables SSL when is `false` and disable SSL `true`. We can talk about this, if there is any doubt on this point
> 
> Second, it would be possible to have the Keycloak SSL enabled based on the `--ssl` option available in the Docker Compose deployment? "One SSL option to rule them all" it is easier to use.
> 
> Please also remove `KEYCLOAK_*` env defined in `deployment/docker/compose/docker-compose.yml` for the `kapua-console` component.

Coduz (approved):

PR #3912, opened 2023-11-30
FIX - CI configuration improvements

Agnul97:
Basically, this PR is a "partial" port of this previous PR merged in develop https://github.com/eclipse/kapua/pull/3785 that involved a lot of optimizations in the ci configuration.

Here, I changed these aspects of the configuration:

1. The caching mechanism was redundant between the _actions/setup-java@v3_ and _actions/cache@v3_ gitActions. I removed the caching in the first one to only have it in the second one.
2. The key used for the caching was wrong, for the same reason explained in point 4 of the pr https://github.com/eclipse/kapua/pull/3785. I corrected it in the same way introducing the same "pr-commit-unique" key. 
3. removed the restore-key parameter that was afflitected by the same problem of the point 3.

### DIFF
--- a/.github/workflows/kapua-ci.yaml
+++ b/.github/workflows/kapua-ci.yaml
@@ -20,8 +20,6 @@ jobs:
         with:
           path: ~/.m2/repository
           key: ${{ github.run_id }}-${{ github.run_number }}-maven-cache
-          restore-keys: |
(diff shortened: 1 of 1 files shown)


riccardomodanese (approved):
looks good! thanks!

PR #3914, opened 2023-12-04
Fix MFA scratch codes and QR code not working

### LINKED ISSUE #3909: Enabling MFA returns encrypted MFA scratch codes
**Describe the bug**
When enabling the MFA the scratch codes are returned encrypted, so they can't be used.

**To Reproduce**

1. Login to the system
2. Enable MFA
3. See the scratch codes

**Expected behavior**
The scratch codes should be given unencrypted.

**Screenshots**
image

MDeLuise:
This PR fix the issue that prevent the scratch codes and QR code to work.
Now the scratch codes are displayed as plain text.

**Related Issue**
This PR fixes #3909.

### DIFF
--- a/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/mfa/shiro/MfaOptionServiceImpl.java
+++ b/service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/mfa/shiro/MfaOptionServiceImpl.java
@@ -12,6 +12,20 @@
  *******************************************************************************/
 package org.eclipse.kapua.service.authentication.credential.mfa.shiro;
 
+import java.awt.Color;
+import java.awt.Graphics;
+import java.awt.image.BufferedImage;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Base64;
+import java.util.Date;
+import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
+import javax.imageio.ImageIO;
+
 import com.google.zxing.BarcodeFormat;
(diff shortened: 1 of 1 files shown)


service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/mfa/shiro/MfaOptionServiceImpl.java
```diff
@@ -212,7 +208,8 @@ public ScratchCodeListResult createAllScratchCodes(TxContext tx, ScratchCodeCrea
             // Create code
             ScratchCodeImpl codeImpl = new ScratchCodeImpl(scratchCodeCreator.getScopeId(), scratchCodeCreator.getMfaOptionId(), encryptedCode);
 
-            ScratchCode scratchCode = scratchCodeRepository.create(tx, codeImpl);
+            ScratchCode scratchCode = txManager.execute(tx -> scratchCodeRepository.create(tx, codeImpl));
+            scratchCode.setCode(code);
```
Coduz, line 214: Please add a comment inline on why we are doing this `setCode` after the creation so next person here will figure that out right away :D

Coduz (changes requested):
Just a small comment on the `setCode`

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/credential/mfa/shiro/MfaOptionServiceImpl.java
MDeLuise: Thanks for the feedback, just updated the changes 😁

Coduz (approved):

PR #3920, opened 2023-12-13
FIX - Fix offset limit datastore api

Agnul97:
For a technical limit imposed by ES, the maximum number of hits for a query is 10k, even when in the store there are more than 10k documents. Currently, the only check performed in our back end regarding the values offset & limit of our rest APIs regarding queries to ES is that the limit value must be <= to 10k. This is not wrong but it's not complete, in fact the real limit imposed by ES is that the values limit + offset must not be more than 10k. 

This pr introduce this new check on the back-end, along with a test for the feature. This has been done augmenting the check on the max limit value of 10k with the new more "strict" check on the sum offset,limit

### DIFF
--- a/service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/ClientInfoRegistryServiceImpl.java
+++ b/service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/ClientInfoRegistryServiceImpl.java
@@ -63,7 +63,8 @@ import java.util.Optional;
 public class ClientInfoRegistryServiceImpl implements ClientInfoRegistryService {
 
     private static final Logger LOG = LoggerFactory.getLogger(ClientInfoRegistryServiceImpl.class);
-
+    
+    private  Integer MAX_RESULT_WINDOW_VALUE;
     private final StorablePredicateFactory storablePredicateFactory;
     private final AccountService accountService;
     private final AuthorizationService authorizationService;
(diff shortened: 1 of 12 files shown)


service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/ClientInfoRegistryServiceImpl.java
```diff
@@ -134,8 +133,13 @@ public ClientInfoListResult query(ClientInfoQuery query)
 
         ArgumentValidator.notNull(query, QUERY);
         ArgumentValidator.notNull(query.getScopeId(), QUERY_SCOPE_ID);
-
+        if (query.getLimit() != null && query.getOffset() != null) {
```
riccardomodanese, line 139: could you find a way to make this code common to Channel and Messaege instead of replicating it?

Agnul97: I already noticed, while working on this, that the classes ClientInfoRegistryServiceImpl, MessageStoreServiceImpl,ClientInfoRegistryImpl, MetricInfoRegistryImpl have a lot of code in common. My idea was to find a common father class that has the redundant code and the 4 classes specialize him with the different *facade classes. I tried to work on this but It was harder than I thought...part of this becase the generics pose a limitation on how much you can abstract the code. Also I remember that I talked about this with @dseurotech and he shared to me similar concerns about this work. I think that what you are pointing with this comment is a little part of a more general work that I did not manage to conclude

riccardomodanese (approved):

PR #3924, opened 2023-12-20
Feature - Add option to disconnect a device from the broker

elbert3:
This adds the option to request that a DeviceConnection be disconnected from the broker via the REST API.

**Related Issue**
None

There was an existing "disconnect" method in the DeviceConnectionService API that had never been implemented and had been marked as deprecated.  This adds an implementation for the disconnect request, but using the KapuaId of the connection instead of the clientId string.  It works by raising a service event which the broker is listening for.

To implement this I added a new DeviceConnectionEventListenerService to setup the event listening, and the ServerPlugin registers itself with this service as a receiver of those messages.  When it receives a disconnect event, it makes use of the existing BrokerEventHandler to perform the disconnection.

In order for this to work I also had to have the ServerPlugin setup the KapuaLocator, startup the ServiceModuleBundle, and add a JAXBContextProvider.

**Screenshots**
None

None

### DIFF
--- a/commons/src/main/java/org/eclipse/kapua/commons/event/ServiceInspector.java
+++ b/commons/src/main/java/org/eclipse/kapua/commons/event/ServiceInspector.java
@@ -115,8 +115,6 @@ public class ServiceInspector {
         Method matchingMethod = null;
 
         if (methods.length > 0) {
-            List<Class<?>> methodParamTypes = Arrays.asList(method.getParameterTypes());
-
             for (Method candidate : methods) {
                 if (!candidate.getName().equals(method.getName())) {
                     continue;
@@ -124,10 +122,7 @@ public class ServiceInspector {
                 if (!candidate.getReturnType().equals(method.getReturnType())) {
                     continue;
                 }
-
-                List<Class<?>> candidateParamTypes = Arrays.asList(method.getParameterTypes());
-
-                if (candidateParamTypes.size() != methodParamTypes.size() || !candidateParamTypes.containsAll(methodParamTypes)) {
+                if(!Arrays.equals(method.getParameterTypes(), candidate.getParameterTypes())) {
                     continue;
                 }
 
--- a/service/device/registry/test-steps/src/main/java/org/eclipse/kapua/service/device/registry/steps/BrokerSteps.java
+++ b/service/device/registry/test-steps/src/main/java/org/eclipse/kapua/service/device/registry/steps/BrokerSteps.java
@@ -13,15 +13,18 @@
  *******************************************************************************/
 package org.eclipse.kapua.service.device.registry.steps;
 
-import com.google.inject.Singleton;
-import io.cucumber.java.After;
-import io.cucumber.java.Before;
-import io.cucumber.java.Scenario;
-import io.cucumber.java.en.And;
-import io.cucumber.java.en.Then;
-import io.cucumber.java.en.When;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import javax.inject.Inject;
+
 import org.eclipse.kapua.KapuaException;
 import org.eclipse.kapua.broker.artemis.plugin.security.setting.BrokerSetting;
+import org.eclipse.kapua.commons.core.ServiceModuleBundle;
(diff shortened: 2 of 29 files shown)


commons/src/main/java/org/eclipse/kapua/commons/event/ServiceInspector.java
```diff
@@ -115,19 +115,14 @@ private static Method getMatchingMethod(Class<?> clazz, Method method) throws No
...
                     continue;
                 }
                 if (!candidate.getReturnType().equals(method.getReturnType())) {
                     continue;
                 }
-
-                List<Class<?>> candidateParamTypes = Arrays.asList(method.getParameterTypes());
-
-                if (candidateParamTypes.size() != methodParamTypes.size() || !candidateParamTypes.containsAll(methodParamTypes)) {
+                if(!Arrays.equals(method.getParameterTypes(), candidate.getParameterTypes())) {
```
riccardomodanese, line 125: did you move this code to avoid NPE?

service/device/registry/test-steps/src/main/java/org/eclipse/kapua/service/device/registry/steps/BrokerSteps.java
```diff
@@ -515,6 +527,34 @@ public void disconnectClient(String clientName) throws Exception {
         }
     }
 
+    @When("I Force Disconnect connection with client id {string}")
+    public void disconnectConnectionWithClientId(String clientId) throws Exception {
+        DeviceConnection deviceConnection = deviceConnectionService.findByClientId(SYS_SCOPE_ID, clientId);
```
riccardomodanese, line 528: could I suggest to use a parameter to provide the scopeId? (so the test can be invoked also for children accounts)

service/device/registry/test-steps/src/main/java/org/eclipse/kapua/service/device/registry/steps/BrokerSteps.java
```diff
@@ -515,6 +527,34 @@ public void disconnectClient(String clientName) throws Exception {
...
+    }
+
+    @Then("Device(s) status is {string} within {int} second(s) for client id {string}")
+    public void deviceStatusIs(String expectedStatus, int timeout, String clientId) throws Exception {
+        String currentStatus = null;
+        while(!expectedStatus.equals(currentStatus) && timeout-->0) {
+            Thread.sleep(1000);
+
+            logger.info("Device(s) status countdown check: {}", timeout);
+            DeviceConnection deviceConnection = deviceConnectionService.findByClientId(SYS_SCOPE_ID, clientId);
```
riccardomodanese, line 545: could I suggest to use a parameter to provide the scopeId? (so the test can be invoked also for children accounts)

riccardomodanese (approved):

Coduz (approved):

commons/src/main/java/org/eclipse/kapua/commons/event/ServiceInspector.java
elbert3: This actually fixes a bug I found with the previous implementation where it was not always finding the correct method since it just uses the `containsAll` check instead of comparing the parameter type at each position.  For example, if the candidateParamTypes list is (String, KapuaId), `containsAll` would return `true` when comparing another list with (String, String) or (KapuaId, String) or (KapuaId, KapuaId).

I updated it to instead use the Arrays.equals comparison which checks that they have the same number of elements, compares for equality at each position of the array, and it removes the need to first convert to a List.

service/device/registry/test-steps/src/main/java/org/eclipse/kapua/service/device/registry/steps/BrokerSteps.java
elbert3: Since it's generated, would it be possible to specify a scopeId from a cucumber test?  I agree that it would be a good idea to extend this (and other tests) to also check child accounts.  But I think it would require a lot of other changes (i.e. setting up the accounts, connecting with those credentials, etc.) and might be better to do as a separate item.

service/device/registry/test-steps/src/main/java/org/eclipse/kapua/service/device/registry/steps/BrokerSteps.java
elbert3: See above.

PR #3959, opened 2024-02-06
Feature - Access Token fetch improval

Agnul97:
**Brief description of the PR and some context**.
This past PR https://github.com/eclipse/kapua/pull/3022 tried to improve the performance of access token retrieval from DB by introducing a new index on the 'expires_on' column. The rationale was that queries to the Access Token table used the index to exclude, at the beginning, expired tokens, thus allowing to search only between the not expired ones. A subsequent PR introduced a similar index on the "refresh_expires_on" column, so fetching tokens for the "refresh token feature" was faster, following the same rationale. Here we try to find a better solution for the problem

The present PR changed the structure of the JWT we issue to find a more elegant, simple and faster solution for this problem. A 'token identifier' column has been added inside the Access Token table, properly indexed. This value (precisely, a UUID) is computed for each new JWT, inserted inside its payload and inside the proper Access Token record in the DB. This value is then used to look up the Access Token record upon receiving a JWT from a client.

The main benefits of this PR are:

1. Both the authentication of the REST API calls and the 'refresh token feature call' now search the respective Access Token record via the AuthenticationService.findAccessToken(tokenId) method, which receives the JWT content, extracts the token identifier from its payload and, exploiting the db inxed, finds the respective Acess Token record. In contrast, before, 2 dedicated queries, with a filter on the "expires_on" and "refresh_expires_on" columns, were used to exploit the 2 early-mentioned indexes, resulting in a redundant code and less efficient operation, on the DB side.
2. The old way of fetching Access Token records by filtering out expired (and invalidated) ones was not allowing us to understand if the provided JWT was indeed expired/invalidated. Changing the way we fetch Access Tokens makes it possible to do so, and dedicated Exceptions/messages have been introduced to cover different error scenarios. For example, now clients are notified if the provided refresh token is expired, or if the JWT has been invalidated and so on

One new set of integration tests (cucumber scenarios) have been added for the authentication and refresh token feature

NB:
Regarding the second point,  I realized in this past PR https://github.com/eclipse/kapua/pull/3954 that CORS filtering must be performed BEFORE sending to the client an error message. I found a way to propagate the error message to the CORS filter Class exploiting a dummy custom HTTP header (see the _onLoginFailure_ methond inside KapuaTokenAuthenticationFilter.java) that is read by the _CORSResponseFilter_ Class to set the error message for the client (and discarded in the end)

**Any side note on the changes made or points of discussion**
The verification of the signature of the JWT has been moved into the doGetAuthenticationToken method inside the AccessToken realm. This is safer wrt what done before 

The "old" tokenId column (containing the whole JWT) has been maintained in the Access Token DB table. It could still be useful in the future


service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
```diff
@@ -355,12 +353,29 @@ public AccessToken findAccessToken(String tokenId) throws KapuaException {
         return accessToken;
     }
 
+    public AccessToken findAccessToken(String jwt) throws KapuaException {
+        final JwtConsumer jwtConsumer = new JwtConsumerBuilder()
+                .setSkipAllValidators()
```
dseurotech: Maybe we can build this jwtConsumer once in the constructor, or even inject it from outside, instead of building a new one at each call?

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
```diff
@@ -355,12 +353,29 @@ public AccessToken findAccessToken(String tokenId) throws KapuaException {
...
+                .setDisableRequireSignature()
+                .setSkipSignatureVerification()
+                .build();
+
+        try {
+            final JwtContext jwtContext = jwtConsumer.process(jwt);
+            final String tokenIdentified = Optional.ofNullable(jwtContext.getJwtClaims().getClaimValue(AccessTokenAttributes.TOKEN_IDENTIFIER, String.class))
```
dseurotech: very minor: tokenIdentifieR (wrong last letter)

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
```diff
@@ -355,12 +353,29 @@ public AccessToken findAccessToken(String tokenId) throws KapuaException {
...
+                    .orElseThrow(() -> new AuthenticationException());
+            return KapuaSecurityUtils.doPrivileged(() -> accessTokenService.findByTokenId(tokenIdentified));
+        } catch (InvalidJwtException | MalformedClaimException e) {
+            throw new AuthenticationException();
+        }
+    }
+
     @Override
     public AccessToken findRefreshableAccessToken(String tokenId) throws KapuaException {
```
dseurotech, line 374: Is this method still used? Should it be? Now that we have a better (as in: more efficient) way to retrive tokens, is it necessary to use queries?

service/security/shiro/src/main/resources/liquibase/2.0.0/atht-newTokenIdentifier.xml
```diff
@@ -0,0 +1,41 @@
...
+    <changeSet id="changelog-access_token-2.0.0_new_token_id" author="eurotech">
+<!--        start by revoking old tokens deleting associated rows in the table -->
+        <delete tableName="atht_access_token"/>
+        <addColumn tableName="atht_access_token">
+            <column name="token_identifier" type="varchar(36)"/>
+        </addColumn>
+        <addUniqueConstraint tableName="atht_access_token"
+                             columnNames="token_identifier"
+                             constraintName="uc_token_identifier"/>
+        <addNotNullConstraint tableName="atht_access_token" columnDataType="varchar(36)" columnName="token_identifier"/>
```
dseurotech: is there a way to specify the contraintName for this notNull constraint?

dseurotech (changes requested):

service/security/shiro/src/main/resources/liquibase/2.0.0/atht-newTokenIdentifier.xml
Agnul97: yes there is https://docs.liquibase.com/change-types/add-not-null-constraint.html. I was just not sure that both h2 and mariadb would support name for the constraint so I added it as we done in other parts of the codebase. I can test this thing

service/security/shiro/src/main/java/org/eclipse/kapua/service/authentication/shiro/AuthenticationServiceShiroImpl.java
Agnul97: it is actually not used but I maintained it as it was. Mainly because it's still part of the interface

service/security/shiro/src/main/resources/liquibase/2.0.0/atht-newTokenIdentifier.xml
Agnul97: edit: it seems that the liquibase version we are using doesn't allow to do so. Liquibase complaints if I try to add the "constraintName" attribute to the _addNotNullConstraint_ element

dseurotech (approved):

Coduz (approved):

PR #3964, opened 2024-02-09
:fix: Role update

dseurotech:
corner case where original entity is detached upon update, failing its refresh

### DIFF
--- a/service/security/shiro/src/main/java/org/eclipse/kapua/service/authorization/role/shiro/RoleServiceImpl.java
+++ b/service/security/shiro/src/main/java/org/eclipse/kapua/service/authorization/role/shiro/RoleServiceImpl.java
@@ -165,7 +165,7 @@ public class RoleServiceImpl extends KapuaConfigurableServiceBase implements Rol
                 throw new KapuaDuplicateNameException(role.getName());
             }
             current.setName(role.getName());
-            current.setName(role.getDescription());
+            current.setDescription(role.getDescription());
             // Do update
             return roleRepository.update(tx, current, current);
         });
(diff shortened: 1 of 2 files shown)


service/security/shiro/src/main/java/org/eclipse/kapua/service/authorization/role/shiro/RoleServiceImpl.java
```diff
@@ -164,8 +164,10 @@ public Role update(Role role) throws KapuaException {
             if (roleRepository.countOtherEntitiesWithNameInScope(tx, role.getScopeId(), role.getId(), role.getName()) > 0) {
                 throw new KapuaDuplicateNameException(role.getName());
             }
+            current.setName(role.getName());
+            current.setName(role.getDescription());
```
Coduz: `current.setDescrption(role.getDescription());`

service/security/shiro/src/main/java/org/eclipse/kapua/service/authorization/role/shiro/RoleServiceImpl.java
```diff
@@ -164,8 +164,10 @@ public Role update(Role role) throws KapuaException {
...
             // Do update
-            return roleRepository.update(tx, current, role);
+            return roleRepository.update(tx, current, current);
```
Coduz: Is it correct passing twice `current`, even if signature is `public E update(TxContext txContext, E currentEntity, E updatedEntity`?

Coduz (changes requested):
One typo to fix
One clarification needed

service/security/shiro/src/main/java/org/eclipse/kapua/service/authorization/role/shiro/RoleServiceImpl.java
dseurotech: Yes. The "current" one is used only to copy the creation fields:
image

So passing the same one, if fetched from the db and not coming from the outside, is a valid case

Coduz (approved):

PR #3986, opened 2024-03-08
:recycle: reenabled mfa retrieval and deletion for users, removed use…

dseurotech:
…less scopeId for self-managing of Mfa

Removes the deprecation from users managing other users for mfa options query, deletion and trust removal (deprecation on mfa creation remains as it is the prerogative of the user self-managing her own credentials.

Moved self-care mfa apis from /{scopeId}/user/mfa... to /user/mfa...  as the scope does not apply to this case anyway.


rest-api/resources/src/main/java/org/eclipse/kapua/app/api/resources/v1/resources/UsersMfa.java
```diff
@@ -0,0 +1,126 @@
...
+     * @return HTTP 200 if operation has completed successfully.
+     * @throws KapuaException Whenever something bad happens. See specific {@link KapuaService} exceptions.
+     * @since 1.4.0
+     */
+    @DELETE
+    @Path("{userId}/mfa")
+    public Response deleteMfa(
+            @PathParam("scopeId") ScopeId scopeId,
+            @PathParam("userId") EntityId userId) throws KapuaException {
+        mfaOptionService.deleteByUserId(scopeId, KapuaSecurityUtils.getSession().getUserId());
```
Coduz: Tested and this is not working. 
It is using the user from the session and not the given one!

Coduz (changes requested):
DELETE _/users/{userId}/mfa does not work.

Other methods work fine!

rest-api/resources/src/main/java/org/eclipse/kapua/app/api/resources/v1/resources/UsersMfa.java
dseurotech: I see that now! Will fix asap

Coduz (approved):

PR #3993, opened 2024-03-15
Fix account rest apis

dseurotech:
Update con account-related apis.
Splitted the updating of the current account (expiration date cannot be edited) vs updating a child account.
DTOs paradigm introduced for account update requests (there is no point in passing the full object and then verifying that only the allowed fields have changed, and all the other read-only fields have been passed as they are).
For some account-specific apis, the prefix {scopeId} has been removed, as the account IS the scope.


Coduz:
Ip review submitted
https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/13945

Coduz (approved):

PR #4012, opened 2024-04-05
:fix: add auto reconnect to service client

riccardomodanese:
Brief description of the PR.
Added a connection retry when the conection is dropped.

**Related Issue**
fixes 4011

Like for the service event connection a connection retry was added. This is triggered, through the connection event listener, when the connection is dropped.

**Screenshots**
none

none


commons/src/main/java/org/eclipse/kapua/commons/event/jms/JMSServiceEventBus.java
```diff
@@ -156,9 +158,8 @@ public ServiceEventBus getEventBus() {
     }
 
     private void waitBeforeRetry() {
-        // wait a bit
         try {
-            Thread.sleep(2000);// TODO move to configuration
+            Thread.sleep(WAIT_BETWEEN_RECONNECTION_ATTEMPT);
```
elbert3, line 162: Should we add a backoff function here, instead of a fixed sleep?

client/security/src/main/java/org/eclipse/kapua/client/security/amqpclient/Client.java
```diff
@@ -73,12 +77,61 @@ public TextMessage createTextMessage() throws JMSException {
...
+            return connection!=null && connection.getClientID()!=null && connectionStatus;
         } catch (JMSException e) {
             logger.warn("Error while validating connection: {}", e.getMessage(), e);
             return false;
         }
     }
+
+    private void waitBeforeRetry() {
+        try {
+            Thread.sleep(WAIT_BETWEEN_RECONNECTION_ATTEMPT);
```
elbert3, line 132: Should we add a backoff function here, instead of a fixed sleep?

client/security/src/main/java/org/eclipse/kapua/client/security/amqpclient/Client.java
```diff
@@ -73,12 +77,61 @@ public TextMessage createTextMessage() throws JMSException {
...
+                connection.setClientID(clientId);
+                connection.start();
+                session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
+                logger.info("AMQP client binding request sender to: {}", requestAddress);
+                logger.info("AMQP client binding message listener to: {}", replyAddress);
+                consumer = session.createConsumer(session.createQueue(replyAddress));
+                consumer.setMessageListener(clientMessageListener);
+                producer = session.createProducer(session.createQueue(requestAddress));
+                clientMessageListener.init(session, producer);
+                connectionStatus = true;
```
elbert3, line 110: Is there not a way to read the connection status from one of these objects instead of managing it ourselves?  Should it be set to false on exception?

client/security/src/main/java/org/eclipse/kapua/client/security/amqpclient/Client.java
```diff
@@ -73,12 +77,61 @@ public TextMessage createTextMessage() throws JMSException {
...
 
-    public boolean isConnected() {
+    public void stop() throws JMSException {
+        active = false;
+        disconnect();
+    }
+
+    private void disconnect() throws JMSException {
+        if (connection != null) {
+            connection.close();
```
elbert3, line 88: Shouldn't we set the connectionStatus here to false?

elbert3 (commented):
Hi @riccardomodanese, I added a few questions.

client/security/src/main/java/org/eclipse/kapua/client/security/amqpclient/Client.java
riccardomodanese: Unfortunately no. The javax.jms.Connection has no connect flag o isConnect method.

commons/src/main/java/org/eclipse/kapua/commons/event/jms/JMSServiceEventBus.java
riccardomodanese: I preferer to stay with a fixed (and not too big) amount of waiting time since I'm expecting that the service broker will be down for not too long periods. With a backof the risk is to have the waiting time very huge at a certain time

client/security/src/main/java/org/eclipse/kapua/client/security/amqpclient/Client.java
riccardomodanese: I preferer to stay with a fixed (and not too big) amount of waiting time since I'm expecting that the service broker will be down for not too long periods. With a backof the risk is to have the waiting time very huge at a certain time

client/security/src/main/java/org/eclipse/kapua/client/security/amqpclient/Client.java
riccardomodanese: I was relying on the ExceptionListener but you are right: it's better to set it here.

commons/src/main/java/org/eclipse/kapua/commons/event/jms/JMSServiceEventBus.java
elbert3: With a backoff we could define a maximum value, but I'm okay proceeding with this simpler solution unless we find a need to change it otherwise.

client/security/src/main/java/org/eclipse/kapua/client/security/amqpclient/Client.java
elbert3: With a backoff we could define a maximum value, but I'm okay proceeding with this simpler solution unless we find a need to change it otherwise.

client/security/src/main/java/org/eclipse/kapua/client/security/amqpclient/Client.java
elbert3: Okay let's add it in, then I'd be okay approving this.

riccardomodanese: done!

elbert3 (approved):

PR #4013, opened 2024-04-08
:bug: random uuid in event client id resulted in multiple instances o…

dseurotech:
…f the same component processing the same messages, resulting in a lot of data update conflicts

Brief description of the PR.
In a previous PR a random component in event bus clients id was introduced in order to allow multiple components (e.g.: lifecycle, telemetry, etc) to consume the same event. 
Multiple instances of the same component, however, should not process the same event multiple times, as this would result in potential update conflicts on affected entities.
As the event module name (unique for each component) was already previously introduces, this PR removes the unnecessary random component which was added at configuration time.

### DIFF
--- a/commons/src/main/java/org/eclipse/kapua/commons/event/ServiceEventTransactionalModule.java
+++ b/commons/src/main/java/org/eclipse/kapua/commons/event/ServiceEventTransactionalModule.java
@@ -12,23 +12,22 @@
  *******************************************************************************/
 package org.eclipse.kapua.commons.event;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+
 import org.apache.commons.lang3.StringUtils;
 import org.eclipse.kapua.KapuaException;
 import org.eclipse.kapua.commons.core.ServiceModule;
 import org.eclipse.kapua.event.ServiceEventBus;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
-import java.util.UUID;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ScheduledFuture;
-import java.util.concurrent.TimeUnit;
 
 public abstract class ServiceEventTransactionalModule implements ServiceModule {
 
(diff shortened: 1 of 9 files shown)


riccardomodanese (approved):

commons/src/main/java/org/eclipse/kapua/commons/event/ServiceEventTransactionalModule.java
```diff
@@ -53,14 +52,6 @@ private String getSubscriptionName(String address, String subscriber) {
     private final ServiceEventHouseKeeperFactory houseKeeperFactory;
     private final ServiceEventBus serviceEventBus;
 
-    public ServiceEventTransactionalModule(
-            ServiceEventClientConfiguration[] serviceEventClientConfigurations,
-            String internalAddress,
-            ServiceEventHouseKeeperFactory serviceEventTransactionalHousekeeperFactory,
-            ServiceEventBus serviceEventBus) {
-        this(serviceEventClientConfigurations, internalAddress, UUID.randomUUID().toString(), serviceEventTransactionalHousekeeperFactory, serviceEventBus);
```
elbert3, line 61: Couldn't there be scenarios where you'd want all instances of a client to receive the events (i.e. it should use a unique id)?  This is still possible by generating a unique client id in the classes that extend this, but it would seem simpler to consolidate that option here (i.e. add an additional "useUniqueId" boolean in the constructor, and append a UUID to the specified clientId if true).

I'm okay with this current solution, but this might be an improvement to consider.

service/account/internal/src/main/java/org/eclipse/kapua/service/account/internal/AccountImpl.java
```diff
@@ -178,5 +182,22 @@ public void setExpirationDate(Date expirationDate) {
...
+        if (this == object) {
+            return true;
+        }
+        if (!(object instanceof Account)) {
+            return false;
+        }
 
+        final Account that = (Account) object;
+
+        return getId() != null ? getId().equals(that.getId()) : that.getId() == null;
```
elbert3: I don't think that just comparing the id is sufficient to determine if two Account objects are equal; it should be comparing all of the fields.  For example, if you are updating the expiration date for an Account, it should no longer be considered "equals" to the original object with the same id but different expiration date.

I realize that this can get complicated dealing with parent classes and other entity types, so I'd be okay proceeding with this and dealing with proper "equals" and "hashCode" implementations as a separate issue.

dseurotech: Considering that the class AccountImpl is the JPA representation of the database table, it is common to reflect the primary key of the table in the equals/hashcode. The kind of comparison you are referring to often requires custom logic, and is better left in an external collaborator. The intended use here is to check whether we are talking about the same entity in a broader sense. 
That said, given the other changes introduced here and elsewhere, this particualar addition of equals and hashcode could even be omitted.

commons/src/main/java/org/eclipse/kapua/commons/event/ServiceEventTransactionalModule.java
dseurotech: It should not really be the concern of this class wheter you want a unique id or not, nor it should be it to decide the randomization strategy (what if you want to not use uuids in the future, for whatever reason?). As you said, as long as it is possible to customize the client id, deciding from outside whether you want a fixed one or a random one, the purpose of this particular class is fulfilled. Composition and SRP both suggest to leave the client id creation outside.

riccardomodanese: agree. let's go with a simpler solution for now.

elbert3 (approved):

PR #4066, opened 2024-06-28
:bug: Avoid multiple GuiceLocator instances, and handle potential multiple KapuaMessageListeners

dseurotech:
Should guicelocator fail at broker startup, multiple instances of (Kapua)MessageListener are istantiated, failing device login due to split request/response maps.
In general, guice modules should avoid doing logic, and should avoid failing. Guice cannot guarantee that singleton objects previously instantiated are killed - which in some cases can be very distruptive.
This pr both avoids multiple guiceLocator instantiation attempts within a single process, but also puts in place safeguards that ensure that message correlation will continue to work even if there are multiple (Kapua)MessageListener (possibly due to changes in the wiring).


Coduz (approved):

riccardomodanese (approved):
do not agree with this solution to solve the locator initialization issue but anyway, I'll aprove it

PR #4017, opened 2024-04-11
Ref inter dependencies cleanup

dseurotech:
Merged into https://github.com/eclipse/kapua/pull/4073

PR #4073, opened 2024-07-03
Break job engine to rest api dependency

dseurotech:
This PR breaks the dependency that job-engine had to the rest-api, reorganising the commons-rest modules into a more structured commons-jersey modules (splitted in to rest and web submodules), that contain all the necessary components to handle jersey-based services (be they web apps or rest apis).

Breaking changes!

Also absorbs: https://github.com/eclipse/kapua/pull/4017
This PR intends to clean up the internal dependencies between kapua modules. In most cases, it just cleans up the pom.xml files from dependencies which are already received transitively via maven (e.g.: if you already depend on the -internal module, it is implicit the dependence from the -api module).
In a few cases data classes have been moved closer to the project that use them -e.g.: IsJobRunningMultipleResponse.java is specific to the job engine and does not need to be in the commons-rest-model module.


Coduz:
This PR has been superseded by https://github.com/eclipse-kapua/kapua/pull/4293 and can be closed when #4293 will be merged

PR #4096, opened 2024-08-06
FIX - deleteAllIndexes & refreshAllIndexes methods re-factoring

Agnul97:
The deleteAllIndexes method, called in tests, was calling 4 times a low-level ES method that was deleting all the indexes stored. This was accomplished by calling the various deleteAllIndexes in the 4 *repositories classes, that had the same common implementation. I re-factored this implementation in order to delete only the respective repository indexes. 
Furthermore, I modified the refreshAllIndexes implementation, with the same logic.
This fix serves mainly to prepare the codebase to the usage of OpenSearch at the place of ES.

### DIFF
--- a/service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/MessageStoreFacadeImpl.java
+++ b/service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/MessageStoreFacadeImpl.java
@@ -498,6 +498,9 @@ public final class MessageStoreFacadeImpl extends AbstractDatastoreFacade implem
     @Override
     public void refreshAllIndexes() throws ClientException {
         messageRepository.refreshAllIndexes();
+        clientInfoRepository.refreshAllIndexes();
+        channelInfoRepository.refreshAllIndexes();
+        metricInfoRepository.refreshAllIndexes();
     }
 
     @Override
--- a/service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/MessageStoreFacadeImpl.java
+++ b/service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/MessageStoreFacadeImpl.java
@@ -498,9 +498,6 @@ public final class MessageStoreFacadeImpl extends AbstractDatastoreFacade implem
     @Override
     public void refreshAllIndexes() throws ClientException {
         messageRepository.refreshAllIndexes();
-        clientInfoRepository.refreshAllIndexes();
-        channelInfoRepository.refreshAllIndexes();
-        metricInfoRepository.refreshAllIndexes();
     }
 
     @Override
--- a/service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/mediator/DatastoreUtils.java
+++ b/service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/mediator/DatastoreUtils.java
@@ -136,9 +136,11 @@ public class DatastoreUtils extends AbstractStoreUtils {
         if (StringUtils.isNotEmpty(prefix)) {
             sb.append(prefix).append("-");
         }
-        String indexName = normalizedIndexName(scopeId.toStringId());
-        if (scopeId.equals(KapuaId.ANY)) {
+        String indexName;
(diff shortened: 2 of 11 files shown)


service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/mediator/DatastoreUtils.java
```diff
@@ -137,6 +137,9 @@ public String getDataIndexName(KapuaId scopeId) {
             sb.append(prefix).append("-");
         }
         String indexName = normalizedIndexName(scopeId.toStringId());
+        if (scopeId.equals(KapuaId.ANY)) {
+            indexName = "*";
```
dseurotech, line 141: Overriding a previously set value is conceptually wrong - both because you waste the first calculation (however trivial it might be) and also because the code becomes harder to read (If I want to read through in order to figure out the final value of the variable, I have to maintain the cognitive load of the first calculation, only to then discard the process when it gets overwritten).
TL;DR: use an if-then-else construct here.

Also, when comparing unknown variables with known quantities, try to always use the known value as the base for the comparison:
KapuaId.ANY.equals(scopeId) will never throw NPE, and will still work against nulls.

service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/mediator/DatastoreUtils.java
```diff
@@ -199,6 +202,9 @@ private String getRegistryIndexName(KapuaId scopeId, IndexType indexType) {
             sb.append(prefix).append("-");
         }
         String indexName = normalizedIndexName(scopeId.toStringId());
+        if (scopeId.equals(KapuaId.ANY)) {
```
dseurotech: See comment above

service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/MessageStoreFacadeImpl.java
```diff
@@ -498,9 +498,6 @@ private boolean isAnyClientId(String clientId) {
     @Override
     public void refreshAllIndexes() throws ClientException {
         messageRepository.refreshAllIndexes();
-        clientInfoRepository.refreshAllIndexes();
-        channelInfoRepository.refreshAllIndexes();
-        metricInfoRepository.refreshAllIndexes();
     }
```
dseurotech: Why removing the other three refreshes? The facade feels like the correct place to call refresh on all involved repositories at once

dseurotech: Ah, I see now that the elasticsearch implementation of refreshAllIndexes really works on all indexes. Is this ok? Does this work with other implementations like opensearch (sounds strange to me that the "delete all" won't work, but "refresh all" is ok.
Conceptually, I'd look into refreshing only the indexes concerning the specific repository, as per the delete, and keep calling all four repositories from the facade.

dseurotech (changes requested):

service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/MessageStoreFacadeImpl.java
Agnul97: I think OpenSearch complains only about the deleteAllIndexes permissions and not refreshAll because of the possible danger of the first operation wrt the other (despite being admin logged)
I did not perform a similar operation as you propose because in our low-level client, there is not a method to refresh a precise index so I left just 1 call to the call to the "refresh all indexes"

service/datastore/internal/src/main/java/org/eclipse/kapua/service/datastore/internal/mediator/DatastoreUtils.java
Agnul97: I've done this because I assumed that in any case first of all we want to normalize the index to validate its format (even if it's null, like it is done now) , but ok I can do as you suggest

dseurotech (approved):

PR #4121, opened 2024-10-18
✨ [REST-API] Add support to query operations by `status` in `GET /{scopeId}/devices/{deviceId}/operations`

MDeLuise:
### Description
This PR implements the capability to search operations by their status using the `GET /{scopeId}/devices/{deviceId}/operations` endpoint.

### Changes Made
- **New Feature**: Added functionality to filter operations based on their status.
- **Refactor**: Modified the `DeviceManagementOperations` query method to support additional predicates.

The current query method was refactored because it was enforcing a "filter by device ID" constraint on all queries, regardless of any other predicates provided. This limitation prevented any other filtering criteria (e.g., status) from being applied, so the method has been updated to allow flexible filtering based on multiple criteria.

### DIFF
--- a/rest-api/resources/src/main/java/org/eclipse/kapua/app/api/resources/v1/resources/DeviceManagementOperations.java
+++ b/rest-api/resources/src/main/java/org/eclipse/kapua/app/api/resources/v1/resources/DeviceManagementOperations.java
@@ -41,6 +41,7 @@ import org.eclipse.kapua.service.device.management.registry.operation.DeviceMana
 import org.eclipse.kapua.service.device.management.registry.operation.DeviceManagementOperationListResult;
 import org.eclipse.kapua.service.device.management.registry.operation.DeviceManagementOperationQuery;
 import org.eclipse.kapua.service.device.management.registry.operation.DeviceManagementOperationRegistryService;
+import org.eclipse.kapua.service.device.management.registry.operation.DeviceManagementOperationStatus;
 import org.eclipse.kapua.service.device.registry.Device;
 
 @Path("{scopeId}/devices/{deviceId}/operations")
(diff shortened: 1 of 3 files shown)


rest-api/resources/src/main/java/org/eclipse/kapua/app/api/resources/v1/resources/DeviceManagementOperations.java
```diff
@@ -117,10 +122,6 @@ public DeviceManagementOperationListResult query(
             DeviceManagementOperationQuery query) throws KapuaException {
         query.setScopeId(scopeId);
 
-        AndPredicate andPredicate = query.andPredicate();
```
Coduz, line 120: This shouldn't be removed otherwise invoking directly `_query` will not filter by `deviceId`.
It should be something like:
```
if (query.getPredicate() != null) {
    AndPredicate andPredicate = query.andPredicate(
        query.attributePredicate(DeviceManagementOperationAttributes.DEVICE_ID, deviceId),
        query.getPredicate()
    )
    query.setPredicate(andPredicate);
}
else {
    query.setPredicate(query.attributePredicate(DeviceManagementOperationAttributes.DEVICE_ID, deviceId));
}
```
which always applies `deviceId` filtering and preserves `query.predicate` if specified (either from REST body or via  `simpleQuery`)

Coduz (changes requested):

rest-api/resources/src/main/java/org/eclipse/kapua/app/api/resources/v1/resources/DeviceManagementOperations.java
MDeLuise: Thanks for your feedback
That sounds good, and I've updated the code accordingly.

Coduz (approved):

PR #4167, opened 2025-01-10
FIX: Improved error message in the case of SQL constraint violation on foreign keys

Agnul97:
Brief description of the PR.
I wanted to return a more specific error message that notifies clients that they are trying to break a fk constraints, to let them know precisely what they are doing wrong. Currently, in such cases, a too general 500 internal persistence error is returned

With these changes, a 409 error is returned instead, with a message that notifies precisely a fk violation if it's detected. Otherwise, a more general error message is returned if other SQL constraint violations are present


Coduz (changes requested):
Changes are ok, but the fact that it would be better if that was a generic exception that does not strictly relates to SQL and Foreign Keys.

`KapuaIntegrityConstraintViolationException` with message `Entity constraint violation error. This entity relates to other entities and cannot be deleted.`

Coduz (approved):

PR #4198, opened 2025-02-19
:fix: fix/improve broker security plugin

riccardomodanese:
Brief description of the PR.
various fixes and improvement to make security plugin more stable

**Related Issue**
none

see comment below

**Screenshots**
none

none


riccardomodanese:
- Fixed the event subscription group handling
renamed variables from eventModuleName to subscriptionGroupId so it's more clear the meaning (since a module could have more subscriptions to the same address but different groups.
This change had impact in other areas because now every event handler should receive, once instantiated, the appropriate subscription group name.
For example, for the service , the generic
`@Named("eventsModuleName") String eventModuleName`
will be replaced with the correct
`@Named("userEvtSubscriptionGroupId") String subscriptionGroupId`
so instead of having the module name as group for this event subscription, the "svc-ath-" + containerIdResolver.getContainerId() string will be returned
```
@Provides
    @Named("userEvtSubscriptionGroupId")
    String userEvtSubscriptionGroupId(ContainerIdResolver containerIdResolver) {
        return getSubscriptionId(containerIdResolver);
    }

    private String getSubscriptionId(ContainerIdResolver containerIdResolver) {
        return "svc-ath-" + containerIdResolver.getContainerId();
    }
}
```

- re-organized the logic to set message headers in the ServerPlugin and the message categorization
- Improved error handling to serialize messages before discarding them (to byte[]) and the abstract message converter (used more generic Camel Message instead of the JmsMessage)
- removed trusted classed from KapuaSession "security" check. That's no security improvement on doing that, only performances impact.
- removed client id set if null on connect (it's not compliant with our use case using JMS 2.0) and modified the Camel connection factory to remove this parameter
- changed the exceptions thrown by ServiceClient (the one used by AuthenticationService) generic because the implementations could be various obviously
- SERVICE_BROKER_HOST and SERVICE_BROKER_PORT are replaced by a new env variable: SERVICE_BROKER_URL (default value amqp://events-broker:5672). Credentials are provided through SERVICE_BROKER_USERNAME and SERVICE_BROKER_PASSWORD.
In this way more complex connection strings can be defined but, also, different event/service broker protocols could be used like MQTT or JMS.
- Same for the event broker where  EVENT_BROKER_URL will contain the full connection string to the event broker (see Artemis documentation from more informations). Credentials are provided through EVENT_BROKER_USERNAME and EVENT_BROKER_PASSWORD.

riccardomodanese:
Backport to 2.0.0 with https://github.com/eclipse-kapua/kapua/pull/4217

PR #4205, opened 2025-03-03
Add the organization name to the "switch to account" menu

### LINKED ISSUE #4204: Add the organization name to the "switch to account" menu

Currently, the Kapua account name is immutable.  But in our case, sometimes, an account belonging to a commercial entity may be selled to another commercial entity. Therefore, because of account name immutability, we cannot use a natural name as an account name.

To workaround this, we use UUIDs for account names and use natural names in the mutable `organisation name` attribute.

This gives us a `switch to account` menu that is hard to use:

Image

What I'd really like is a mutable account name. But I guess there are implications and difficulties I am not aware of. So instead, I propose to add the organization name to the account name in the `switch to account` menu.

Something like `$account_name ($organization_name)`.

Mutable account names ?

reda-alaoui:
This change adds the organization name to the `switch to account` gwt menu

**Related Issue**
This PR fixes #4204


reda-alaoui:
Ping :)

PR #4224, opened 2025-03-31
:bug: [Datastore] Fixed the logic checking message uniqueness when a insert is retried

Coduz:
This PR fixes the check performed to honor the message uniqueness check strategy.

**Related Issue**
_None_

The `capturedOn` field can be `null` and it would produce an error when the selected strategy is BOUND.
We switched to the indexed date which is always not null 

**Screenshots**
_None_

_None_


stefanomorson (approved):
Looks fine to me as it alignes with the scope of the BOUND policy.

PR #4259, opened 2025-07-24
:bug: [REST API] Fixed querying for binary Metrics

Coduz:
This PR fixes querying for `binary` metrics via REST API

`binary` mapped metrics are not searchable in Elasticsearch: https://www.elastic.co/guide/en/elasticsearch/reference/7.17/binary.html

**Related Issue**
_None_

Fixed handling of parameters `metricType` and `metricMin`/`metricMax` to be consistent with the semantics of the API when `metricType` is `binary`. In fact, the API should find all messages where the metric with name `metricName` exists and (optionally) the value included in the range [`metricMin`, `metricMax`]. 
Unfortunately with current datastore engine, search for existence of a `binary` field is not supported, moreover by their nature binary values are not comparable, hence specifying the parameters `metricMin` and `metricMax` is meaningless.
For these reasons when metric type is `binary` an illegal argument error is thrown/returned. User can still use the API to query the datastore without the metric parameter and then inspect the returned messages to find the ones which satisfy the criteria. 
 
**Screenshots**
_None_

_None_

### DIFF
--- a/rest-api/core/src/main/java/org/eclipse/kapua/app/api/core/model/MetricType.java
+++ b/rest-api/core/src/main/java/org/eclipse/kapua/app/api/core/model/MetricType.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2022 Eurotech and/or its affiliates and others
+ * Copyright (c) 2016, 2025 Eurotech and/or its affiliates and others
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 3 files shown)


stefanomorson:
@Coduz, as far as I understand from the code, when the GET is made on a metric of type `binary` with `metricMin` and `metricMax` undefined, it simply returns an empty result set while if the query the GET also includes a value for `metricMin` or `metricMax` then an error is thrown. Correct? If so I'm ok with this change as the underlying engine can't index a binary field.

stefanomorson (approved):

PR #4310, opened 2025-11-25
:bug: [Rest API] API call /{scopeId}/authorization/claims now correctly retrieve claims when permissions restricted to groups

Agnul97:
Brief description of the PR.
The issue showed up when, for example, the API /{scopeId}/authorization/claims is executed; only permissions granted to the user in session with access group = ALL are listed. If a permission is restricted to an access group, the related claim is not returned. 

A new approach for the fetchUserClaims method is implemented, which reverts the enumeration of all possible domain/actions to a simpler approach, that works with groups.

I also added a lot of test scenarios to cover this, and to do so I evalutaed as convenient to insert an abstraction for the "ClaimsFetcher", with 2 possible implementation: the new one  that I wrote and the old approach that is used in tests to compare results of "claims fetching" between the 2 approaches/implementations

### DIFF
--- a/qa/integration/src/test/resources/features/user/UserRoleServiceI9n.feature
+++ b/qa/integration/src/test/resources/features/user/UserRoleServiceI9n.feature
@@ -589,8 +589,8 @@ Feature: User role service integration tests
     When I restrict permission to last created group
     When I create the permissions
     Given I select the domain "job"
-    And The permissions "all"
-    When I create the permissions
+    And The permission "all"
+    When I create the permission
     Then exception is not thrown
     And I logout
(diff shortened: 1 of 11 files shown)


stefanomorson (commented):
I would give the `BruteForceClaimsFetcher` a different name as `bruteforce`evokes a scenario related to security stuff or at least to a process that tries break something. 
Regarding the comment `//BE AWARE THAT PERFORMS WRONG IN PRESENCE OF PERMISSIONS RESTRICTED TO GROUPS`
If it is wrong why do we need to keep it in the code base? Or is it used in some specific scenario?

Coduz (approved):

PR #4314, opened 2025-11-27
:bug: [Docker] Allow java process to receive properly the sigterm from container

riccardomodanese:
Brief description of the PR.
The java process for consumers and service doesn't receive the sigterm from Docker container properly.
This pr will fix it.

**Related Issue**
none

Added 'EXEC' before the java command in the consumers/service entrypoints. In this way the java process will have the process id 1 and will receive the sigterm from the docker container properly.

**Screenshots**
none

none


matteoipri (approved):
LGTM. This is the best practice for bash script entrypoint.

Coduz (approved):

PR #4357, opened 2026-02-12
✨ [REST-API] Device management for Kura Wires

Agnul97:
Kura WIRE-V1 request handler allows to remotely manage several aspects of the Wire Graph functionality

Among the endpoints offered by the request handler, the following three are the object of this PR:

GET {scopeId}/devices/{deviceId}/graph/snapshot
PUT {scopeId}/devices/{deviceId}graph/snapshot
DEL {scopeId}/devices/{deviceId}graph

With the GET a user is able to download the wire graph snapshot, a subset of the device configuration limited to the wire graph and wiregraph-related components, with the Kapua normalised encoding. Wiregraph snapshot can be prepared for later upload on another device or in the same device.

With the DEL the user can cleanly delete the current wire graph in the target device so that it can be safely replaced by a new graph avoiding issues due to the merge of components operated by Kura.

With the PUT the user can upload the wire graph snapshot.

This PR adds such operations along with their rest-api.


stefanomorson (changes requested):
To contextualise these resources (graph and snapshots) to the Kura wire feature it would be better to add "wire" in the path so that the three new API will look like the following:
- GET {scopeId}/devices/{deviceId}/wire/graph/snapshot
- PUT {scopeId}/devices/{deviceId}/wire/graph/snapshot
- DEL {scopeId}/devices/{deviceId}/wire/graph

In case that in the future we decide to support other Kura Wire endpoints like fore example 
- GET /assets/pids
- GET /driver/pids

We can keep them clearly separated from existing Kapua API like:
- GET /{scopeId}/devices/{deviceId}/assets

Coduz (approved):

PR #4373, opened 2026-03-25
:sparkles: Added to UI the checkbox to install package after download

Agnul97:


stefanomorson (changes requested):
The `Install After Download` should be the first of the list to reflect the logical sequence. Please move it before the `Reboot After Install`.

stefanomorson (approved):

